PR #5, opened 2017-12-05
Fix registry strategy tests that were incorrectly testing that resources were not namespace-scoped

staebler:
Cluster, NodeGroup, and Node should all be namespace-scoped.


staebler:
Included in #6.

PR #6, opened 2017-12-05
Resolve outstanding issues related to jenkins CI

staebler:
* Default to building natively instead of using docker containers. Set USE_DOCKER=1 to build inside docker container. Replaces NO_DOCKER=1 to build natively.
* Fix bugs in registry strategy tests so that they verify that Cluster, NodeGroup, and Node are namespace-scoped.
* Disable integration tests in CI until integration tests are set up properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ coverage: .init
 	$(DOCKER_CMD) contrib/hack/coverage.sh --html "$(COVERAGE)" \
 	  $(addprefix ./,$(TEST_DIRS))
 
-test: .init build test-unit test-integration
+test: .init build test-unit # test-integration # turn off test-integration until integration tests are set up
 
 # this target checks to see if the go binary is installed on the host
 .PHONY: check-go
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -240,7 +240,7 @@ coverage: .init
 	$(DOCKER_CMD) contrib/hack/coverage.sh --html "$(COVERAGE)" \
 	  $(addprefix ./,$(TEST_DIRS))
 
-test: .init build test-unit test-integration
+test: .init build test-unit # test-integration # turn off test-integration until integration tests are set up
```
eparis: could we fix the CI system to use `make test-unit` instead of disabling what a human is going to want to use?

staebler: My comments are a bit misleading, it appears. A human running `make test` will get the same failures when it comes time to run the integration tests. That defect is not limited to CI. I'll make it a priority to get integration tests working soon so that we can re-enable it for both humans and CI.

staebler: In other words, I don't think it is worthwhile to spend time changing what the CI system is running.

PR #7, opened 2017-12-06
Infra controller skeleton, logrus, developer improvements

dgoodwin:
Vendor and minimally use logrus. We're still using both in controller manager, trying to work this out a little better and see how it goes.

Allow re-running deploy.sh without breaking SSL.

Fix permissions errors in controllers being unable to interact with all the resources in our api group.

Skeleton for infra controller.


contrib/examples/deploy.yaml
```diff
@@ -284,7 +280,9 @@ objects:
           - --contention-profiling
           - "false"
           - -v
-          - "10"
+          - "0"
```
staebler: Can we leave this at 10 for now until everything is switched over to logrus?

contrib/examples/deploy.sh
```diff
@@ -2,7 +2,12 @@
 oc create namespace cluster-operator
 
 # Create ssl certs for api server
-./contrib/apiserver-aggregation-tls-setup.sh
+if [ ! -e apiserver.pem ]
+then
+	./contrib/apiserver-aggregation-tls-setup.sh
+else
+	echo "Re-using existing certs."
```
staebler: What is the benefit of re-using the certs? Is this a step towards making deploy.sh idempotent?

pkg/controller/infra/infra_controller.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// Clusters that need to be synced
+	queue workqueue.RateLimitingInterface
+
+	logger *log.Entry
+}
+
+func (c *InfraController) addCluster(obj interface{}) {
+	cluster := obj.(*clusteroperator.Cluster)
+	c.logger.Debugf("adding cluster %s", cluster.Name)
```
staebler: Its unfortunate that logrus does not have more granular logging levels than the 6.

contrib/examples/deploy.sh
dgoodwin: It appeared that controller manager would get SSL errors if you re-ran with new certs, that might be only if you have modified the controller manager deployment yaml and not the apiserver though.

contrib/examples/deploy.yaml
dgoodwin: Can do.

dgoodwin: Actually FWIW glog still logs things at 0, I was told recently that "0 is more verbose than info" by deads, which I totally do not understand, among several other things with glog.

pkg/controller/infra/infra_controller.go
dgoodwin: I could see the argument for one more "insane" level below debug but beyond that I think it gets a bit confusing.

I'm also not crazy about the mixing or loss of glog statements from the kube infra we're re-using. 

However for me the pros outweigh the cons. Our logs are what I'm interested in and we should be able to assume the kube infra we're re-using works, at least at some point once it's wired up. If we do end up with controllers in their own pods then we no longer will care much about glog output from the kube infra.

debug would be our highly verbose mode purely for debugging. (possibly re-use that to automatically re-enable glog once we shut it off), info is more high level and informative but still suitable to run in prod, and warn if you only care that something is, or might be wrong. I would envision that glog is silent by default once we reach a certain point.

The structured logging approach *should*  be really beneficial to us if we have the ELK stack in our final cluster, and from what I can tell a lot of projects in the kube/docker ecosystem are pushing away from glog to logrus. This should help us think more intelligently about how we'll log, and query those logs in prod.

I can be swayed though, WDYT?

contrib/examples/deploy.sh
staebler: Ah, I see. Yeah, restarting one of the deployments but not the other would cause problems. Is re-running deploy.sh going to work in that case any more after the changes from #8?

Should we just pull the cert creation script out entirely and require it be run separately at some point in the past?

pkg/controller/infra/infra_controller.go
staebler: The glogs coming from kube aren't only useful to trace errors in the infrastructure. They are also useful to trace errors in how our code is interacting with the infrastructure. For example, if a controller is not sending the correct information in a status update for a resource, that can be seen in the logs coming from the infrastructure.

It may very well be that the formalized structure that logrus gives in its log output obviates the usefulness of more granular log levels. I am willing to be amazed.

contrib/examples/deploy.sh
staebler: I think I may be ignorant about exactly the behavior of openshift templates. Does `oc apply` return 0 when the template would create resources that already exist? So that the `oc apply` part essentially skips over the apiserver deployment and only attempts to create the controller-manager deployment? If so, is it ignoring that the parameters given to the template have been changed?

dgoodwin: I can't see any issues with #8 at a glance, will test today.

Feels ok to let it decide if it should gen or not based on if they exist. (my check is a bit rudimentary right now but still) Little smoother onboarding new folks, one less pre-development step. I wouldn't mind nesting them under a certs/ dir though so they're easy to blow away.

dgoodwin: Regarding apply, my suspicion is that it updates the secret correctly with the new cert, but there's nothing that will automatically redeploy when a secret changed, so as far as kube is concerned the apiserver deployment is still the same.

PR #13, opened 2017-12-07
Set up integration test framework.

staebler:
* Fix test server start-up so that it works standalone.
* Add some initial clustercontroller integration tests that verify nodegroup creation when cluster is created.
* Re-enable integration tests in CI.

### DIFF
--- a/test/integration/clustercontroller_test.go
+++ b/test/integration/clustercontroller_test.go
@@ -1,0 +1,263 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/fake"
+
+	// avoid error `clusteroperator/v1alpha1 is not enabled`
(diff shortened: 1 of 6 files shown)


test/integration/clustercontroller_test.go
```diff
@@ -0,0 +1,263 @@
...
+// startServerAndClusterController creates a new test ClusterController injected with
+// fake clients and returns:
+//
+// - a fake kubernetes core api client
+// - a cluster-operator api client
+// - a function for shutting down the controller and API server
+//
+// If there is an error, startServerAndClusterController calls 'Fatal' on the
+// injected testing.T.
+func startServerAndClusterController(t *testing.T) (
```
dgoodwin: I really like this approach, it's a big win IMO to be able to just bring up our apiserver and etcd independently. However when we get to our controllers, can we integration test those in this same manner? The infra controller for example will attempt to create configmaps and jobs. We could stub that such that the pods it will launch mock what they would do somehow, but without a Kube API server + controllers, am I correct that that aspect of our code can't be integration tested in this fashion? 

We can mimic an awful lot of that with straight unit tests however, so maybe that's good enough, but I'm wondering what you envision for our controllers running ansible in this integration suite.

markturansky: You can have all the controllers, too, in your integration tests.  See 
https://github.com/openshift/online/blob/master/accountant/test/integration/helper.go#L103

There's `StartTestMaster` and there's `StartTestMasterAPI`.  The former has the API and all controllers (but not kubelet). The latter is just API.

dgoodwin: Yeah that should do it, but I think it means vendoring a whole lot more code. (all of origin)

staebler: The interactions with the Kube API server is done via the fake kube client. The integration tests would verify that the expected actions were taken and would establish reactions to those actions. This is the same as what we should expect to do with the AWS fake. The Kube API server is outside of the responsibility of what the cluster-operator integration tests should be testing.

staebler: Oh, I see the problem. The issue is that the infra controller is creating a pod to do the work, rather than doing it in-process. I don't think that this integration test framework is going to be able to test the behavior inside the pod, as it is.

dgoodwin: It's a question of where we draw the line of integration. If we could hit a point where we could literally run ansible pods and hit that mock AWS thing @twiest mentioned the other day, that would be about perfect.

If not, we could rig it such that it will launch jobs that just no-op and report success, and update status accordingly to unblock the other phases of the provisioning.

Or we can not launch the jobs at all. We can get an awful lot of coverage on the controllers with unit tests.

PR #15, opened 2017-12-08
Add validation for the API resources.

staebler:
* Add validation for the API resources.
* Remove NodeGroupName from Node as it is not currently used. It may not be needed at all depending upon how we implement nodes. So, let's get rid of it for now.
* Fix some fmt and lint warnings.


dgoodwin:
Very nice.

PR #16, opened 2017-12-08
Add an ansible playbook for devel deployments.

dgoodwin:
Replaces deploy.sh.

Begin creating a secret for AWS creds read from developers
~/.aws/credentials.


contrib/ansible/deploy-devel.yaml
```diff
@@ -0,0 +1,34 @@
...
+    command: oc create namespace cluster-operator
+    register: createns_reg
+    failed_when: createns_reg.rc > 0 and "AlreadyExists" not in createns_reg.stderr
+    changed_when: createns_reg.rc == 0
+
+  - name: generate apiserver certs
+    command: "{{ playbook_dir }}/../apiserver-aggregation-tls-setup.sh"
+    args:
+      # ensure these land in the top level of the project where expected
+      chdir: "{{ playbook_dir }}/../../"
```
staebler: Should be dirt simple to move these into their own directory now.

PR #17, opened 2017-12-11
provision cluster infra with openshift-ansible

dgoodwin:
Currently blocked on error where no boto3/botocore is available in the container. Will be submitting a PR tomorrow to openshift-ansible and possibly building my own container to bypass.

Open for review none the less.

### DIFF
--- a/contrib/examples/deploy.yaml
+++ b/contrib/examples/deploy.yaml
@@ -73,7 +73,7 @@ objects:
   metadata:
     name: "clusteroperator.openshift.io:controller-manager"
   rules:
-  # configmpas for leader election
+  # configmaps for leader election
   - apiGroups: [""]
     resources: ["configmaps"]
     verbs:     ["create"]
@@ -85,6 +85,13 @@ objects:
   - apiGroups: [""]
     resources: ["events"]
     verbs:     ["create","patch","update"]
+  # allow operations on required resources in our namespace
+  - apiGroups:     [""]
+    resources:     ["configmaps", "pods", "secrets"]
+    verbs: ["*"]
+  - apiGroups:     ["batch"]
+    resources:     ["jobs"]
+    verbs: ["*"]
   # allow all operations on all resources in our API group
   - apiGroups: ["clusteroperator.openshift.io"]
     resources: ["*"]
@@ -104,6 +111,19 @@ objects:
     kind: ServiceAccount
     name: cluster-operator-controller-manager
     namespace: ${CLUSTER_OPERATOR_NAMESPACE}
+
+# Allow our service account to run Ansible pods as UID 0:
+- kind: SecurityContextConstraints
+  apiVersion: v1
+  metadata:
+    name: cluster-operator-runasuser
+  allowPrivilegedContainer: true
+  runAsUser:
+    type: RunAsAny
+  seLinuxContext:
+    type: RunAsAny
+  users:
+  - system:serviceaccount:${CLUSTER_OPERATOR_NAMESPACE}:cluster-operator-controller-manager
 
 # Secret to pass the SSL certs to the API Server
 - apiVersion: v1
@@ -294,7 +314,7 @@ objects:
           - --contention-profiling
           - "false"
           - -v
-          - "10"
+          - "1"
           - --log-level
           - "debug"
           ports:
--- a/pkg/ansible/generate.go
+++ b/pkg/ansible/generate.go
@@ -1,0 +1,217 @@
+package ansible
+
+import (
+	"bytes"
+	"fmt"
+	"text/template"
+
+	coapi "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
+)
+
+// varsTemplate contains some hardcoded variables that are constant for all clusters,
+// as well as the go template annotations to substitute in values we do need to change.
+// Today this is AWS specific but will need to be broken down for multi-provider.
+const varsTemplate = `---
+# Variables that are commented in this file are optional; uncommented variables
+# are mandatory.
+
+# Default values for each variable are provided, as applicable.
+# Example values for mandatory variables are provided as a comment at the end
+# of the line.
+
+# ------------------------ #
+# Common/Cluster Variables #
+# ------------------------ #
+# Variables in this section affect all areas of the cluster
+ansible_ssh_user: centos
+ansible_become: true
+################################################################################
+# Ensure these variables are set for bootstrap
+################################################################################
+# Deployment type must be specified. ('origin' or 'openshift-enterprise')
+openshift_deployment_type: openshift-enterprise
+
+openshift_master_bootstrap_enabled: True
+
+openshift_hosted_router_wait: False
+openshift_hosted_registry_wait: False
+
+openshift_master_api_port: 443
+openshift_master_console_port: 443
+
+################################################################################
+# cluster specific settings
+################################################################################
+
+# openshift_release must be specified.  Use whatever version of openshift
+# that is supported by openshift-ansible that you wish.
+openshift_release: "v3.7" # v3.7
+
+# This will be dependent on the version provided by the yum repository
+openshift_pkg_version: -3.7.0 # -3.7.0
+
+# TODO: development specific
+openshift_disable_check: disk_availability,memory_availability,docker_storage,package_version,docker_image_availability
+openshift_repos_enable_testing: true
+
+# AWS region
+# This value will instruct the plays where all items should be created.
+# Multi-region deployments are not supported using these plays at this time.
+
+
+#openshift_aws_create_launch_config: true
+#openshift_aws_create_scale_group: true
+
+# --- #
+# VPC #
+# --- #
+
+# openshift_aws_create_vpc defaults to true.  If you don't wish to provision
+# a vpc, set this to false.
+openshift_aws_create_vpc: true
+
+# Name of the subnet in the vpc to use.  Needs to be set if using a pre-existing
+# vpc + subnet.
+#openshift_aws_subnet_name: cluster-engine-subnet-1
+
+# -------------- #
+# Security Group #
+# -------------- #
+
+# openshift_aws_create_security_groups defaults to true.  If you wish to use
+# an existing security group, set this to false.
+openshift_aws_create_security_groups: true
+
+# openshift_aws_build_ami_group is the name of the security group to build the
+# ami in.  This defaults to the value of openshift_aws_clusterid.
+#openshift_aws_build_ami_group: cluster-engine
+
+# openshift_aws_launch_config_security_groups specifies the security groups to
+# apply to the launch config.  The launch config security groups will be what
+# the cluster actually is deployed in.
+openshift_aws_launch_config_security_groups:
+  compute:
+  - cluster-engine
+  - cluster-engine_compute
+  - cluster-engine_compute_k8s
+  infra:
+  - cluster-engine
+  - cluster-engine_infra
+  - cluster-engine_infra_k8s
+  master:
+  - cluster-engine
+  - cluster-engine_master
+  - cluster-engine_master_k8s
+
+# openshift_aws_node_security_groups are created when
+# openshift_aws_create_security_groups is set to true.
+#openshift_aws_node_security_groups: see roles/openshift_aws/defaults/main.yml
+
+# -------- #
+# ssh keys #
+# -------- #
+
+# Specify the key pair name here to connect to the provisioned instances.  This
+# can be an existing key, or it can be one of the keys specified in
+# openshift_aws_users
+openshift_aws_ssh_key_name: libra # myuser_key
+
+# This will ensure these user and public keys are created.
+#openshift_aws_users:
+#- key_name: myuser_key
+#  username: myuser
+#  pub_key: |
+#         ssh-rsa AAAA
+
+# -- #
+# S3 #
+# -- #
+
+# Create an s3 bucket.
+openshift_aws_create_s3: false
+
+# --- #
+# ELB #
+# --- #
+
+# openshift_aws_elb_name will be the base-name of the ELBs.
+# TODO: looks like this is supposed to just be basename variant
+#openshift_aws_elb_name: dgoodwin-dev
+
+# custom certificates are required for the ELB
+openshift_aws_iam_cert_path: /home/abutcher/rhat/certs/wildcard-flibberty-jibbet.com.crt  # '/path/to/wildcard.<clusterid>.example.com.crt'
+openshift_aws_iam_cert_key_path: /home/abutcher/rhat/certs/wildcard-flibberty-jibbet.com.key  # '/path/to/wildcard.<clusterid>.example.com.key'
+openshift_aws_iam_cert_chain_path: /home/abutcher/rhat/certs/ca.crt  # '/path/to/cert.ca.crt'
+
+openshift_aws_create_iam_role: True
+openshift_node_use_instance_profiles: True
+
+[[ range $key, $value := . ]]
+[[ $key ]]: [[ $value ]]
+[[ end ]]
+`
+
+// TODO: these {{}} don't play nice in a go template, complete this when we get to processing
+// node groups dynamically
+/*
+# tags
+openshift_aws_master_group_tags:
+  Name: "{{ openshift_aws_clusterid }}-master"
+  host-type: master
+  sub-host-type: default
+
+openshift_aws_node_group_compute_tags:
+  Name: "{{ openshift_aws_clusterid }}-compute"
+  host-type: node
+  sub-host-type: compute
+
+openshift_aws_node_group_infra_tags:
+  Name: "{{ openshift_aws_clusterid }}-infra"
+  host-type: node
+  sub-host-type: infra
+*/
+
+const (
+	cloudProviderKindVar = "openshift_cloudprovider_kind"
+	awsClusterIDVar      = "openshift_aws_clusterid"
+	elbBasenameVar       = "openshift_aws_elb_basename"
+	awsVPCNameVar        = "openshift_aws_vpc_name"
+	awsRegionVar         = "openshift_aws_region"
+	awsAMIVar            = "openshift_aws_ami"
+)
+
+type VarsGenerator struct {
+	Vars map[string]string
+}
+
+func NewVarsGenerator(cluster *coapi.Cluster) *VarsGenerator {
+
+	// Any vars that change from cluster to cluster need to be defined here:
+	inv := &VarsGenerator{
+		Vars: map[string]string{
+			cloudProviderKindVar: "aws",
+			awsRegionVar:         "us-east-1",
+			awsClusterIDVar:      cluster.Name,
+			elbBasenameVar:       cluster.Name,
+			awsVPCNameVar:        cluster.Name,
+			awsAMIVar:            fmt.Sprintf("%s-ami-base", cluster.Name),
+		},
+	}
+	return inv
+}
+
+// GenerateVars returns a string suitable for use as a vars file when calling openshift-ansible pods.
+func (ig *VarsGenerator) GenerateVars() (string, error) {
+	// Change template delimiters to avoid conflict with {{ }} use in ansible vars:
+	t, err := template.New("varsfilecontents").Delims("[[", "]]").Parse(varsTemplate)
+	if err != nil {
+		return "", err
+	}
+
+	var buf bytes.Buffer
+	err = t.Execute(&buf, ig.Vars)
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
--- a/pkg/ansible/runner.go
+++ b/pkg/ansible/runner.go
@@ -1,0 +1,199 @@
+package ansible
+
+import (
+	"fmt"
+	"path"
+
+	log "github.com/sirupsen/logrus"
+
+	kbatch "k8s.io/api/batch/v1"
+	kapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	openshiftAnsibleImage          = "openshift/origin-ansible:v3.7"
+	openshiftAnsibleServiceAccount = "cluster-operator-controller-manager"
+	sshPrivateKeySecret            = "ssh-private-key"
+	operatorNamespace              = "cluster-operator"
+	openshiftAnsibleContainerDir   = "/usr/share/ansible/openshift-ansible/"
+	awsCredentialsSecret           = "aws-credentials"
+)
+
+type ansibleRunner struct {
+	KubeClient kubernetes.Interface
+	Image      string
+}
+
+func NewAnsibleRunner(kubeClient kubernetes.Interface) *ansibleRunner {
+	return &ansibleRunner{
+		KubeClient: kubeClient,
+		Image:      openshiftAnsibleImage,
+	}
+}
+func (r *ansibleRunner) createInventoryConfigMap(clusterName, jobPrefix, inventory, vars string) (*kapi.ConfigMap, error) {
+	log.Debugln("Creating inventory/vars ConfigMap")
+	genNamePrefix := fmt.Sprintf("%s%s-", jobPrefix, clusterName)
+	cfgmap := &kapi.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: genNamePrefix,
+		},
+		Data: map[string]string{
+			"hosts": inventory,
+			"vars":  vars,
+		},
+	}
+	cfgMap, err := r.KubeClient.CoreV1().ConfigMaps(operatorNamespace).Create(cfgmap)
+	if err != nil {
+		return cfgMap, err
+	}
+	log.WithField("configmap", cfgMap.Name).Infoln("created successfully")
+
+	return cfgMap, err
+}
+
+func (r *ansibleRunner) RunPlaybook(clusterName, jobPrefix, playbook, inventory, vars string) error {
+	logger := log.WithFields(log.Fields{
+		"cluster":  clusterName,
+		"playbook": playbook,
+	})
+	logger.Infoln("running ansible playbook")
+
+	// TODO: cleanup configmaps when their job is deleted
+	cfgMap, err := r.createInventoryConfigMap(clusterName, jobPrefix, inventory, vars)
+	if err != nil {
+		return err
+	}
+
+	env := []kapi.EnvVar{
+		{
+			Name:  "INVENTORY_FILE",
+			Value: "/ansible/inventory/hosts",
+		},
+		{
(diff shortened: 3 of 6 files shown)


pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+openshift_aws_iam_cert_chain_path: /home/abutcher/rhat/certs/ca.crt  # '/path/to/cert.ca.crt'
+
+openshift_aws_create_iam_role: True
+
+{{ range $key, $value := . }}
+   {{ $key }}: {{ $value }}
+{{ end }}
+`
+
+// TODO: these {{}} don't play nice in a go template, complete this when we get to processing
```
staebler: Since ansible uses {{ and }} as delimiters, we should use different delimiters for the go template.

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+[OSEv3:vars]
+
+[masters]
+
+[etcd]
+
+[nodes]
+`
+
+// varsTemplate defines all the data we consider unchangable for all clusters.
```
staebler: I find this comment confusing. The comment first talks about defining all the unchangeable data. Then the comment talks about how the data is changed.

Also, if we keep this comment, change `unchangable` to `unchangeable`.

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+	/*
+		var buffer bytes.Buffer
+
+		for k, v := range ig.Vars {
+			buffer.WriteString(fmt.Sprintf("%s=%s", k, v))
+		}
+
+		return fmt.Sprintf(invTemplate, buffer.String())
+	*/
+	t := template.Must(template.New("varsfilecontents").Parse(varsTemplate))
```
staebler: As `GenerateVars` returns an error to pass along problems, it would be more appropriate to return any error from `Parse` rather than panicking here.

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+	var buf bytes.Buffer
+	err := t.Execute(&buf, ig.Vars)
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+func (ig *VarsGenerator) GenerateInventory() string {
```
staebler: It seems strange to me that a type named `VarsGenerator` is generating inventory.

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+#openshift_aws_elb_name: dgoodwin-dev
+
+# custom certificates are required for the ELB
+openshift_aws_iam_cert_path: /home/abutcher/rhat/certs/wildcard-flibberty-jibbet.com.crt  # '/path/to/wildcard.<clusterid>.example.com.crt'
+openshift_aws_iam_cert_key_path: /home/abutcher/rhat/certs/wildcard-flibberty-jibbet.com.key  # '/path/to/wildcard.<clusterid>.example.com.key'
+openshift_aws_iam_cert_chain_path: /home/abutcher/rhat/certs/ca.crt  # '/path/to/cert.ca.crt'
+
+openshift_aws_create_iam_role: True
+
+{{ range $key, $value := . }}
```
staebler: Perhaps this is just an over-simplified example start, but using a template here seems to me to be a convoluted way of adding more variables at the end of a string. Do you anticipate that there will be actual variable replacement elsewhere in the template at some point?

pkg/controller/infra/infra_controller.go
```diff
@@ -233,8 +214,20 @@ func (c *InfraController) syncCluster(key string) error {
 	// TODO: Deep-copy only when needed.
 	cluster = cluster.DeepCopy()
 
-	c.logger.WithField("cluster", cluster.Name).Infoln("provisioning cluster")
-	// TODO
+	c.logger.WithField("cluster", cluster.Name).Infoln("provisioning cluster infrastructure")
+	ansibleRunner := ansible.NewAnsibleRunner(c.kubeClient)
+	inventoryGenerator := ansible.NewVarsGenerator(cluster)
```
staebler: We need some name consistency here. The type and the function use "vars generator". The variable uses "inventory generator". The type actually does both vars and inventory generation.

pkg/ansible/runner.go
```diff
@@ -0,0 +1,201 @@
...
+	return &ansibleRunner{
+		KubeClient: kubeClient,
+		Image:      openshiftAnsibleImage,
+	}
+}
+func (r *ansibleRunner) createInventoryConfigMap(inventory string, vars string) (*kapi.ConfigMap, error) {
+	log.Debugln("Creating inventory/vars ConfigMap")
+	cfgmap := &kapi.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: inventoryConfigMap,
```
staebler: Does each cluster need its own configmap? We can't use a constant name for the configmap, can we?

pkg/ansible/runner.go
```diff
@@ -0,0 +1,201 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	openshiftAnsibleImage          = "openshift/origin-ansible:v3.7"
+	openshiftAnsibleServiceAccount = "cluster-operator-controller-manager"
+	inventoryConfigMap             = "ansible-inventory"
+	sshPrivateKeySecret            = "ssh-private-key"
+	operatorNamespace              = "cluster-operator"
```
staebler: We can't hard-code this namespace. The user needs to be able to specify their own namespace in which to run cluster-operator.

pkg/ansible/runner.go
```diff
@@ -0,0 +1,201 @@
...
+					},
+					{
+						Name:      "sshkey",
+						MountPath: "/ansible/ssh/",
+					},
+				},
+				//Command: []string{"sleep", "1000000"},
+
+				// TODO: drop this once https://github.com/openshift/openshift-ansible/pull/6320 merges, the default run script should then work, minus the vars file :(
+				Command: []string{"ansible-playbook", "-i", "/ansible/inventory/hosts", "--private-key", "/ansible/ssh/privatekey.pem", playbookPath, "-e", "@/ansible/inventory/vars"},
```
staebler: Does ansible allow for multiple `-e` arguments? Would it be possible to pass in the unchangeable inventory in one file and the changeable in another? I don't know yet if that is any merit to that: I was just wondering if that is even a possibility.

pkg/ansible/runner.go
```diff
@@ -0,0 +1,201 @@
...
+		Data: map[string]string{
+			"hosts": inventory,
+			"vars":  vars,
+		},
+	}
+	cfgMap, err := r.KubeClient.CoreV1().ConfigMaps(operatorNamespace).Create(cfgmap)
+	if err != nil && kapierrors.IsAlreadyExists(err) {
+		// Update existing configmap if it already exists:
+		log.WithField("configmap", cfgMap.Name).Infoln("already exists, attempting update")
+		cfgMap, err = r.KubeClient.CoreV1().ConfigMaps(operatorNamespace).Update(cfgmap)
```
staebler: I would be very cautious about updating en existing configmap unless we can be absolutely sure that there are no pods still using the configmap.

pkg/ansible/runner.go
staebler: Now that I think about it more, I am leaning towards putting the.configmap and job into the namespace of the cluster anyway instead of the namespace of the controller manager. Are there reasons, such as ease of rbac, that make it better or easier if they are in the namespace of the controller manager?

contrib/examples/deploy.yaml
```diff
@@ -85,6 +85,13 @@ objects:
   - apiGroups: [""]
     resources: ["events"]
     verbs:     ["create","patch","update"]
+  # allow operations on required resources in our namespace
```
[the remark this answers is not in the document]
dgoodwin: Yes good catch, it looks like I should switch this to a role and rolebinding.

contrib/examples/deploy.yaml
```diff
@@ -294,7 +314,7 @@ objects:
           - --contention-profiling
           - "false"
           - -v
-          - "10"
+          - "1"
```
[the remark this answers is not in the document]
dgoodwin: Sorry this wasn't intentionally added to commit, but I do need to drop it down so I can watch what my controller is doing without the constant logging. If you feel 1 is too low for right now I can bump it back up and make do.

pkg/ansible/generate.go
dgoodwin: Will update it.

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+
+[etcd]
+
+[nodes]
+`
+
+// varsTemplate defines all the data we consider unchangable for all clusters.
+// Today this is AWS specific but will need to be broken down for multi-provider.
+// Vars that do change are calculated and stored on the generator, and templated in.
+const varsTemplate = `---
```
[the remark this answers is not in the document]
dgoodwin: Will give this some thought.

pkg/ansible/generate.go
dgoodwin: We're about to get to the more complicated stuff where we need a bunch of things per node group defined, some of them won't be simple string to string maps either. In all cases though they will still be appending items to the end of a string, I'm not sure which will play out easier though, the template looping might be handy when we get to:

```
openshift_aws_launch_config_security_groups:
  compute:
  - cluster-engine
  - cluster-engine_compute
  - cluster-engine_compute_k8s
  infra:
  - cluster-engine
  - cluster-engine_infra
  - cluster-engine_infra_k8s
  master:
  - cluster-engine
  - cluster-engine_master
- cluster-engine_master_k8s
```

pkg/ansible/generate.go
dgoodwin: Didn't think of that, will change this.

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+const (
+	cloudProviderKindVar = "openshift_cloudprovider_kind"
+	awsClusterIDVar      = "openshift_aws_clusterid"
+	elbBasenameVar       = "openshift_aws_elb_basename"
+	awsVPCNameVar        = "openshift_aws_vpc_name"
+	awsRegionVar         = "openshift_aws_region"
+	awsBaseAMINameVar    = "openshift_aws_base_ami_name"
+)
+
+type VarsGenerator struct {
```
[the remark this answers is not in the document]
dgoodwin: Yeah we'll need to start doing more complicated things when we get to the node groups and there will likely be some logic involved, we'll also need varying implementations for different cloud providers at some point.

pkg/ansible/generate.go
dgoodwin: Agreed, adjusting.

pkg/ansible/generate.go
dgoodwin: Yeah this actually had a TODO to remove it, it won't be there long term I'm just not sure how inventory will play out. I suspect for provisioning we'll just use a flat empty inventory string like above, and per last discussion we'll use something like ec2.py in container to calculate the real one on the fly.

pkg/ansible/runner.go
dgoodwin: Agreed we'll need some way to configure this. Any thoughts on how? My last controller with online team I was using a configmap and mounting in as if it were a config file in the pod.

Are you envisioning that every cluster the CO manages gets it's own namespace? I was envisioning the opposite, that all data for all manages CO clusters lives in the cluster-operator namespace. I don't think we've ever actually discussed one way or the other.

One namespace per managed cluster has some implications for deployment, it would largely assume the cluster we run in is used only for cluster-operator I think, otherwise we might have some issues with namespace clutter and collisions. We might be able to swing that for our purposes but it could cause problems if we ever get to on-premise customer use cases. (a topic we all need to talk a lot more about to understand anyhow)

I'm not real familiar with pros/cons of either approach, will need to think on this and maybe we can discuss with team next chance we get.

pkg/ansible/runner.go
dgoodwin: Absolutely not, I think each execution of an ansible job will need it's own. Haven't gotten this far yet, but my plan was to pre-create the config map (obviously), label it with something indicating it's for a particular job (maybe pre-generate random job names), and then add a controller that would watch configmaps with that label, and cleanup any whose job no longer exists. (open to better ideas if anyone has any)

pkg/ansible/runner.go
dgoodwin: Yeah this update block goes away once we're on one configmap per job.

pkg/ansible/runner.go
dgoodwin: I would definitely expect that it would. I'm not sure which would be easier but it wouldn't be hard to do if we want to go that route.

pkg/controller/infra/infra_controller.go
dgoodwin: inventory generator is a relic of where I originally thought this was going, will adjust.

dgoodwin:
Also don't worry about it, that's great review, thanks!

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+openshift_master_api_port: 443
+openshift_master_console_port: 443
+
+################################################################################
+# cluster specific settings
+################################################################################
+
+# openshift_release must be specified.  Use whatever version of openshift
+# that is supported by openshift-ansible that you wish.
+openshift_release: "3.7" # v3.7
```
csrwng: 3.7 or v3.7 ?

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+# --- #
+# ELB #
+# --- #
+
+# openshift_aws_elb_name will be the base-name of the ELBs.
+# TODO: looks like this is supposed to just be basename variant
+#openshift_aws_elb_name: dgoodwin-dev
+
+# custom certificates are required for the ELB
+openshift_aws_iam_cert_path: /home/abutcher/rhat/certs/wildcard-flibberty-jibbet.com.crt  # '/path/to/wildcard.<clusterid>.example.com.crt'
```
csrwng: are these getting mounted via secret/configmap?

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+# AMI Build #
+# --------- #
+# Variables in this section apply to building a node AMI for use in your
+# openshift cluster.
+
+# must specify a base_ami when building an AMI
+openshift_aws_base_ami: ami-ae7bfdb8 # ami-12345678
+
+# this is the ami we built
+openshift_aws_ami: ami-9a0d95e0 # ami-2f53ce55
```
csrwng: should this be hardcoded? (seems like it should be generated)

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+
+	// Any vars that change from cluster to cluster need to be defined here:
+	inv := &VarsGenerator{
+		Vars: map[string]string{
+			cloudProviderKindVar: "aws",
+			awsRegionVar:         "us-east-1",
+			awsClusterIDVar:      cluster.Name,
+			elbBasenameVar:       cluster.Name,
+			awsVPCNameVar:        cluster.Name,
+			awsBaseAMINameVar:    fmt.Sprintf("%s-ami-base", cluster.Name),
```
csrwng: I would have thought that you'd specify the `openshift_aws_ami` var and not the `openshift_aws_base_ami_name`

pkg/ansible/runner.go
csrwng: If the ansible container is required to run as root, one important consideration is that wherever it runs, it needs to have access to a service account that has the runAsAny SCC. Right now that is only the cluster-operator namespace. 

In my opinion, if this is something that will be run by the controller it's acceptable to run it in a special namespace where you have the service account you need. On the other hand, from an accounting/resource usage perspective, it'd be nice to track the resource usage of this job with the rest of the cluster resources for the user (in the namespace where those cluster resources live). Maybe it's worth investigating whether we really do need to run the ansible container as root instead of a random user.

pkg/ansible/runner.go
csrwng: You can simply switch to a GenerateName and use the constant as a prefix. It may also be useful to add a label to identify what cluster this configmap belongs to so that you can identify cluster-related resources later.

pkg/ansible/runner.go
```diff
@@ -0,0 +1,201 @@
...
+			},
+		},
+	}
+
+	completions := int32(1)
+	deadline := int64(60 * 60) // one hour for now
+
+	meta := metav1.ObjectMeta{
+		GenerateName: jobNamePrefix,
+		Namespace:    operatorNamespace,
```
csrwng: It may be useful to add labels to correlate the job to a cluster resource.

pkg/ansible/runner.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	job := &kbatch.Job{
```
csrwng: I'm wondering whether a job is the most appropriate resource here. The job controller will re-run the job blindly until it either succeeds or the completion deadline expires. I'm wondering if we'd want to use our own controller to manage the job (and implement things like exponential backoff and fixed number of retries, etc). If that's the case, then it's better off as a run-once pod.

contrib/examples/deploy.yaml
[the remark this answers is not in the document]
staebler: We can leave it at 1. I'll refrain from checking in changes to bump it back up in future.

pkg/ansible/runner.go
staebler: Agree with that. I would expect this name to be something like "someconstant-nameofcluster-somerandom".

pkg/ansible/runner.go
staebler: Definitely. We will some one to associate the jobs with the cluster for no other reason than ensuring that we are cleaning everything up when the cluster is deleted. @csrwng, is there a preference on using labels versus using owner references? For ClusterController, I set the Cluster as the controlling owner of each NodeGroup created for the Cluster.

csrwng: @staebler it will depend on whether the jobs/configmaps will live in the same namespace as the cluster resource or not. An owner ref is definitely preferable, but I don't think you can have owner refs across namespaces.

pkg/ansible/runner.go
staebler: With the understanding that I have no experience with jobs, they seem to me to be the appropriate thing to use here. My understanding is that the job controller already does exponential backoff of failed jobs. I don't see much value in having a fixed number of retries over a deadline for retries.

pkg/ansible/runner.go
staebler: I'm still hoping that we can have all the resources needed to stand up a cluster created in the same namespace as the cluster.

pkg/ansible/generate.go
dgoodwin: Both should work but I'll flip to v.

pkg/ansible/generate.go
dgoodwin: Not long term, I'm not 100% sure how we'll know which is the latest / correct AMI to use just yet.

pkg/ansible/generate.go
dgoodwin: Not yet, we haven't gotten to the point where they're used, and we're not setting them up yet. A secret mounted in would seem like a decent option.

pkg/ansible/runner.go
csrwng: See my comment about the service account. It'd be great if we don't need a runAsAny SCC

pkg/ansible/runner.go
csrwng: Fair point

pkg/ansible/runner.go
dgoodwin: Any thoughts @abutcher ? I tried and got into a problem with ansible erroring out trying to look up the user identity for an obscure UID which looked like it was the fsGroupID for the namespace.

pkg/ansible/runner.go
dgoodwin: Yeah we did plan to have this thing re-try until completion so the Job felt like a good fit, at least for now. I'd pitch we stick with Job until it starts to fail us.

contrib/examples/deploy.yaml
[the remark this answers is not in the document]
dgoodwin: Hmm based on our discussion this morning, we probably do now need this for all namespaces don't we.

pkg/ansible/runner.go
csrwng: @dgoodwin in the scl containers that require a valid username, we use libnss_wrapper to trick the main process into thinking that the users exists in passwd:
https://github.com/sclorg/postgresql-container/blob/master/9.5/root/usr/share/container-scripts/postgresql/common.sh#L137-L145

pkg/ansible/generate.go
```diff
@@ -0,0 +1,257 @@
...
+# openshift_aws_elb_name will be the base-name of the ELBs.
+# TODO: looks like this is supposed to just be basename variant
+#openshift_aws_elb_name: dgoodwin-dev
+
+# custom certificates are required for the ELB
+openshift_aws_iam_cert_path: /home/abutcher/rhat/certs/wildcard-flibberty-jibbet.com.crt  # '/path/to/wildcard.<clusterid>.example.com.crt'
+openshift_aws_iam_cert_key_path: /home/abutcher/rhat/certs/wildcard-flibberty-jibbet.com.key  # '/path/to/wildcard.<clusterid>.example.com.key'
+openshift_aws_iam_cert_chain_path: /home/abutcher/rhat/certs/ca.crt  # '/path/to/cert.ca.crt'
+
+openshift_aws_create_iam_role: True
```
joelddiaz: From my experience yesterday, I think you'll also want to define openshift_node_use_instance_profiles: True so that the infra and compute scale groups/launch configs get enough permissions to query instances.

pkg/ansible/runner.go
abutcher: I think we should try to run as non-root if we can. I'm digging into the current image now to see what it's up to.

pkg/ansible/generate.go
dgoodwin: Added thanks!

dgoodwin:
Ok I've pushed fixes for most of the low hanging fruit, I have filed several takeaways from review for followup PRs. (drop need to run as root, integration/unit testing, and a few other items)

Would like to merge in this state, it will echo the playbook called and no-op unless you uncomment the line above that in runner.go. Will send full instructions to mailing list.

Could I get another round of review on what's here and see if we can get this in to unblock some folks.

contrib/examples/deploy.yaml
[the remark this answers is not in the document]
staebler: My remark was directed more at the comment that states "in our namespace" whereas the permissions are actually being granted for in all namespaces. I was suggesting that the comment should be changed, although I was not clear about that.

pkg/ansible/runner.go
```diff
@@ -0,0 +1,208 @@
...
+
+func NewAnsibleRunner(kubeClient kubernetes.Interface) *ansibleRunner {
+	return &ansibleRunner{
+		KubeClient: kubeClient,
+		Image:      openshiftAnsibleImage,
+	}
+}
+func (r *ansibleRunner) createInventoryConfigMap(clusterName, inventory, vars string) (*kapi.ConfigMap, error) {
+	log.Debugln("Creating inventory/vars ConfigMap")
+	genNamePrefix := fmt.Sprintf("%s%s-", namePrefix, clusterName)
```
staebler: I would like to see some additional label added to this name that identifies which specific ansible job is being run. I don't know that the "ansible" portion is necessary. For example, "infra-provision-job-myCluster-" It is fine for now, though.

pkg/ansible/runner.go
```diff
@@ -0,0 +1,208 @@
...
+	deadline := int64(60 * 60) // one hour for now
+
+	// TODO: right now we get a configmap names ansible-job-cluster-X, and a job
+	// named ansible-job-cluster-Y. Find a good way to link these two randomly named
+	// resources. Note that jobs implicitly seem to get a random prefix, even if you
+	// try to set the name directly.
+	genNamePrefix := fmt.Sprintf("%s%s-", namePrefix, clusterName)
+	meta := metav1.ObjectMeta{
+		GenerateName: genNamePrefix,
+		Namespace:    operatorNamespace,
```
staebler: It is not necessary to set the namespace here. The namespace will be set by the API server.

pkg/ansible/runner.go
```diff
@@ -0,0 +1,208 @@
...
+			},
+		},
+	}
+
+	completions := int32(1)
+	deadline := int64(60 * 60) // one hour for now
+
+	// TODO: right now we get a configmap names ansible-job-cluster-X, and a job
+	// named ansible-job-cluster-Y. Find a good way to link these two randomly named
+	// resources. Note that jobs implicitly seem to get a random prefix, even if you
```
staebler: The job is getting a random *prefix* or a random *suffix*? I don't know why the API server would be overwriting the name that you supply for the job at all, though. Are you looking at the job or the pod created to service the job?

dgoodwin: Oh man that might be what I did, I will re-test.

dgoodwin: Yup that was it. I'm pushing a fix to resume re-using the configmaps random name.

pkg/ansible/runner.go
csrwng: Because we're creating these resources in a single namespace for now, shouldn't the name also include the cluster's namespace?

dgoodwin: Updated a little to job-infra-clusterid-RANDOM, and now used on both configmap and job exactly.

dgoodwin: It includes the clusters ID, right now the cluster doesn't have a namespace, I do have a todo on my trello card to test multi-namespace shortly.

staebler: What do you mean by "the cluster doesn't have a namespace"? The cluster certainly has a namespace. Also, are the terms ID and Name interchangeable in the kube world? Or does the ID of a resource refer to the UID? Or is the term ID not used at all?

eparis: "ID" is not a kube term of art that I am familiar with. We have metadata.uid and metadata.name.

dgoodwin: I meant that there isn't a namespace being created per cluster by anything per se, the cluster object itself would land in one when created for sure. I'll send in a follow-up that re-uses that namespace for the configmap and job, which is probably what Cesar meant for me to do.

Cluster ID is coming from my ops rotation, just a prevalent concept in all their ansible, I'll try to remember to stick to Name. Cluster ID on their side is just a short label for each cluster such as "free-int" or "online-stg", well suited for a kube name.

dgoodwin:
Added some notes on how to test provisioning in it's current state, multi-ns coming shortly in separate PR.

Issue #2, opened 2017-12-05
controllers are skipped when controller-manager starts before CO apiserver is ready


staebler:
When controller-manager starts a controller, it checks to see that the resource is available from the api server. If the CO apiserver is not ready yet when the check is performed, the controller-manager ends up skipping the controller, so it is never started.

staebler:
Add a wait.PollImmediate [here](https://github.com/openshift/cluster-operator/blob/577652adc554319431122e552ee0d67d87041f84/cmd/cluster-operator-controller-manager/app/controllermanager.go#L345) to retry for a bit if discoveryClient.ServerResources returns an error.

PR #21, opened 2017-12-12
WIP - Proposal for Cluster API

csrwng:


csrwng:
@dgoodwin @staebler @joelddiaz @abutcher ptal

docs/cluster_api.md
```diff
@@ -0,0 +1,106 @@
...
+For the Status part, if the cluster has been instantiated, we need to know how to access it (Kubeconfig may be sufficient)
+
+In more detail:
+
+#### Spec:
+* Cluster-wide configuration:
+  * Name
+  * Hardware-spec
+    * Type (AWS, Azure, GCE, etc)
+    * AWSConfig:
```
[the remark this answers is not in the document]
csrwng: The intent is that we'd have an optional struct for each provider as we add support for them

docs/cluster_api.md
```diff
@@ -0,0 +1,106 @@
...
+      - AWSSecret
+      - Region
+      - VPCName
+      - VPCSubnet
```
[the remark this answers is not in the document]
csrwng: If the per-zone breakdown is something we can derive automatically for AWS, I'd rather not have that as part of the API just yet. The comment ~ line 122 of http://pastebin.test.redhat.com/536174 seems to indicate so.

docs/cluster_api.md
```diff
@@ -0,0 +1,106 @@
...
+  * OpenShift Configuration:
+    * DeploymentType
+    * Version of OpenShift to install:
```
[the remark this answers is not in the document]
csrwng: Good question. I'd say we start with the simplest option, will switch to a single string.

docs/cluster_api.md
```diff
@@ -0,0 +1,106 @@
...
+      - RPM URL
+      - Image Template
+    * Public hostname
+    * SDN Plugin Name
+    * Service Network CIDR
+    * Pod Network CIDR
```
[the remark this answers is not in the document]
csrwng: My understanding is that they're separate (and virtual), but will defer to folks with more expertise on openshift-ansible.

docs/cluster_api.md
```diff
@@ -0,0 +1,106 @@
...
+    * Variables
+* Nodeset configuration (array):
```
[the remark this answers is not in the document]
csrwng: No, should be NodeGroup

docs/cluster_api.md
```diff
@@ -0,0 +1,106 @@
...
+      - AMI Name
+      - Additional Hardware Details? (Additional volumes, for instance)
+  - OpenShift Configuration:
+    - Node Labels
+    - Variables [Array of AnsibleVariable]
+
+#### Status:
+* Access Information
+  - KubeConfigSecret (LocalReference)
+* Conditions
```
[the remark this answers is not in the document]
csrwng: Thanks for the heads up. Will follow up with folks from the master team

docs/cluster_api.md
```diff
@@ -0,0 +1,106 @@
...
+  - Type (HardwareReady, MasterReady, etc.)
+  - Status (True/False)
+  - LastProbeTime
+  - LastTransitionTime
+  - Message
+  - Reason
+* NodeGroupCount
```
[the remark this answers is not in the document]
csrwng: The count of current nodes available on the nodegroup. See https://github.com/openshift/cluster-operator/blob/master/pkg/apis/clusteroperator/v1alpha1/types.go#L60-L66
But it really should be an array of structs with name, count as members.

docs/cluster_api.md
```diff
@@ -0,0 +1,106 @@
...
+  - Type (Provisioning, Resizing, Ready)
+  - Status (True/False)
+  - LastProbeTime
+  - LastTransitionTime
+  - Message
+  - Reason
+* Phase
+
+
+### AnsibleVariable
```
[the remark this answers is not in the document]
csrwng: It doesn't need to be called 'Ansible'. However, it was a way to allow you to specify additional one-off settings that we don't support directly in the API when we construct the inventory. We'd have a list of them for the cluster scope and for the nodegroup scope. If we think that it's not useful/helpful to have this out from the start, I'll gladly remove.

docs/cluster_api.md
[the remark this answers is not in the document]
abutcher: Will this be the specific version or the release, like `openshift_release="v3.7"`? If we went with the release I think it'd be reasonable to set specific version variables (`openshift_pkg_version`, `openshift_image_tag`) in the `AnsibleVariable` or more variable section.

csrwng:
Made updates and added types for clarity

docs/cluster_api.md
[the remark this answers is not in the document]
dgoodwin: I've vote for remove for now and we can re-evaluate if necessary. I'm hoping that'll keep us from coupling ourselves too closely to openshift-ansible internals and the API more generic.

docs/cluster_api.md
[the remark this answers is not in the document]
abutcher: These are the SDN cluster network and kubernetes service CIDR blocks that will end up in the master config `https://github.com/openshift/openshift-ansible/blob/master/inventory/hosts.example#L709-L728`.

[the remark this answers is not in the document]
abutcher: Note: openshift-ansible does not support changing these values with re-runs currently.

docs/cluster_api.md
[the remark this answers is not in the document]
csrwng: Removed

csrwng:
Here's a list of my references so far this API:
- Current code: https://github.com/openshift/cluster-operator/blob/master/pkg/apis/clusteroperator/v1alpha1/types.go
- Kubernetes Cluster Registry: https://github.com/kubernetes/cluster-registry/blob/master/pkg/apis/clusterregistry/v1alpha1/types.go
- Kubicorn - https://github.com/kris-nova/kubicorn/blob/master/apis/cluster/cluster.go
- Kops - https://github.com/kubernetes/kops/blob/master/pkg/apis/kops/cluster.go
- Input to ops cluster: http://pastebin.test.redhat.com/536174
- OpenShift ansible install docs: https://docs.openshift.org/latest/install_config/install/advanced_install.html

docs/cluster_api.md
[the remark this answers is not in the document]
staebler: It shouldn't be an array of structs. It *could* be an array of names of the node groups if we care to track it with that level of specificity in the cluster status. I would argue that a count is sufficient, however. It is easy enough to look at the NodeGroup resources to get details about which specific ones have been created and which still need to be created. In practice, I would expect that the count is almost immediately transitioned to its final value when a Cluster resource is created. Most everything that would prevent a NodeGroup resource from being successfully created should be captured in the API validation for the Cluster resource itself.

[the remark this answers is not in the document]
csrwng: @staebler so if I understand correctly, this count will simply keep track of the number of nodegroups created for the cluster and nothing more? If so, I agree that a simple count is sufficient.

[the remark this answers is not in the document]
csrwng: Will update, unless there's any objections.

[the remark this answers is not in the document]
staebler: @csrwng Yes. That is correct. All the count reflects is the number of NodeGroup resources that have been created. I am having a hard time coming up with a good reason to even have that, if we had a status condition that reflected that the NodeGroup resources have all been created.

csrwng:
Regarding the use of conditions, this seems to be the latest summary from Brian Grant:
https://github.com/kubernetes/kubernetes/issues/7856#issuecomment-335687733

In a nutshell, it seems we still want to use conditions for extended status, but use single fields to report common status (such as Ready).

Recent PRs are still adding conditions to existing resources:
https://github.com/kubernetes/kubernetes/pull/55272
https://github.com/kubernetes/kubernetes/pull/55268

So I don't think we should leave them out of our API if we want to express extended state. However, based on the discussion from Brian Grant's issue, we should likely remove the Phase field, and replace with simple fields that we can use to reflect the most frequently queried state.

/cc @dgoodwin @staebler

csrwng:
I would like to get some more input from folks with ops and openshift-ansible experience. Particularly these questions:

1 - Are the fields we are using to define virtual hardware and cluster config sufficient
2 - Regarding the VPC subnet, Is it possible to derive the per-zone subnets (us-east-1a, 1b, etc) given a single subnet, or should we force you to specify them?

/cc @twiest @eparis

dgoodwin:
Sounds good to me regarding conditions and status, thanks for looking into that.

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+* Cluster-wide configuration:
+  * ObjectMetadata (inline ObjectMetadata)
+  * TypeMetadata (inline TypeMetadata)
+  * Hardware-spec
+    * Type (AWS, Azure, GCE, etc) [string]
+    * AWSConfig: [struct]
+      - AWSSecret [LocalReference]
+      - Region [string]
+      - VPCName [string]
+      - VPCSubnet [string]
```
joelddiaz: I'm told that one of these days we'll automate building a multi-AZ clusters, so we likely want VPCSubnets (plural) so we can have an AZ-aware list of subnets that will be attached to the VPC

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+### NodeGroup
+
+#### Spec
+- Type (Master/Compute) [string]
+- Infra [boolean]
+- Count [unsigned int]
+- Hardware spec (cloud-specific):
+  - AWSConfig:
+    - InstanceType [string]
+    - AMIName [string]
```
joelddiaz: related to multi-AZ/VPCs/subnets, will need a field here to hold the subnet the NodeGroup will be a part of

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+  * Hardware-spec
+    * Type (AWS, Azure, GCE, etc) [string]
+    * AWSConfig: [struct]
+      - AWSSecret [LocalReference]
+      - Region [string]
+      - VPCName [string]
+      - VPCSubnet [string]
+  * OpenShift Configuration:
+    * DeploymentType [string]
+    * OpenShiftVersion [string]
```
joelddiaz: Will reving the OpenShiftVersion field (ie 3.6 to 3.7) cause the nodegroups to point to a new AMI? Or will an upgrade simply be updating the AMI on the nodegroup(s).

csrwng: My understanding is that this is more for the use of the master. For nodes themselves, yes we would just change the AMI

docs/cluster_api.md
csrwng: thx @joelddiaz. I guess that means we can't derive the AZ subnets from the main VPC ? (this made me think we could http://pastebin.test.redhat.com/541799 - see highlighted). If not, I guess we want a map of AZ->subnet, correct?

docs/cluster_api.md
csrwng: Thx, will add

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+    * Public hostname [string]
+    * SDN Plugin Name [string]
+    * Service Network CIDR [string]
+    * Pod Network CIDR [string]
+    * Variables [array of Variable struct]
```
smarterclayton: Is variables an escape hatch?

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+For the Status part, if the cluster has been instantiated, we need to know how to access it (Kubeconfig may be sufficient)
+
+In more detail:
+
+#### Spec:
+* Cluster-wide configuration:
+  * ObjectMetadata (inline ObjectMetadata)
+  * TypeMetadata (inline TypeMetadata)
+  * Hardware-spec
+    * Type (AWS, Azure, GCE, etc) [string]
```
smarterclayton: We already have an "API" as defined by the variables that openshift-ansible accepts.  You will then have to transform this API into that API before openshift-ansible will run.  Why not define your API *exactly* as the the openshift-ansible variables would be, and formalize those more strongly?  

Articulate how having these spec variables is going to solve a user problem (you have no justification here for why), vs creating more work and maintenance.  Are any of these values going to be changeable post create?  Which ones?  We put things in API fields so that end users can set them.  Which ones of these will be set?

Is this API going to be the only source of truth?  If not, that's a third API you will be converting from and to (which is probably at least 1 too many).  If it is the source of truth, do you expect to support being able to create and recreate this entity and deal with how changes would prevent it from running?

docs/cluster_api.md
smarterclayton: Who is going to update the AMI name?  What sort of user will be setting the AMI field besides the deployment automation?  How does a user look up the correct AMI name?

docs/cluster_api.md
joelddiaz: We pretty much carve up the subnets however we want. In practice we have a defined list of regions/subnet-lists that we use in case the specific cluster doesn't have any particular requirements, but in the case where we need to VPC pair with a customer's existing VPC, we massage these values as needed.

docs/cluster_api.md
csrwng: @smarterclayton we need to have a mechanism to discover it, whether it's based on querying AWS for AMI tags or using a configmap. Are you saying that because it's not something end-users would know, that we should not expose it?

docs/cluster_api.md
csrwng: It is and should have been removed.

docs/cluster_api.md
csrwng: > We already have an "API" as defined by the variables that openshift-ansible accepts. You will then have to transform this API into that API before openshift-ansible will run. Why not define your API exactly as the the openshift-ansible variables would be, and formalize those more strongly?

We need to have separate identifiable entities that we can work with (a cluster, machine set, etc). The ansible API is a flat list of variables, not exactly sure what you mean by having them match exactly. I agree we should avoid inventing new things so things are harder to maintain, and I don't think there's anything wildly different here than what you can specify in ansible.

csrwng: > Articulate how having these spec variables is going to solve a user problem (you have no justification here for why), vs creating more work and maintenance. Are any of these values going to be changeable post create? Which ones? We put things in API fields so that end users can set them. Which ones of these will be set?

I can clarify a little bit more, but the basic idea is that the spec portion is end-user changeable and to "make it so" we will do what's necessary, including spinning down your cluster and bring a new one up.

csrwng: > Is this API going to be the only source of truth?

For clusters managed by this API, yes, that's the intent.

eparis: I would pretty strongly disagree that the openshift-ansible variable set should be the 'API'. We are likely going to need to merge this and the upstream cluster api. So while we need to recognize what openshift-ansible is asking for we need to also recognize that in the long run we will likely need to do translations. So making this API seem as sane as possible will be the most useful as we try to reconcile with upstream.

smarterclayton: I'm not saying it should be "the" API.  I'm saying it is our existing provision API.  It already has to be stable for customers.  Because it must be stable for customers, it should also be stable for other consumers like cluster-operator.  Cluster operator is mapping part of the provision API up a level.  Like any "abstraction" API there will be constant urge to expose things up from the lower level API (i.e. node tuning parameters), and the more delta there is between how the ansible API is designed and how the cluster API is designed there will be more bugs, confusion, documentation, and human errors, which result in tech debt, failures in customer environments, and unhappy users.

So while you may have to align with upstream, you also need to reduce the amount of conceptual delta between ansible and the operator, force ansible to declare and support stable APIs for you, and test those APIs.  Even the naming and structure is important - what unit does ansible accept, how about nils, what happens if the value is unrecognized.  It's insufficient to design this API without understanding the API we expose today, even if we wish to change it.

smarterclayton: Re: layers of abstraction, with cluster operator here are the layers:

1. cluster api (source of truth)
2. ansible via install and update
3. master-config, node-config
4. kube component flags and config

Customers will continue to use #2, and we are going to work at and evolve collapsing 3-4 (although our approach may change).  It's really important that when you have to expose something up from 3-4 through 2 to 1, that the mapping be as simple as possible.

smarterclayton: > I can clarify a little bit more, but the basic idea is that the spec portion is end-user changeable and to "make it so" we will do what's necessary, including spinning down your cluster and bring a new one up.

Talked with cesar in person, but it's extremely unlikely any sort of core cloud configuration can be changed.  None of your IP addresses are going to be able to move, for one thing, and data PV transfer is far more complex.  Probably better to just say - certain fields that are fundamental are not changeable by default, and you identify the fields that *can* change with good use cases.  I don't care what the use cases are (well, I do), but make sure you are justifying *why* they can change.

docs/cluster_api.md
smarterclayton: This is along the lines of the AMI comment, but the code you deploy with needs to be tied together.  There are two cases:

1. use binaries, images, and AMIs that come from a concrete release tag
2. use binaries, images, and AMIs that come from arbitrary locations (PR jobs, for example)

The API consumer needs to be able to say "upgrade to v3.10.1" but the test infrastructure needs to be able to say "upgrade to v3.11.0-prerelease.pr-19034 which comes from here, here, and here".  That's probably calling out that you may be missing another construct like a "Version" (which contains AMI names, image names, etc) and some API callers can override those on the field.

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+      - Region [string]
+      - VPCName [string]
+      - VPCSubnet [string]
+  * OpenShift Configuration:
+    * DeploymentType [string]
+    * OpenShiftVersion [string]
+    * Public hostname [string]
+    * SDN Plugin Name [string]
+    * Service Network CIDR [string]
+    * Pod Network CIDR [string]
```
eparis: pod network CIDR is something best defined inside and sdn specific configuration block I believe. As different SDNs can have very different pod network CIDR abilities/restrictions.

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+      - AWSSecret [LocalReference]
+      - Region [string]
+      - VPCName [string]
+      - VPCSubnet [string]
+  * OpenShift Configuration:
+    * DeploymentType [string]
+    * OpenShiftVersion [string]
+    * Public hostname [string]
+    * SDN Plugin Name [string]
+    * Service Network CIDR [string]
```
eparis: I think this needs to be an array of strings. Even if we only allow len(1) to start.

docs/cluster_api.md
smarterclayton: I should either be setting v3.7.9 on the master and v3.7.6 on the node, or I should be setting direct AMI values and direct binaries, but not both.  This API will be driven from automation that creates these artifacts and then expects to spin up a cluster with them - whether that's a "Version" API object that you always reference by name that contains the artifacts, or a struct that takes a version ref or concrete value depending on what sort of use case it is.

Humans selecting that value aren't going to get the right AMI, and machines  probably want to be direct.

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+    * Type (AWS, Azure, GCE, etc) [string]
+    * AWSConfig: [struct]
+      - AWSSecret [LocalReference]
+      - Region [string]
+      - VPCName [string]
+      - VPCSubnet [string]
+  * OpenShift Configuration:
+    * DeploymentType [string]
+    * OpenShiftVersion [string]
+    * Public hostname [string]
```
eparis: What is public hostname? Do clusters have a 'public hostname' ?

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+    * SDN Plugin Name [string]
+    * Service Network CIDR [string]
+    * Pod Network CIDR [string]
+    * Variables [array of Variable struct]
+* NodeGroups [array of NodeGroupSpec]:
+  - Type: (Master, Compute) [string]
```
eparis: I'm not sure why these defs are here and down below. Can we just have the 'down below'?

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+* NodeGroupCount [unsigned int]
+* Provisioned [boolean]
+* Running [boolean]
+
+
+### NodeGroup
+
+#### Spec
+- Type (Master/Compute) [string]
+- Infra [boolean]
```
eparis: Why is 'infra' magic? Isn't 'infra' just a NodeLabel?

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+  - Hardware spec (cloud-specific):
+    - AWSConfig:
+      - InstanceType [string]
+      - AMIName [string]
+  - OpenShift Configuration:
+    - NodeLabels [array of String]
+
+#### Status:
+* Access
+  - KubeConfigSecret [LocalReference]
```
eparis: Is this just `system:admin`? (I'm not really sure what [LocalReference] means.) Is `KubeConfigSecret` the most descriptive name?

Are there other access types we will want to expose here? I guess those can be added later without a problem.

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+* Conditions [array]
+  - Type (HardwareReady, MasterReady, etc.) [string]
+  - Status (True/False/Unknown) [string]
+  - LastProbeTime [string]
+  - LastTransitionTime [string]
+  - Message [string]
+  - Reason [string]
+* NodeGroupCount [unsigned int]
```
eparis: I'm not sure what this is. Is it some summary of each of the NodeGroups statuses?

docs/cluster_api.md
```diff
@@ -0,0 +1,95 @@
...
+* Provisioned [boolean]
+* Running [boolean]
```
eparis: I'm not sure what `provisioned` and `running` are telling me either. Is it somehow related to `Conditions`?

docs/cluster_api.md
eparis: ummmm, how can you define a nodegroup without defining the AMI?

I also don't believe one would ever update the AMI of a nodegroup. They would make a new node group and roll between them. Updating the AMI is like updating the image in a replicaset. You don't. You update the deployment and the deployment makes a new replicaset.

eparis: Sounds like clayton wants an image stream for AMIs. In any case, it's a higher level construct then belongs in the NodeGroup.spec.  NodeGroup.spec should be a very specific AMI on AWS.

docs/cluster_api.md
eparis: What is 'deployment type' ?
Do we really want to call this out as a 'master config'?
Could we maybe declare 'containerized only' for the master components? Then 'version' is really 'image name'. Which should let us write higher level controllers to do anything smart around upgrading. It also starts to treat masters a bit like we're eventually going to treat the router/registry/et al.

dgoodwin: Deployment type per openshift ansible is essentially "origin" or "openshift-enterprise". (I believe that is what was intended here)

Containerized masters would be interesting, makes me wonder if we should target Atomic as well, and where things are with the recent discussion around system containers + static pods/daemonsets. I'll hunt down some info on what we should target for this.

dgoodwin: This also got me thinking about the conversation awhile back on moving towards system containers for the essentials (kubelet and container runtime) and then static pods or daemon sets for everything else. Checked in with Scott this morning and it sounds like that's still the plan for the 3.10 or 3.11 timeframe so we should keep that containers first approach in mind, even though the golden images today are entirely RPM based.

Upstream cluster API already has a version struct differentiating kubelet/control plane/docker version, we'll need to think about the same for OpenShift and cri-o.

csrwng: > What is 'deployment type'

it's this variable: https://github.com/openshift/openshift-ansible/blob/master/inventory/hosts.example#L30-L31

docs/cluster_api.md
csrwng: Yup, will consolidate. In code, I've defined a MachineSetConfig that is shared between the master and the machineset/nodegroup

docs/cluster_api.md
csrwng: The words in brackets are the type of the variable. It's a kube corev1.LocalReference because it refers to a resource in the same namespace.

docs/cluster_api.md
csrwng: Just a count of how many nodegroups have been actually created for the cluster resource.

docs/cluster_api.md
csrwng: They are simple boolean status telling you whether 1) the hardware has been provisioned and 2) the cluster is up and running.

This is meant to be used for commonly queried status (vs more detailed status in the conditions), as suggested here: kubernetes/kubernetes#7856 (comment)

docs/cluster_api.md
csrwng: It is but we should only have one machineset/nodegroup labeled as infra. This helps enforce that (through validation and such). Knowing which MachineSet is labeled as infra, we can set the registry and router node selectors when setting up the master.

docs/cluster_api.md
csrwng: @eparis in the master-config we seem to only be able to specify one:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master/templates/master.yaml.v1.j2#L146

Actually for pod network, we can specify many, so maybe make that one an array?
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master/templates/master.yaml.v1.j2#L146

docs/cluster_api.md
csrwng: I'll put in the an SDN config type that includes plugin name, subnets, etc.

docs/cluster_api.md
smarterclayton: Deployment type is a legacy of a stupid choice we had to make long ago to split our packages.  We really need to collapse all of that - install should care about only:

1. where it gets RPMs
2. what is the matching golden image for that AMI
3. what is the location of images to fetch

We should be collapsing down everything else behind those.  My point about version is that those three are tied - you will always get them as a group.  An end user facing value is "v3.7.9".  A machine facing value is "v3.7.9-pr19384-x-93842948" - both are use cases that the API needs to tolerate.  I would expect that the locations of binaries and the image name would be part of a Version resource and referenced.  I'm not opposed to also supporting direct reference.

docs/cluster_api.md
smarterclayton: Note that other cloud providers don't have regional AMIs (the same way), so there will be AWS specific nuances that can be ignored.

Yes, it is analogous to image stream, but it's more like

```
type Version struct {
  TypeMeta
  ObjectMeta // name is v3.7.9

  // other human presented metadata

  RPMRepos []string // populated 
  ImageFormat string // "openshift/origin-${component}:v3.7.9", or "ci.openshift.org/ci-jobs/pr-2034834098243:${component}"
  VMImages struct {
    GCPImage string
    GCPImageProject string
    
    AWSImageByRegion map[string]string
    
    ... // other cloud providers
  }
}
```

docs/cluster_api.md
smarterclayton: Yes, public hostname is the domain name of the masters as seen from the outside, for console, API discovery, etc

docs/cluster_api.md
```diff
@@ -93,3 +80,37 @@ In more detail:
...
+clusters with that.
+
+- `ObjectMetadata (inline) [metav1.ObjectMetadata]`
+- `TypeMetadata (inline) [metav1.TypeMetadata]`
+
+#### Spec
+- `ImageURL [string]` *-optional- template for images to be used on the cluster (oreg_url in ansible)*
+- `ImageTag [string]` *-optional- the image tag for OpenShift images (either ImageURL or ImageTag can be specified, but not both, openshift_image_tag in ansible)*
+- `Release [string]` *-optional- simple release name (openshift_release in ansible)*
+- `RPMRepository [string]` *-optional- URL for the RPM repository to use*
```
dgoodwin: An array might be useful here. Also we might want a struct for an RPM repository to cover those basic settings we need. (name, URL, gpg key, gpgcheck, etc)

docs/cluster_api.md
```diff
@@ -93,3 +80,37 @@ In more detail:
...
+### ClusterVersion
+Standalone resource that encapsulates details of a named version. It
+defines general OpenShift details about a particular version such as 
+the image template and RPM version as well as cloud-specific information
+such as the AWS AMI name.
+
+A set of predefined ClusterVersion resources can be initially created in a shared 
+namespace (like `openshift`) to define the versions supported out of the box.
+Users are able to create their own in their own namespace and associate their
+clusters with that.
```
dgoodwin: Would it be reasonable to note that a version cannot ever be modified, and only deleted when it's no longer referenced by any cluster or machineset?

docs/cluster_api.md
```diff
@@ -93,3 +80,37 @@ In more detail:
...
+
+- `ObjectMetadata (inline) [metav1.ObjectMetadata]`
+- `TypeMetadata (inline) [metav1.TypeMetadata]`
+
+#### Spec
+- `ImageURL [string]` *-optional- template for images to be used on the cluster (oreg_url in ansible)*
+- `ImageTag [string]` *-optional- the image tag for OpenShift images (either ImageURL or ImageTag can be specified, but not both, openshift_image_tag in ansible)*
+- `Release [string]` *-optional- simple release name (openshift_release in ansible)*
```
dgoodwin: I'm unsure if we're going to need this one given how things are playing out. This was pretty focused on the end customer using ansible to just install the latest v3.x, in our world with pre-defined versions including specific AMIs, we might just want to stay specific and skip this one.

docs/cluster_api.md
csrwng: ack, the struct makes sense.
What would be the purpose of the array?

dgoodwin: Was thinking if we needed to land more than one repo on the system. (EPEL for example) Maybe that's just past experience and not revelant now, when I worked on cluster lifecycle my typical inventory had two repos, one for the OCP bits and one for RHEL nightly extras.

docs/cluster_api.md
csrwng: we could... though I'm not sure there's precedent for doing that elsewhere in kube. Determining that there's no references to a resource is hard and can be racy. But I guess we could do some kind of mark and sweep. My preference would be to grab whatever we need from the currently associated version, stamp out a new machine set or whatever else we need and set the fields of that as the source of truth.

csrwng:
Updated to address recent comments, added field descriptions and cleaned up a few things.

docs/cluster_api.md
eparis: we prevent the deletion of a PVC so long as it is in use by a pod (in 3.8)

eparis: Is this something we would expect (someday, please not as a part of our first efforts) that this could be a ClusterVersionStream, where if we release an errata a customers ClusterVersion would immediately include it?

Is ClusterVersion required in a minimal viable product?

docs/cluster_api.md
```diff
@@ -0,0 +1,121 @@
...
+- `AdminKubeconfig [corev1.LocalReference]` *Reference to secret that contains an admin.kubeconfig to access the cluster as cluster administrator after the cluster is up and running*
+- `Conditions [array]` *Detailed status for the cluster*
+  - `Type [string]` *Type of condition - InfraProvisioned, MasterConfigured, Running*
+  - `Status [string]` *Status of the condition - True, False, Unknown*
+  - `LastProbeTime [string]` *-optional- last time the condition was checked*
+  - `LastTransitionTime [string]` *-optional- last time the condition transitioned*
+  - `Message [string]` *-optional- Human readable message associated with the condition*
+  - `Reason [string]` *-optional- Single word camel-case reason for the condition*
+- `MachineSetCount [int]` *count of machine sets that have been created for this cluster*
+- `Provisioned [boolean]` *true if cluster-level infrastructure has been provisioned for this cluster*
```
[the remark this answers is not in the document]
staebler: My thought would be that it would do exactly that.

docs/cluster_api.md
```diff
@@ -0,0 +1,121 @@
...
+  - `Status [string]` *Status of the condition - True, False, Unknown*
+  - `LastProbeTime [string]` *-optional- last time the condition was checked*
+  - `LastTransitionTime [string]` *-optional- last time the condition transitioned*
+  - `Message [string]` *-optional- Human readable message associated with the condition*
+  - `Reason [string]` *-optional- Single word camel-case reason for the condition*
+
+
+### MachineSetConfig
+  - `NodeType [NodeType:string]` *Enumerated node type for the machine set (compute or master)*
+  - `Infra [boolean]` *True if this machine set should contain infrastructure pods (registry, router)*
```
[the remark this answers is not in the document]
staebler: The controller needs to know the node type to determine whether to run the master playbook on the node. Consequently, at least for the master nodes, the value needs to be there, whether it is in a label or in it's own field. It doesn't seem appropriate to me to require the user to know that certain labels are required. What is the benefit of having the information in a label instead of a field? What is the reason for resistance to a field?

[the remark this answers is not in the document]
eparis: My main opposition to fields stems largely from the fact that infra nodes are an openshift only concept that don't actually make any sense to me in the greater picture of what we are trying to build. There should never be any special magic nodes. All pinning that we do to special nodes is wrong. I understand why we do some of that today (mostly around the lack of a deschedular, the fact that we don't really trust resource usage separation, and difficulties getting traffic to our routers), but codifying specialness into our API for all time seems very wrong. I want infra nodes to die.

Masters are a little bit different. Masters themselves wouldn't exist at all in a self-hosted world (aka, what I think bootkube ends up doing) but I think the path to success for having non-special masters is long enough are far enough out of my vision that I'm ok codifying masters into the api.

All of this is also colored with the fact that I believe we MUST merge with the cluster-api working group. They are starting from the bottom and working up. We are starting from the top and working down. But eventually we must go to them. Upstream has NO concept of an infra. Infra will not survive the merger.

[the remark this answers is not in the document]
staebler: It sounds like we are conflating issues a bit.

`NodeType` as it is now only distinguishes between machines that have the master setup run on them and those that do not. It has nothing to do with infra in any way. Writing that out makes it striking to me that a boolean would be a better choice than an enumeration. If we have a boolean that identified whether a node was a master node or not, would you have an objection to that? In terms of staying in line with upstream, the current iteration of that cluster API has a string array of roles for each machine. Valid roles currently are "Master" and "Node". So, I don't think that we are far apart from upstream with what we have here, with respect to node types.

For determining which machines have infra setup run on them, I could see that fitting into the roles concept from upstream, where another role name "Infra" is added to the current roles of "Master" and "Node". Perhaps the roles can be extensible beyond the kubernetes-wide roles of "Master" and "Node". Openshift could accept a role of "Infra" that is ignored or not accepted by other platforms.

docs/cluster_api.md
csrwng: > we prevent the deletion of a PVC so long as it is in use by a pod (in 3.8)

It'd be good to understand how we prevent raciness in this case. From the time that you determine that a PVC is safe to delete to the time that it's actually deleted, a pod could be created that references the PVC unless we do some kind of marking of the PVC as being in process of deletion.



> Is ClusterVersion required in a minimal viable product?

I don't think it is. But I don't think it would hurt either. I kind of like the concept of keeping all the things that work together like image tags and ami names in a single place. Do you want to leave it out of an initial implementation?

docs/cluster_api.md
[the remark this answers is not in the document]
csrwng: +1 to the idea of having roles. @eparis would that be acceptable?
It would give me enough flexibility to create a cluster with a single machine set where the machine set has the role of master, node and infra. And the infra role is something upstream doesn't have to know about.

docs/cluster_api.md
```diff
@@ -0,0 +1,121 @@
...
+  - `Status [string]` *Status of the condition - True, False, Unknown*
+  - `LastProbeTime [string]` *-optional- last time the condition was checked*
+  - `LastTransitionTime [string]` *-optional- last time the condition transitioned*
+  - `Message [string]` *-optional- Human readable message associated with the condition*
+  - `Reason [string]` *-optional- Single word camel-case reason for the condition*
+- `MachineSetCount [int]` *count of machine sets that have been created for this cluster*
+- `Provisioned [boolean]` *true if cluster-level infrastructure has been provisioned for this cluster*
+- `Running [boolean]` *true if the cluster is running and is ready to accept API requests* 
+- `MasterMachineSetName [string]` *name of the machine set designated as master*
+- `InfraMachineSetName [string]` *name of the machine set designated as infra (could be the same as master)*
```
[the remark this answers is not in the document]
csrwng: How do you propose that we determine which node labels to use for infra pods (registry/router) when building an inventory for the master? Would it be by convention? (say, the ClusterMachineSet named 'infra' for the cluster machine sets?)

docs/cluster_api.md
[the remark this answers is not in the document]
dgoodwin: Agreed.

docs/cluster_api.md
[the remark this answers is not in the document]
dgoodwin: Could the machine set carry labels to apply to those nodes, and we could enforce in our controllers and reject cluster definitions that don't include anything labelled infra.

The fact that it won't make much sense to upstream Kube makes me think it's probably not something we should bake too much into the data model.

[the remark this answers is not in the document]
twiest: +1 on node labels for this.

docs/cluster_api.md
```diff
@@ -0,0 +1,121 @@
...
+#### Spec:
+- `Hardware`
+  - `AWS: [struct, optional]` *AWS-specific hardware configuration for the cluster*
+    - `AccountSecret [corev1.LocalReference]` *Reference to a secret that contains an AWS access key id and secret access key*
+    - `SSHKeyName [string]` *SSH secret name to use for machines in this cluster (ie. 'libra')*
+    - `SSHSecret [corev1.LocalReference]` *Reference to a secret that contains a SSH private key to SSH into the cluster*
+    - `Region [string]` *Region where the cluster will be created (us-east-1 for example)*
+    - `VPCName [string, optional]` *Name of the VPC to use for this cluster (or create if not present)*
+    - `VPCSubnet [string]` *Subnet of the VPC (only used if creating a new one)*
+- OpenShift Configuration: *Configuration of the cluster that's not cloud-provider dependent*
```
twiest: I'm not seeing a place here to define the actual cluster name. Am I just missing it?

We need it in order to have multiple kube clusters in the same AZ: [KubernetesCluster](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/tags.go#L30-L34).

docs/cluster_api.md
```diff
@@ -0,0 +1,121 @@
...
+
+### MachineSetConfig
+  - `NodeType [NodeType:string]` *Enumerated node type for the machine set (compute or master)*
+  - `Infra [boolean]` *True if this machine set should contain infrastructure pods (registry, router)*
+  - `Size [int]` *How many machines should be created for this machine set*
+  - `Version [corev1.LocalObjectReference]` *Reference to ClusterVersion resource*
+  - `Hardware`
+    - `AWS` *-optional- contains AWS specific configuration*
+      - `InstanceType [string]` *EC2 instance type (ie. m4.xlarge)*
+      - `AMIName [string]` *Name of AMI used to create this machine set (derived from Version object)*
```
twiest: I think it'd be good to be able to define cloud instance tags here as well (e.g. AWS instance tags).

We use instance tags for quite a few things and being able to easily define them here would be very powerful, flexible and convenient.

docs/cluster_api.md
```diff
@@ -0,0 +1,121 @@
...
+#### Spec
+- `ImageURL [string]` *-optional- template for images to be used on the cluster (oreg_url in ansible)*
+- `ImageTag [string]` *-optional- the image tag for OpenShift images (either ImageURL or ImageTag can be specified, but not both, openshift_image_tag in ansible)*
+- `RPMRepositories [array]` *-optional- description of RPM repositories to use in the cluster (openshift_additional_repos in ansible)*
+  - `URL [string]` *URL to the RPM repository*
+  - `GPGKey [string]` *URL to GPG key for repository*
+- `PackageVersion [string]` *-optional- RPM package version (openshift_pkg_version in ansible)*
+- `Hardware` *contains cloud-provider specific configuration (AWS for now)*
+  - `AWS` 
+    - `AMIName [string]` *name of the AMI that corresponds to this version*
```
twiest: This is defined in MachineSetConfig as well as here. I get why it's defined in MachineSetConfig. Why is it defined here as well?

In fact, why is Hardware defined here at all?

docs/cluster_api.md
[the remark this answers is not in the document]
jorgemoralespou: keep in mind that our subscription model accounts for these types of nodes (Infra nodes where application workloads are not run are not count as a worker node, and come for free as a master type of sub). While I do agree that modeling this is bad, you'll probably need to get back to @mpbarrett  so this is known, and possibly the how subs work revisited.

docs/cluster_api.md
```diff
@@ -0,0 +1,121 @@
...
+- `MasterMachineSetName [string]` *name of the machine set designated as master*
+- `InfraMachineSetName [string]` *name of the machine set designated as infra (could be the same as master)*
+
+### MachineSet
+
+- `ObjectMetadata [metav1.ObjectMetadata]`
+- `TypeMetadata [metav1.TypeMetadata]`
+
+#### Spec
+- `MachineSetConfig [MachineSetConfig,inline]` *Configuration for the machine set (see below)*
```
dgoodwin: Why do we have this additional layer for config, vs just the spec?

csrwng:
Closing this since we already have self-documenting API types

PR #26, opened 2017-12-15
API Updates

csrwng:
Implements the current state of the API proposal
Renames NodeGroup -> MachineSet, Node -> Machine


csrwng:
@staebler @dgoodwin still fixing up the cluster controller unit test that merged yesterday, should update shortly

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
+	MachineSetCount int
 
-	Name string
+	// MasterMachineSetName is the name of the master machine set
+	MasterMachineSetName string
+
+	// InfraMachineSetName is the name of the infra machine set
+	InfraMachineSetName string
+
+	// Secret to access the cluster master when running
```
dgoodwin: Could we rename or comment this to indicate it's a bootstrap concept for nodes to join the cluster.

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
+	KubeconfigSecret *corev1.LocalObjectReference
+
+	// Provisioned is true if the hardware for the cluster has been provisioned
+	Provisioned bool
```
dgoodwin: Is this referring specifically to cluster prerequisites/infra, or also the machine sets?

dgoodwin:
Just a couple comments from me, feel free to merge when you're had a chance to look through @staebler

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
-	// active for the cluster
-	MasterNodeGroups int
+// AWSClusterSpec contains cluster-wide configuration for a cluster on AWS
+type AWSClusterSpec struct {
+	// AccountSeceret refers to a secret that contains the AWS account access
+	// credentials
+	AccountSecret corev1.LocalObjectReference
+
+	// Region specifies the AWS region where the cluster will be created
+	Region string
```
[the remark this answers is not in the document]
csrwng: My understanding is that we wouldn't support multiple regions for a single cluster, based on a talk a while ago with @eparis

pkg/apis/clusteroperator/types.go
csrwng: It's not only for bootstrapping, but for general access to the cluster by the cluster operator.

csrwng: Will update the comment

pkg/apis/clusteroperator/types.go
csrwng: I was thinking of cluster-level prereqs, but should make that clear. For machine set status, we should look into the machine set objects.

csrwng:
Addressed comments so far. Unit tests are now passing locally for me.

csrwng:
forgot about the integration tests :) fixing now

csrwng:
integration test fixed

pkg/apis/clusteroperator/validation/machine.go
```diff
@@ -22,43 +22,43 @@ import (
...
 
-	allErrs = append(allErrs, validateNodeSpec(&node.Spec, field.NewPath("spec"))...)
+	allErrs = append(allErrs, validateMachineSpec(&node.Spec, field.NewPath("spec"))...)
 
 	return allErrs
 }
 
-// validateNodeSpec validates the spec of a node.
-func validateNodeSpec(spec *clusteroperator.NodeSpec, fldPath *field.Path) field.ErrorList {
+// validateMachineSpec validates the spec of a node.
```
staebler: Change node to machine.

pkg/apis/clusteroperator/validation/machine.go
```diff
@@ -22,43 +22,43 @@ import (
...
 	allErrs := field.ErrorList{}
 
 	allErrs = append(allErrs, validateNodeType(spec.NodeType, fldPath.Child("nodeType"))...)
 
 	return allErrs
 }
 
-// validateNodeStatus validates the status of a node.
-func validateNodeStatus(status *clusteroperator.NodeStatus, fldPath *field.Path) field.ErrorList {
+// validateMachineStatus validates the status of a node.
```
staebler: Change node to machine.

pkg/apis/clusteroperator/validation/machine.go
```diff
@@ -22,43 +22,43 @@ import (
...
+func validateMachineStatus(status *clusteroperator.MachineStatus, fldPath *field.Path) field.ErrorList {
 	return field.ErrorList{}
 }
 
-// ValidateNodeUpdate validates an update to the spec of a node.
-func ValidateNodeUpdate(new *clusteroperator.Node, old *clusteroperator.Node) field.ErrorList {
+// ValidateMachineUpdate validates an update to the spec of a node.
```
staebler: Change node to machine.

pkg/apis/clusteroperator/validation/machine.go
```diff
@@ -22,43 +22,43 @@ import (
...
 
-	allErrs = append(allErrs, validateNodeSpec(&new.Spec, field.NewPath("spec"))...)
+	allErrs = append(allErrs, validateMachineSpec(&new.Spec, field.NewPath("spec"))...)
 
 	return allErrs
 }
 
-// ValidateNodeStatusUpdate validates an update to the status of a node.
-func ValidateNodeStatusUpdate(new *clusteroperator.Node, old *clusteroperator.Node) field.ErrorList {
+// ValidateMachineStatusUpdate validates an update to the status of a node.
```
staebler: Change node to machine.

pkg/apis/clusteroperator/validation/machine_test.go
```diff
@@ -24,31 +24,31 @@ import (
 	"github.com/openshift/cluster-operator/pkg/apis/clusteroperator"
 )
 
-// getValidNode gets a node that passes all validity checks.
-func getValidNode() *clusteroperator.Node {
-	return &clusteroperator.Node{
-		Spec: clusteroperator.NodeSpec{
+// getValidMachine gets a node that passes all validity checks.
```
staebler: Change node to machine.

pkg/apis/clusteroperator/validation/machineset.go
```diff
@@ -25,19 +25,19 @@ import (
...
-	allErrs = append(allErrs, validateNodeGroupClusterOwner(nodeGroup.GetOwnerReferences(), field.NewPath("metadata").Child("ownerReferences"))...)
-	allErrs = append(allErrs, validateNodeGroupSpec(&nodeGroup.Spec, field.NewPath("spec"))...)
+	allErrs = append(allErrs, validateMachineSetClusterOwner(machineSet.GetOwnerReferences(), field.NewPath("metadata").Child("ownerReferences"))...)
+	allErrs = append(allErrs, validateMachineSetSpec(&machineSet.Spec, field.NewPath("spec"))...)
 
 	return allErrs
 }
 
-// validateNodeGroupClusterOwner validates that a node group has an owner and
+// validateMachineSetClusterOwner validates that a node group has an owner and
```
staebler: Change node group to machine set.

pkg/apis/clusteroperator/validation/machineset.go
```diff
@@ -50,14 +50,14 @@ func validateNodeGroupClusterOwner(ownerRefs []metav1.OwnerReference, fldPath *f
...
-		allErrs = append(allErrs, field.Required(fldPath, "nodegroup must have an owner"))
+		allErrs = append(allErrs, field.Required(fldPath, "machineset must have an owner"))
 	}
 
 	return allErrs
 }
 
-// validateNodeGroupSpec validates the spec of a node group.
-func validateNodeGroupSpec(spec *clusteroperator.NodeGroupSpec, fldPath *field.Path) field.ErrorList {
+// validateMachineSetSpec validates the spec of a node group.
```
staebler: Change node group to machine set.

pkg/apis/clusteroperator/validation/machineset.go
```diff
@@ -69,41 +69,41 @@ func validateNodeGroupSpec(spec *clusteroperator.NodeGroupSpec, fldPath *field.P
 	return allErrs
 }
 
-// validateNodeGroupStatus validates the status of a node group.
-func validateNodeGroupStatus(status *clusteroperator.NodeGroupStatus, fldPath *field.Path) field.ErrorList {
+// validateMachineSetStatus validates the status of a node group.
```
staebler: Change node group to machine set.

pkg/apis/clusteroperator/validation/machineset.go
```diff
@@ -69,41 +69,41 @@ func validateNodeGroupSpec(spec *clusteroperator.NodeGroupSpec, fldPath *field.P
...
+func validateMachineSetStatus(status *clusteroperator.MachineSetStatus, fldPath *field.Path) field.ErrorList {
 	return field.ErrorList{}
 }
 
-// ValidateNodeGroupUpdate validates an update to the spec of a node group.
-func ValidateNodeGroupUpdate(new *clusteroperator.NodeGroup, old *clusteroperator.NodeGroup) field.ErrorList {
+// ValidateMachineSetUpdate validates an update to the spec of a node group.
```
staebler: Change node group to machine set.

pkg/apis/clusteroperator/validation/machineset.go
```diff
@@ -69,41 +69,41 @@ func validateNodeGroupSpec(spec *clusteroperator.NodeGroupSpec, fldPath *field.P
...
-	allErrs = append(allErrs, validateNodeGroupImmutableClusterOwner(new.GetOwnerReferences(), old.GetOwnerReferences(), field.NewPath("metadata").Child("ownerReferences"))...)
+	allErrs = append(allErrs, validateMachineSetSpec(&new.Spec, field.NewPath("spec"))...)
+	allErrs = append(allErrs, validateMachineSetImmutableClusterOwner(new.GetOwnerReferences(), old.GetOwnerReferences(), field.NewPath("metadata").Child("ownerReferences"))...)
 
 	return allErrs
 }
 
-// validateNodeGroupImmutableClusterOwner validates that the cluster owner of
+// validateMachineSetImmutableClusterOwner validates that the cluster owner of
 // a node group is immutable.
```
staebler: Change node group to machine set.

pkg/apis/clusteroperator/validation/machineset.go
```diff
@@ -69,41 +69,41 @@ func validateNodeGroupSpec(spec *clusteroperator.NodeGroupSpec, fldPath *field.P
...
 	} else {
-		allErrs = append(allErrs, field.Required(fldPath, "nodegroup must have an owner"))
+		allErrs = append(allErrs, field.Required(fldPath, "machineset must have an owner"))
 	}
 
 	return allErrs
 }
 
-// ValidateNodeGroupStatusUpdate validates an update to the status of a node
+// ValidateMachineSetStatusUpdate validates an update to the status of a node
```
staebler: Change node group to machine set.

pkg/apis/clusteroperator/validation/machineset_test.go
```diff
@@ -37,44 +37,44 @@ func getValidClusterOwnerRef() metav1.OwnerReference {
 	}
 }
 
-// getValidNodeGroup gets a node group that passes all validity checks.
-func getValidNodeGroup() *clusteroperator.NodeGroup {
-	return &clusteroperator.NodeGroup{
+// getValidMachineSet gets a node group that passes all validity checks.
```
staebler: Change node group to machine set.

pkg/apis/clusteroperator/validation/machineset_test.go
```diff
@@ -37,44 +37,44 @@ func getValidClusterOwnerRef() metav1.OwnerReference {
...
-			nodeGroup: getValidNodeGroup(),
+			nodeGroup: getValidMachineSet(),
 			valid:     true,
 		},
 		{
 			name: "invalid cluster owner",
-			nodeGroup: func() *clusteroperator.NodeGroup {
-				ng := getValidNodeGroup()
+			nodeGroup: func() *clusteroperator.MachineSet {
+				ng := getValidMachineSet()
```
staebler: Change `ng` variable names to something appropriate for a MachineSet.

pkg/apis/componentconfig/types.go
```diff
@@ -79,20 +79,20 @@ type ControllerManagerConfiguration struct {
 	// but more CPU (and network) load.
 	ConcurrentClusterSyncs int32
 
-	// concurrentNodeGroupSyncs is the number of node group objects that are
+	// concurrentMachineSetSyncs is the number of node group objects that are
```
staebler: Change node group to machine set.

pkg/apis/componentconfig/types.go
```diff
@@ -79,20 +79,20 @@ type ControllerManagerConfiguration struct {
...
 	// allowed to sync concurrently. Larger number = more responsive node groups,
 	// but more CPU (and network) load.
-	ConcurrentNodeGroupSyncs int32
+	ConcurrentMachineSetSyncs int32
 
-	// concurrentNodeSyncs is the number of node objects that are
+	// concurrentMachineSyncs is the number of node objects that are
```
staebler: Change node to machine.

pkg/apis/componentconfig/types.go
```diff
@@ -79,20 +79,20 @@ type ControllerManagerConfiguration struct {
...
 	// allowed to sync concurrently. Larger number = more responsive nodes,
 	// but more CPU (and network) load.
-	ConcurrentNodeSyncs int32
+	ConcurrentMachineSyncs int32
 
-	// concurrentMasterNodeSyncs is the number of master node objects that are
+	// concurrentMasterSyncs is the number of master node objects that are
```
staebler: Change node to machine

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
 
-	// ComputeNodeGroups is the number of actual compute node groups that are
-	// active for the cluster
-	ComputeNodeGroups int
+	// +optional
+	// VPCName specifies the name of the VPC to associate with the cluster.
+	// If existing, it will not be created. If left blank, the VPC name will
```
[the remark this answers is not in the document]
csrwng: It is indeed the former. Will definitely clarify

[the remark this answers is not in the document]
csrwng: If we derive a name, I would have envisioned some random suffix to not clash with existing stuff.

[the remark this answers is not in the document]
csrwng: So I would say that we *always* create it, if blank

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
+	VPCSubnet string
 }
 
-// ClusterNodeGroup is a node group defined in a Cluster resource
-type ClusterNodeGroup struct {
-	Size int
+// ClusterConfigSpec contains OpenShift configuration for a cluster
+type ClusterConfigSpec struct {
+	// DeploymentType indicates the type of OpenShift deployment to create
+	DeploymentType string
```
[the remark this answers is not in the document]
csrwng: Yes, we have specific deployment types, will create a go type

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
+
+	// OpenShiftVersion is the version of OpenShift to install
+	OpenshiftVersion string
+
+	// SDNPluginName is the name of the SDN plugin to use for this install
+	SDNPluginName string
```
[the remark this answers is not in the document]
csrwng: Yes, will also define in a go type

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
+
+	// ServiceNetworkSubnet is the CIDR to use for service IPs in the cluster
+	ServiceNetworkSubnet string
```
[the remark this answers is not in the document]
csrwng: Yes they are, a default value is used in the ansible playbook when not specified.

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
+	Items []MachineSet
+}
+
+// ClusterMachineSet is the specification of a machine set in a cluster
+type ClusterMachineSet struct {
+	// Name is a unique name for the machine set within the cluster
+	Name string
+
+	// Spec is the specification of the machine set
+	Spec MachineSetSpec
```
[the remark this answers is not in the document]
csrwng: I'll create another type that's shared between MachineSetSpec and ClusterMachineSet named *MachineSetConfig* if there are no objections.

pkg/apis/clusteroperator/types.go
```diff
@@ -47,109 +48,279 @@ type ClusterList struct {
...
+	AWS *MachineSetAWSHardwareSpec
 }
 
+// MachineSetAWSHardwareSpec specifies AWS hardware for a MachineSet
+type MachineSetAWSHardwareSpec struct {
+	// InstanceType is the type of instance to use for machines in this MachineSet
+	InstanceType string
+
+	// AMIName is the name of the AMI to use for machines in this MachineSet
+	AMIName string
```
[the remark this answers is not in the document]
csrwng: It'll likely be the case for the InstanceType as well. Does it make sense to have a field on the ClusterSpec named DefaultMachineSetHardwareSpec of type MachineSetHardwareSpec ?

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -47,109 +48,278 @@ type ClusterList struct {
...
-type ClusterComputeNodeGroup struct {
-	ClusterNodeGroup `json:",inline"`
+// ClusterStatus contains the status for a cluster
+type ClusterStatus struct {
+	// MachineSetCount is the number of actual machine sets that are active for the cluster
+	MachineSetCount int `json:"machineSetCount"`
 
-	Name string `json:"name"`
+	// MasterMachineSetName is the name of the master machine set
+	MasterMachineSetName string `json:"masterMachineSetName,omitempty"`
```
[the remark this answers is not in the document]
csrwng: The latter. They are set by the ClusterController when syncing a Cluster. They contain the names of the actual MachineSets that are designated as master and infra. My thinking with this was that it would be useful when generating inventories. They would also help in the case of the MachineSetController checking whether for a given cluster the master MachineSet is ready and it can proceed to provision compute machineSets. Sure, we can find this by querying MachineSets owned by a cluster, but this is information we have available to us when calculating the Status of a cluster, so why not use it.

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -47,109 +48,278 @@ type ClusterList struct {
...
+
+	// AMIName is the name of the AMI to use for machines in this MachineSet
+	AMIName string `json:"amiName"`
 }
 
-type NodeGroupStatus struct {
+// MachineSetStatus is the status of a MachineSet
+type MachineSetStatus struct {
+	// MachineCount is the total number of provisioned machines for the MachineSet
+	MachineCount int `json:"machineCount"`
```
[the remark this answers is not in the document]
csrwng: ack

pkg/apis/clusteroperator/validation/cluster.go
```diff
@@ -43,56 +41,71 @@ func ValidateCluster(cluster *clusteroperator.Cluster) field.ErrorList {
...
-
-	if status.MasterNodeGroups < 0 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("masterNodeGroups"), status.MasterNodeGroups, "must be non-negative"))
+	if ref == nil {
+		return allErrs
 	}
-	if status.ComputeNodeGroups < 0 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("computeNodeGroups"), status.ComputeNodeGroups, "must be non-negative"))
+	if len(ref.Name) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("name"), ""))
```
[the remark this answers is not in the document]
csrwng: will add

csrwng:
@staebler thank you much for the review so far, working on making updates now.

pkg/apis/clusteroperator/types.go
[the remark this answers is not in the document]
staebler: I am less sure about the likelihood of it being the same for InstanceType. However, I think it's a good idea to do as you suggest with putting a DefaultMachineSetHardwareSpec in the cluster spec (although that is a mouthful). The logic in validation would be that, for each field in MachineSetHardwareSpec, either the value is set in DefaultMachineSetHardwareSpec or it is set in **all** of the machine set specs. So the user does not have to set defaults if they do not want to.

pkg/apis/clusteroperator/v1alpha1/types.go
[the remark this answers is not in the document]
staebler: Ah, I see where I got confused. This is the full, realized name of the machine set. It is not the name from `ClusterMachineSet`. OK. It makes more sense to me to have it then. I still am not entirely convinced that they are worthwhile, though, based on my understanding of how things will work. In both of the cases that you provided, the controller wouldn't make use of the name. Inventory and ready checking will both be done using tags applied to the auto scaling group. They won't be doing a query based upon the kube name of the machine set. In any event, I am fine with leaving them in for now.

pkg/apis/clusteroperator/types.go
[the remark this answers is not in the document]
csrwng: Actually for this one I'm less sure of confining it to a predefined set of options. We would normally pick between `ovs-subnet` ,`ovs-multitenant` and `ovs-networkpolicy`. But we also support other plugins from upstream like flannel. I'll leave it as a string.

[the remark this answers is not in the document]
staebler: It's a choice between providing the user with early feedback when there is a misconfiguration and being able to support new features without changes. In this case, it is specifically something that is meant to be a plug-in, which implies flexibility in accepting new features. So I agree that we should keep it as a string and not do API validation for known values.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -526,14 +502,14 @@ func (c *ClusterController) manageNodeGroups(nodeGroups []*clusteroperator.NodeG
...
-		return buildNewNodeGroup(cluster, clusterNodeGroup, nodeType, nodeGroupNamePrefix), false, nil
+func (c *ClusterController) manageMachineSet(cluster *clusteroperator.Cluster, machineSet *clusteroperator.MachineSet, clusterMachineSetSpec clusteroperator.MachineSetSpec, machineSetNamePrefix string) (*clusteroperator.MachineSet, bool, error) {
+	if machineSet == nil {
+		return buildNewMachineSet(cluster, clusterMachineSetSpec, machineSetNamePrefix), false, nil
 	}
 
-	if nodeGroup.Spec.Size != clusterNodeGroup.Size {
-		glog.V(2).Infof("Changing size of node group %s from %v to %v", nodeGroup.Name, nodeGroup.Spec.Size, clusterNodeGroup.Size)
-		return buildNewNodeGroup(cluster, clusterNodeGroup, nodeType, nodeGroupNamePrefix), true, nil
+	if machineSet.Spec.Size != clusterMachineSetSpec.Size {
```
staebler: Presumably, there is more that we need to check here other than the size of the machine set. With the change to use a shared spec, it may be that we can get away with doing a `reflect.DeepEqual`. Of course, there may be changes to do not require a delete and re-create of the machine set and can instead do an update on the existing machine set.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -594,37 +571,38 @@ func buildNewNodeGroup(cluster *clusteroperator.Cluster, clusterNodeGroup cluste
...
-	return fmt.Sprintf("%s-compute-%s-", cluster.Name, nodeGroupName)
+func getNamePrefixForMachineSet(cluster *clusteroperator.Cluster, machineSet clusteroperator.ClusterMachineSet) string {
+	nodeTypePrefix := ""
+	switch machineSet.Spec.NodeType {
+	case clusteroperator.NodeTypeMaster:
+		nodeTypePrefix = "master"
+	case clusteroperator.NodeTypeCompute:
+		nodeTypePrefix = "compute"
+	}
+	return fmt.Sprintf("%s-%s-%s-", cluster.Name, nodeTypePrefix, machineSet.Name)
```
staebler: The nodeTypePrefix is not necessary. It is sufficient for the machine set name prefix to be `fmt.Sprintf("%s-%s-", cluster.Name, machineSet.Name)`. The "compute" was necessary before because you could have conceivably had a compute node that the user had named master.

pkg/controller/cluster/cluster_controller_test.go
```diff
@@ -85,42 +85,50 @@ func getKey(cluster *clusteroperator.Cluster, t *testing.T) string {
...
 			Namespace: testNamespace,
 		},
 		Spec: clusteroperator.ClusterSpec{
-			MasterNodeGroup: clusteroperator.ClusterNodeGroup{
-				Size: masterSize,
-			},
-			ComputeNodeGroups: computes,
+			MachineSets: append(computes, clusteroperator.ClusterMachineSet{
+				Name: "master",
+				Spec: clusteroperator.MachineSetSpec{
```
staebler: Should the master machine set have `infra` true to match up with what is in the cluster status?

pkg/controller/cluster/cluster_controller_test.go
```diff
@@ -319,27 +346,26 @@ func (ea expectedNodeGroupDeleteAction) validate(t *testing.T, action clientgote
...
+		namePrefix: getNamePrefixForComputeMachineSet(cluster, computeName),
 	}
 }
 
 // expectedClusterStatusUpdateAction is an expected client action to update
 // the status of a cluster.
 type expectedClusterStatusUpdateAction struct {
-	masterNodeGroups  int
-	computeNodeGroups int
+	machineSets int
```
staebler: We can add this later, but we will want to verify `MasterMachineSetName` and `InfraMachineSetName` in this expected action as well.

pkg/controller/master/master_controller.go
```diff
@@ -49,107 +49,107 @@ const (
...
+// NewMasterController returns a new *MasterController.
+func NewMasterController(machineInformer informers.MachineInformer, kubeClient kubeclientset.Interface, clusteroperatorClient clusteroperatorclientset.Interface) *MasterController {
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartLogging(glog.Infof)
 	// TODO: remove the wrapper when every clients have moved to use the clientset.
 	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(kubeClient.CoreV1().RESTClient()).Events("")})
 
 	if kubeClient != nil && kubeClient.CoreV1().RESTClient().GetRateLimiter() != nil {
-		metrics.RegisterMetricAndTrackRateLimiterUsage("clusteroperator_master_node_controller", kubeClient.CoreV1().RESTClient().GetRateLimiter())
+		metrics.RegisterMetricAndTrackRateLimiterUsage("clusteroperator_master_machine_controller", kubeClient.CoreV1().RESTClient().GetRateLimiter())
```
staebler: This should probably be "clusteroperator_master_controller" or we should change the name of the controller to `MasterMachineController`.

pkg/controller/master/master_controller.go
```diff
@@ -49,107 +49,107 @@ const (
...
 	}
 
-	c := &MasterNodeController{
+	c := &MasterController{
 		client: clusteroperatorClient,
-		queue:  workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "masterNode"),
+		queue:  workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "masterMachine"),
```
staebler: "master" instead of "masterMachine"

pkg/controller/master/master_controller.go
```diff
@@ -49,107 +49,107 @@ const (
...
-func (c *MasterNodeController) Run(workers int, stopCh <-chan struct{}) {
+// machines will be handled in parallel.
+func (c *MasterController) Run(workers int, stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
 	defer c.queue.ShutDown()
 
-	glog.Infof("Starting master node controller")
-	defer glog.Infof("Shutting down master node controller")
+	glog.Infof("Starting master machine controller")
+	defer glog.Infof("Shutting down master machine controller")
```
staebler: Replace "master machine controller" with "master controller".

pkg/registry/clusteroperator/machine/storage.go
```diff
@@ -136,11 +136,11 @@ func NewStorage(opts etcd.Options) (nodes, nodesStatus rest.Storage) {
 		// Used to match objects based on labels/fields for list.
 		PredicateFunc: Match,
 		// DefaultQualifiedResource should always be plural
-		DefaultQualifiedResource: clusteroperator.Resource("clusterservicenodes"),
+		DefaultQualifiedResource: clusteroperator.Resource("clusterservicemachines"),
```
staebler: This is a mistake in the update from the code that was brought over from service catalog. This should be "machines" not "clusterservicemachines".

pkg/registry/clusteroperator/machineset/storage.go
```diff
@@ -136,11 +136,11 @@ func NewStorage(opts etcd.Options) (nodegroups, nodegroupsStatus rest.Storage) {
 		// Used to match objects based on labels/fields for list.
 		PredicateFunc: Match,
 		// DefaultQualifiedResource should always be plural
-		DefaultQualifiedResource: clusteroperator.Resource("clusterservicenodegroups"),
+		DefaultQualifiedResource: clusteroperator.Resource("clusterservicemachinesets"),
```
staebler: Likewise here. This should be "machinesets" instead of "clusterservicemachinesets".

csrwng:
@staebler this should be ready for a follow-up review. All review changes are in an additional commit.

pkg/apis/clusteroperator/validation/machine.go
```diff
@@ -22,30 +22,30 @@ import (
...
 	return allErrs
 }
 
-// validateMachineSpec validates the spec of a node.
+// validateMachineSpec validates the spec of a machine.
 func validateMachineSpec(spec *clusteroperator.MachineSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	allErrs = append(allErrs, validateNodeType(spec.NodeType, fldPath.Child("nodeType"))...)
+	allErrs = append(allErrs, validateNodeType(spec.NodeType, fldPath.Child("machineType"))...)
```
staebler: Change "machineType" to "nodeType".

pkg/apiserver/apiserver.go
```diff
@@ -22,14 +22,14 @@ import (
 	serverstorage "k8s.io/apiserver/pkg/server/storage"
 )
 
-// ServiceCatalogAPIServer contains the base GenericAPIServer along with other
+// ClusterOperatorAPIServer contains the base GenericAPIServer along with other
 // configured runtime configuration
-type ServiceCatalogAPIServer struct {
+type ClusterOperatorAPIServer struct {
```
staebler: Nice catch!

pkg/controller/cluster/cluster_controller_test.go
```diff
@@ -218,10 +220,10 @@ func processSync(c *ClusterController, key string) error {
 
 // validateClientActions validates that the client experienced the specified
 // expected actions.
-func validateClientActions(t *testing.T, clusterOperatorClient *clusteroperatorclientset.Clientset, expectedActions ...expectedClientAction) {
+func validateClientActions(t *testing.T, name string, clusterOperatorClient *clusteroperatorclientset.Clientset, expectedActions ...expectedClientAction) {
```
staebler: I don't understand why we need to pass the test name in here. The output is already grouped by the test/subtest.

staebler: And if we really, *really* want to pass in the name, let's at least call the parameter testName.

pkg/controller/cluster/cluster_controller_test.go
```diff
@@ -399,22 +367,78 @@ func (ea expectedClusterStatusUpdateAction) validate(t *testing.T, action client
 
 // validateControllerExpectations validates that the specified cluster is
 // expecting the specified number of creations and deletions of machine sets.
-func validateControllerExpectations(t *testing.T, ctrlr *ClusterController, cluster *clusteroperator.Cluster, expectedAdds, expectedDeletes int) {
+func validateControllerExpectations(t *testing.T, name string, ctrlr *ClusterController, cluster *clusteroperator.Cluster, expectedAdds, expectedDeletes int) {
```
staebler: Again, I don't think name is necessary. But if it is, then call it `testName`.

pkg/apis/clusteroperator/types.go
```diff
@@ -225,26 +250,34 @@ type MachineSetSpec struct {
...
 	// AWS specifies the hardware spec for an AWS MachineSet
+	// +optional
 	AWS *MachineSetAWSHardwareSpec
 }
 
 // MachineSetAWSHardwareSpec specifies AWS hardware for a MachineSet
 type MachineSetAWSHardwareSpec struct {
 	// InstanceType is the type of instance to use for machines in this MachineSet
+	// +optional
 	InstanceType string
```
staebler: We are going to need to validate in the API (1) that the Cluster and the ClusterMachineSet use the same type of hardware spec and (2) that each field in the hardware spec that **needs** a value has one in either the Cluster or the ClusterMachineSet.

This can be done in a follow-on if you write up an issue for it.

pkg/apis/clusteroperator/types.go
```diff
@@ -102,12 +111,25 @@ type ClusterConfigSpec struct {
...
 
+// ClusterDeploymentType is a valid value for ClusterConfigSpec.DeploymentType
+type ClusterDeploymentType string
+
+// These are valid values for cluster deployment type
+const (
+	// ClusterDeploymentTypeOrigin is a deployment type of origin
+	ClusterDeploymentTypeOrigin ClusterDeploymentType = "origin"
+	// ClusterDeploymentTypeAtomicOpenshift is a deployment type of atomic openshift (enterprise)
+	ClusterDeploymentTypeAtomicOpenshift ClusterDeploymentType = "atomic-openshift"
```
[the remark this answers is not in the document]
csrwng: Hmm, it should be. Now I can't find where I'd seen atomic-openshift instead of openshift-enterprise. Will update.

pkg/apis/clusteroperator/types.go
csrwng: https://github.com/openshift/cluster-operator/issues/33

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -199,51 +225,59 @@ type ClusterMachineSet struct {
...
+	MachineSetConfig `json:",inline"`
 }
 
-// MachineSetSpec is the specification for a MachineSet
-type MachineSetSpec struct {
+// MachineSetConfig contains configuration for a MachineSet
+type MachineSetConfig struct {
 	// NodeType is the type of nodes that comprise the MachineSet
-	NodeType NodeType `json:"nodeType"`
+	NodeType NodeType
```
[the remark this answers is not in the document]
csrwng: not intentionally, most likely copy/paste

pkg/controller/cluster/cluster_controller_test.go
csrwng: In the output I was only getting the line of the failure and the message. Because we use a common validation function, it was hard to tell which test/testcase had caused it. I will update the parameter name.

csrwng:
@staebler thx again for the thoughtful review. Latest review updates submitted in separate commit.

csrwng:
rebased, squashed, and fixed infra test

PR #27, opened 2017-12-15
Allow running Ansible without root/elevated SCC.

dgoodwin:
Also adds support for creating clusters in namespaces other than
"cluster-operator".

For now, secrets we use for AWS and SSH are copied to any namespace where a cluster appears. These will fold into the cluster API and/or provisioning eventually.

Ansible jobs will also run in the namespace where the cluster lands keeping things nicely segmented.

### DIFF
--- a/pkg/ansible/runner.go
+++ b/pkg/ansible/runner.go
@@ -8,17 +8,17 @@ import (
 
 	kbatch "k8s.io/api/batch/v1"
 	kapi "k8s.io/api/core/v1"
+	kapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
 const (
-	openshiftAnsibleImage          = "openshift/origin-ansible:v3.7"
-	openshiftAnsibleServiceAccount = "cluster-operator-controller-manager"
-	sshPrivateKeySecret            = "ssh-private-key"
-	operatorNamespace              = "cluster-operator"
-	openshiftAnsibleContainerDir   = "/usr/share/ansible/openshift-ansible/"
-	awsCredentialsSecret           = "aws-credentials"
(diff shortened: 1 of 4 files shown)


dgoodwin:
Worth noting that if you uncomment the "echo" in runner.go, infra provisioning should now be operational.

pkg/ansible/runner.go
```diff
@@ -129,10 +203,8 @@ func (r *ansibleRunner) RunPlaybook(clusterName, jobPrefix, playbook, inventory,
 					},
 				},
 
-				// TODO: possibly drop this once https://github.com/openshift/openshift-ansible/pull/6320 merges, the default run script should then work, minus the vars file :(
-				//Command: []string{"ansible-playbook", "-i", "/ansible/inventory/hosts", "--private-key", "/ansible/ssh/privatekey.pem", playbookPath, "-e", "@/ansible/inventory/vars"},
-
-				// TODO: committing as no-op for now, uncomment command above to start testing provisioning
+				// TODO: committing as no-op for now, uncomment to use the default
+				// origin-ansible container entrypoint and actually run playbooks.
```
staebler: I am confused about what is supposed to be uncommented here? I thought it was the previous line that has been removed.
```
Command: []string{"ansible-playbook", "-i", "/ansible/inventory/hosts", "--private-key", "/ansible/ssh/privatekey.pem", playbookPath, "-e", "@/ansible/inventory/vars"},
```

staebler: Should this be that the next line should be **commented out**?

dgoodwin: You just uncomment the explicit Command and let it fall back to the default container entrypoint. One of our run as root problems was already solved in their entry scripts, and originally we found and fixed a bug but they've got that merged and in the container now. The env vars above control all the options we were doing in previous iteration of the Command. 

Wording isn't good here, should be "comment out" not uncomment. Fixing now.

PR #28, opened 2017-12-19
Provide a fake openshift-ansible image to avoid real AWS

staebler:
* `make images` will now also build a fake-openshift-ansible image built on top of the openshift/origin-ansible image. This new image will not actually execute the playbook. All it does now is validate that certain environment variables are set and files exist. Eventually, the playbook will do more validation and log playbooks run to a server for further validation.
* The ansible image to use is configurable in controller-manager via command-line arguments.
* contrib/ansible/deploy-devel.yaml has been modified to allow the user to use either the fake image or connect to the real AWS. If the `use_real_aws` variable is set to true when running the playbook, then the real openshift-ansible image will be used, which connects to the real AWS.
* The echo command has been removed from `ansibleRunner` since the playbook execution can be avoided by using the fake image now.
* I have made the fake image the preference at this point because I want to avoid accidental resource allocations on AWS. I expect that once clust op is more mature we will make the default in the controller-manager the real image. I do expect that the development deployment will still prefer the fake image, but that the production deployment will prefer the real image.

### DIFF
--- a/contrib/ansible/deploy-devel.yaml
+++ b/contrib/ansible/deploy-devel.yaml
@@ -5,6 +5,9 @@
   gather_facts: no
   vars:
     aws_section: default
+    use_real_aws: false
+    ansible_image: "fake-openshift-ansible:canary"
+    ansible_image_pull_policy: "Never"
   tasks:
   - name: create cluster-operator namespace
     command: oc create namespace cluster-operator
@@ -25,6 +28,11 @@
     when: cli_aws_section is defined
 
   - set_fact:
+      ansible_image: "openshift/origin-ansible:v3.8"
+      ansible_image_pull_policy: "Always"
+    when: use_real_aws
+
+  - set_fact:
       l_serving_ca: "{{ lookup('file', playbook_dir + '/../../ca.pem') | b64encode }}"
       l_serving_cert: "{{ lookup('file', playbook_dir + '/../../apiserver.pem') | b64encode }}"
       l_serving_key: "{{ lookup('file', playbook_dir + '/../../apiserver-key.pem') | b64encode }}"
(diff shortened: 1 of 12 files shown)


contrib/ansible/deploy-devel.yaml
```diff
@@ -5,6 +5,9 @@
   gather_facts: no
   vars:
     aws_section: default
+    use_real_aws: false
```
dgoodwin: Lets document the use of this in the README.

contrib/fake-openshift-ansible/fake-openshift-ansible
```diff
@@ -0,0 +1,19 @@
...
+
+# Ensure that the inventory file exists.
+[ -z $INVENTORY_FILE ] && { echo "INVENTORY_FILE environment variable not set."; exit 1; }
+[ ! -f "${INVENTORY_FILE}" ] && { echo "Inventory file ${INVENTORY_FILE} not found."; exit 1; }
+
+# Ensure that the AWS access key ID exists.
+[ -z $AWS_ACCESS_KEY_ID ] && { echo "AWS_ACCESS_KEY_ID environment variable not set."; exit 1; }
+
+# Ensure that the AWS secret access key exists.
+[ -z $AWS_SECRET_ACCESS_KEY ] && { echo "AWS_SECRET_ACCESS_KEY environment variable not set."; exit 1; }
```
dgoodwin: AWS vars would only be required for the infra and scale group provisioning, when we get to control plane install and such they wouldn't necessarily be set, but I guess we could just for consistency.

contrib/ansible/deploy-devel.yaml
staebler: Yep. Good catch.

contrib/fake-openshift-ansible/fake-openshift-ansible
staebler: I would assume that we would still pass them into the ansible playbook pod anyway. Otherwise the clust op has to have more knowledge about exactly how the playbooks are implemented.

PR #29, opened 2017-12-19
Basic unit test for infra controller

dgoodwin:
This ended up not being much, will expand over time as we add more logic.


pkg/controller/infra/infra_controller.go
```diff
@@ -100,6 +100,7 @@ func NewInfraController(clusterInformer informers.ClusterInformer, kubeClient ku
 
 	c.syncHandler = c.syncCluster
 	c.enqueueCluster = c.enqueue
+	c.ansibleRunner = ansible.NewAnsibleRunner(kubeClient)
```
staebler: Would you have any qualms about the ansible runner being injected as a parameter to `NewInfraController`? I thought about making that change for #28, but ultimately decided not to. With the changes here, I am leaning again towards providing the ansible runner as a parameter. Note that I am not suggesting making any change for this PR here; I only wanted to know your thoughts on it as a possibility of making the change in my PR.

glide.lock
```diff
@@ -220,6 +220,11 @@ imports:
   version: f62e98d28ab7ad31d707ba837a966378465c7b57
 - name: github.com/spf13/pflag
   version: 9ff6c6923cfffbcd502984b8e0c80539a94968b7
+- name: github.com/stretchr/testify
```
staebler: Some of these testify tests have been duplicated in test/util/assertions.go. It's nice to bring in the real testify so that we can get rid of that assertions.go file. I'll write up an issue to do that.

dgoodwin: Sounds good to me.

pkg/controller/infra/infra_controller.go
dgoodwin: Sounds good to me.

glide.lock
```diff
@@ -608,7 +613,5 @@ testImports:
   version: d8ed2627bdf02c080bf22230dbb337003b7aba2d
   subpackages:
   - difflib
-- name: github.com/stretchr/testify
-  version: f6abca593680b2315d2075e0f5e2a9751e3f431a
-  subpackages:
-  - assert
+- name: github.com/stretchr/objx
```
staebler: Is this dependency actually needed? Did you run `glide up -v` to remove sub-vendors?

dgoodwin: Will check, I was using testify's mock lib briefly working on this and  I think that's what pulled it. It sucked though so it's gone now.

dgoodwin:
Updated after rebase and the objx dep gone.

pkg/ansible/runner.go
```diff
@@ -14,19 +14,25 @@ import (
...
 const (
-	operatorNamespace            = "cluster-operator"
-	openshiftAnsibleContainerDir = "/usr/share/ansible/openshift-ansible/"
-	sshKeySecretName             = "ssh-private-key"
-	awsCredsSecretName           = "aws-credentials"
+	operatorNamespace             = "cluster-operator"
+	openshiftAnsibleContainerDir  = "/usr/share/ansible/openshift-ansible/"
+	sshKeySecretName              = "ssh-private-key"
+	awsCredsSecretName            = "aws-credentials"
+	DefaultAnsibleImage           = "openshift/origin-ansible:v3.8"
```
staebler: These two consts should be removed. I see that they are used in infra_controller_test.go. However, different dummy values can be used there. Having them here gives the impression that they are somehow useful to the production code.

pkg/controller/infra/infra_controller.go
```diff
@@ -109,6 +109,7 @@ func NewInfraController(
 
 	c.syncHandler = c.syncCluster
 	c.enqueueCluster = c.enqueue
+	c.ansibleRunner = ansible.NewAnsibleRunner(c.kubeClient, c.ansibleImage, c.ansibleImagePullPolicy)
```
staebler: Since the ansible runner is created in `NewInfraController`, the `InfraController` no longer needs to hold onto the `ansibleImage` and `ansibleImagePullPolicy`.

dgoodwin:
Thanks @staebler good catches, updated version pushed.

PR #38, opened 2018-01-03
Sync ansible job status with cluster status

csrwng:
Adds a queue to the infra controller to track ansible jobs that need to be watched.
Changes ansible runner into a job generator.


cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -385,20 +390,23 @@ func GetAvailableResources(clientBuilder controller.ClientBuilder) (map[schema.G
 // controllers such as the clientBuilder.
 func CreateControllerContext(s *options.CMServer, clientBuilder controller.ClientBuilder, stop <-chan struct{}) (ControllerContext, error) {
 	versionedClient := clientBuilder.ClientOrDie("shared-informers")
+	kubeClient := clientBuilder.KubeClientOrDie("shared-informers")
```
dgoodwin: Just curious if this name here needs to be unique for any reason?

cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -385,20 +390,23 @@ func GetAvailableResources(clientBuilder controller.ClientBuilder) (map[schema.G
...
 	sharedInformers := clusteroperatorinformers.NewSharedInformerFactory(versionedClient, ResyncPeriod(s)())
+	kubeSharedInformers := kubeinformers.NewSharedInformerFactory(kubeClient, 10*time.Minute)
```
dgoodwin: Could we use the ResyncPeriod function above or is the static ten minutes safe here?

pkg/ansible/jobgenerator.go
```diff
@@ -0,0 +1,191 @@
...
+		Spec: kbatch.JobSpec{
+			Completions:           &completions,
+			ActiveDeadlineSeconds: &deadline,
+			Template: kapi.PodTemplateSpec{
+				Spec: podSpec,
+			},
+		},
+	}
+
+	return job, cfgMap
```
dgoodwin: Returning both rather than creating the config map as well. 👍

pkg/apis/clusteroperator/types.go
```diff
@@ -81,6 +81,14 @@ type AWSClusterSpec struct {
 	// credentials
 	AccountSecret corev1.LocalObjectReference
 
+	// SSHSecret refers to a secret that contains the ssh private key to access
+	// EC2 instances for this cluster
+	SSHSecret corev1.LocalObjectReference
+
+	// SSHSecretName is the name of the secret to use for SSH access to EC2
+	// instances in this cluster
+	SSHSecretName string
```
dgoodwin: Why do we need a SSHSecretName as well as the SSHSecret object ref? Looks like it may be unused.

pkg/ansible/jobgenerator.go
dgoodwin: This now looks extremely unit testable. Should we do this now or perhaps add a trello card to add them in future?

cmd/cluster-operator-controller-manager/app/controllermanager.go
csrwng: it's just the string in the user agent that will show up in logs afaik

pkg/apis/clusteroperator/types.go
csrwng: The secret name is needed so you can use a known secret name like "libra" (if not we can create one for you)

The secret itself is for the ssh private key that we will use when ssh'ing into created instances.

pkg/ansible/jobgenerator.go
csrwng: I haven't added any unit tests at all for this pr yet, but we can do in a follow up pr if that's ok

csrwng:
Updated to use a LocalObjectReference in the cluster status to track the job being used for provisioning

cmd/cluster-operator-controller-manager/app/controllermanager.go
csrwng: I would think that the static is ok, but will check

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -151,6 +159,10 @@ type ClusterStatus struct {
 	// For machine set hardware, see the status of each machine set resource.
 	Provisioned bool `json:"provisioned"`
 
+	// ProvisioningJob is the reference to the Job performing infrastructure provisioning
+	// for this cluster. It will be set while the infrastructure is being provisioned.
+	ProvisioningJob *corev1.LocalObjectReference `json:"provisioningJob,omitempty"`
```
smarterclayton: Is this the only job type?  Is this guaranteed to be nil when Provisioned is set to true?

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -81,6 +81,14 @@ type AWSClusterSpec struct {
 	// credentials
 	AccountSecret corev1.LocalObjectReference `json:"accountSecret"`
 
+	// SSHSecret refers to a secret that contains the ssh private key to access
+	// EC2 instances for this cluster
+	SSHSecret corev1.LocalObjectReference `json:"sshSecret"`
+
+	// SSHSecretName is the name of the secret to use for SSH access to EC2
+	// instances in this cluster
+	SSHSecretName string `json:"sshSecretName"`
```
smarterclayton: Why both?

csrwng: We need both -- this is the name of the ssh secret object in AWS (ie. "libra"). The reference is for a secret that contains the private key for that ssh secret.

pkg/apis/clusteroperator/v1alpha1/types.go
csrwng: > Is this the only job type?
yes (for what we need right now)




> Is this guaranteed to be nil when Provisioned is set to true?

yes

csrwng:
Added handling of job deletes

pkg/apis/clusteroperator/v1alpha1/types.go
smarterclayton: These names are super confusing.  We use SecretName to mean "name of a Kube secret".  Might be better to use something instead of `SecretName` and improve the godoc.

pkg/apis/clusteroperator/v1alpha1/types.go
smarterclayton: Can you add that to the godoc so it's clear?

csrwng: done

cmd/cluster-operator-controller-manager/app/controllermanager.go
staebler: The `ResyncPeriod` does some jittering to reduce the likelihood that each controller will be hitting the API server in lock-step with each other. We should use `ResyncPeriod` for the kube informers, too.

csrwng:
Pushed additional fixes including resync period. Looking into issue of multiple jobs getting created when a status update fails due to conflict. Will likely need a retry with backoff for that.

csrwng:
Now retrying status update on conflict

csrwng:
@stevekuznetsov @kargakis the unit test for this pr seems to be passing ok but it's being reported as failed 
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-operator/38/test_pull_request_cluster_operator_unit/85/

kargakis:
The link in Details points to a failed test

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-operator/38/test_pull_request_cluster_operator_unit/90/
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_cluster_operator_unit/90/

@stevekuznetsov I thought we stopped failing jobs when there was an issue with recording metadata.

stevekuznetsov:
This is a fairly rare bug that is completely inscrutable... the Python script just fails with no ouput. We do run this to make the step not fail when it doesn't run:

```
+ trap 'exit 0' EXIT
```

To attempt to override the exit code but it seems that doesn't work in all cases? Unclear.

pkg/controller/infra/infra_controller.go
```diff
@@ -118,99 +138,369 @@ type InfraController struct {
...
+	cluster, err := c.clusterForJob(job)
+	if err != nil {
+		return fmt.Errorf("could not retrieve cluster for job %s/%s: %v", job.Namespace, job.Name, err)
+	}
+	if cluster == nil {
+		return nil
+	}
+
+	// only proceed if this is the current provisioning job for the cluster
+	if cluster.Status.ProvisioningJob == nil || cluster.Status.ProvisioningJob.Name != job.Name {
```
staebler: I think that we have a race condition here? There is a real possibility that the controller will sync the provision job before getting the update cluster object with its `ProvisionJob` field set. This will cause the controller to skip the sync of the job. It is even possible (although not practically) that the job finished completely before the lister gets the update cluster object. In that worst case scenario, the controller will never sync the job.

pkg/controller/infra/infra_controller.go
```diff
@@ -118,99 +138,369 @@ type InfraController struct {
...
+	}
+
+	return c.syncClusterStatusWithJob(cluster, job)
+}
+
+func (c *InfraController) syncClusterStatusWithJob(cluster *clusteroperator.Cluster, job *v1batch.Job) error {
+	cluster = cluster.DeepCopy()
+	now := metav1.Now()
+	jobCompleted := jobCondition(job, v1batch.JobComplete)
+	if jobCompleted != nil {
```
staebler: Don't we also need to check that `jobCompleted.Status == kapi.ConditionTrue`? Even if we don't have any code that sets a condition of Complete/False, we should still respect that the condition has a status field and not blindly assume that the status will be true,

pkg/controller/infra/infra_controller.go
```diff
@@ -118,99 +138,369 @@ type InfraController struct {
...
+				LastTransitionTime: now,
+				Reason:             "JobCompleted",
+				Message:            fmt.Sprintf("Job %s/%s completed at %v", job.Namespace, job.Name, jobCompleted.LastTransitionTime),
+			})
+			cluster.Status.Provisioned = true
+		}
+		cluster.Status.ProvisioningJob = nil
+	}
+	jobFailed := jobCondition(job, v1batch.JobFailed)
+	if jobFailed != nil {
```
staebler: `jobFailed.Status == kapi.ConditionTrue`?

staebler: Is there value in a condition that states that the provisioning failed? As it stands, the only way to tell if the provisioning failed is to recognize that the provisioning condition is false (or not present) and the provisioned condition is not present. It may be difficult to distinguish this from a cluster that has not attempted provisioning (yet).

pkg/controller/infra/infra_controller.go
```diff
@@ -118,99 +138,369 @@ type InfraController struct {
...
+			clusterProvisioning.Status = kapi.ConditionFalse
+			clusterProvisioning.LastTransitionTime = now
+			clusterProvisioning.LastProbeTime = now
+			clusterProvisioning.Reason = "JobFailed"
+			clusterProvisioning.Message = fmt.Sprintf("Job %s/%s failed at %v, reason: %s", job.Namespace, job.Name, jobFailed.LastTransitionTime, jobFailed.Reason)
+		}
+		cluster.Status.ProvisioningJob = nil
+	}
+
+	if jobCompleted == nil && jobFailed == nil && job.Status.Active > 0 {
```
staebler: I'm not sure about the check for `job.Status.Active > 0` here? If we are going to change the reason for the provisioning condition to JobRunning based upon there being active pods for the job, then we should also change it to something like JobWaiting when there are no active pods for the job. I am not sure whether any check for the number of active pods is really necessary, though, and whether it might be sufficient to just set the reason to JobRunning so long as the job is not completed or failed.

pkg/controller/infra/infra_controller.go
```diff
@@ -118,99 +138,369 @@ type InfraController struct {
...
+				Message:            fmt.Sprintf("Job %s/%s is running since %v. Pod completions: %d, failures: %d", job.Namespace, job.Name, job.Status.StartTime, job.Status.Succeeded, job.Status.Failed),
+			})
+		}
+	}
+
+	return c.updateClusterStatus(cluster)
+}
+
+func (c *InfraController) updateClusterStatus(cluster *clusteroperator.Cluster) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
staebler: I don't think that this is the right thing to do here. A conflict occurs because the controller has a cluster object that is out of date. The request will perpetually fail since the cluster object that the controller has will remain out of date. The controller is just wasting time attempting to retry the request.

pkg/controller/infra/infra_controller.go
```diff
@@ -118,99 +138,369 @@ type InfraController struct {
...
-	if !controller.WaitForCacheSync("cluster", stopCh, c.clustersSynced) {
+	if !controller.WaitForCacheSync("infra", stopCh, c.clustersSynced, c.jobsSynced) {
+		c.logger.Errorf("Could not sync caches for infra controller")
 		return
 	}
 
 	for i := 0; i < workers; i++ {
-		go wait.Until(c.worker, time.Second, stopCh)
+		go wait.Until(c.clusterWorker, time.Second, stopCh)
+		go wait.Until(c.jobWorker, time.Second, stopCh)
```
staebler: I am very cautious against having a separate queue for syncing the jobs. Why can't we just enqueue the cluster associated with the job when the controller notices an add/update/delete for the job? When syncing the cluster, the controller will look at the provisioning job for that cluster. Is the separate queue needed so that we can mark the probe time for the provisioning condition without having that cause another immediate sync of the cluster with a corresponding job check, causing a hot loop. If so, is there some other way to avoid the loop short of using a separate queue?

pkg/controller/infra/infra_controller.go
```diff
@@ -240,10 +530,97 @@ func (c *InfraController) syncCluster(key string) error {
...
+func clusterCondition(cluster *clusteroperator.Cluster, conditionType clusteroperator.ClusterConditionType) *clusteroperator.ClusterCondition {
+	for i, condition := range cluster.Status.Conditions {
+		if condition.Type == conditionType {
+			return &cluster.Status.Conditions[i]
+		}
+	}
+	return nil
+}
 
+func setJobNotFoundStatus(cluster *clusteroperator.Cluster) {
```
staebler: Here is another place where another condition would be helpful to indicate that the provisioning was attempted but failed.

pkg/controller/infra/infra_controller.go
```diff
@@ -252,11 +629,20 @@ func (c *InfraController) syncCluster(key string) error {
...
-		provisionInventoryTemplate, vars)
+	job, cfgMap := c.ansibleGenerator.GeneratePlaybookJob(cluster, jobPrefix, infraPlaybook, provisionInventoryTemplate, vars)
+
+	_, err = c.kubeClient.CoreV1().ConfigMaps(cluster.Namespace).Create(cfgMap)
 	if err != nil {
 		return err
 	}
 
-	return nil
+	_, err = c.kubeClient.BatchV1().Jobs(cluster.Namespace).Create(job)
```
staebler: Are we punting on cleaning up the jobs and configmaps for the time being? When we get to that, we need to make sure that we are resilient to the scenario where the configmap is created successfully but the job fails--ensuring that the configmap is cleaned up in that case.

contrib/examples/cluster.yaml
```diff
@@ -9,6 +9,12 @@ objects:
   metadata:
     name: ${CLUSTER_NAME}
   spec:
+    hardware:
+      aws:
+        accountSecret: 
+          name: aws-credentials
```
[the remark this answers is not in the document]
csrwng: the aws-credentials secret gets created by the ansible playbook that sets up cluster-operator and we tell you to create the ssh secret in the README instructions.

pkg/ansible/jobgenerator.go
```diff
@@ -0,0 +1,191 @@
...
+
+const (
+	openshiftAnsibleContainerDir = "/usr/share/ansible/openshift-ansible/"
+)
+
+var (
+	clusterKind = clusteroperator.SchemeGroupVersion.WithKind("Cluster")
+)
+
+type AnsibleJobGenerator interface {
```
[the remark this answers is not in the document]
csrwng: yup, will fix

pkg/ansible/jobgenerator.go
```diff
@@ -0,0 +1,191 @@
...
+					SecretKeyRef: &kapi.SecretKeySelector{
+						LocalObjectReference: cluster.Spec.Hardware.AWS.AccountSecret,
+						Key:                  "aws_secret_access_key",
+					},
+				},
+			},
+		}...)
+	}
+
+	sshKeyFileMode := int32(0600)
```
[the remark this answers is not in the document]
csrwng: will do

pkg/apis/clusteroperator/types.go
```diff
@@ -151,6 +159,11 @@ type ClusterStatus struct {
 	// For machine set hardware, see the status of each machine set resource.
 	Provisioned bool
 
+	// ProvisioningJob is the reference to the Job performing infrastructure provisioning
+	// for this cluster. It will be set while the infrastructure is being provisioned.
+	// This reference is guaranteed to be nil if Provisioned is true
```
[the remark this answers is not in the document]
csrwng: We're not saying that the job will go away, simply that we won't keep a reference to it in the cluster status when not provisioning the cluster. I added the comment because @smarterclayton asked for it.

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -151,6 +159,11 @@ type ClusterStatus struct {
 	// For machine set hardware, see the status of each machine set resource.
 	Provisioned bool `json:"provisioned"`
 
+	// ProvisioningJob is the reference to the Job performing infrastructure provisioning
+	// for this cluster. It will be set while the infrastructure is being provisioned.
+	// This reference is guaranteed to be nil if Provisioned is true
+	ProvisioningJob *corev1.LocalObjectReference `json:"provisioningJob,omitempty"`
```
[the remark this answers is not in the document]
csrwng: will do

pkg/controller/infra/infra_controller.go
```diff
@@ -118,99 +138,369 @@ type InfraController struct {
...
+
+func (c *InfraController) updateJob(old, obj interface{}) {
+	job := obj.(*v1batch.Job)
+	if c.isInfraJob(job) {
+		c.logger.Debugf("enqueueing updated infra job %s/%s", job.Namespace, job.Name)
+		c.enqueueJob(job)
+	}
+}
+
+func (c *InfraController) deleteJob(obj interface{}) {
```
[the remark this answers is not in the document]
csrwng: yes, that's wrong... it should be referenced by the event handler

pkg/controller/infra/infra_controller.go
csrwng: We can use one queue. The way this PR started, I had no reference from the cluster to the actively running job, which meant that I couldn't just look at the cluster and immediately find the job. Now that we have a ref, I don't see a good reason to keep the 2 queues. Will change.

pkg/controller/infra/infra_controller.go
csrwng: yup, thx, will update

pkg/controller/infra/infra_controller.go
csrwng: yes

pkg/controller/infra/infra_controller.go
csrwng: yeah, I think if it's not completed or failed, it may be safe to assume that is still active

pkg/controller/infra/infra_controller.go
csrwng: for now, yes, we are -- we need to add it to the todo list

pkg/controller/infra/infra_controller.go
csrwng: duh, yes, I do need to retry on conflict, but will need to do a get on the cluster resource

contrib/examples/cluster.yaml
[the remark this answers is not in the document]
staebler: Sure. But the secrets are not created in the right namespace to be used by pods. The old code used to copy the secrets to the namespace in which the cluster resides.

pkg/apis/clusteroperator/types.go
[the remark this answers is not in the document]
staebler: Sure, the job won't go away. But without the reference, it may not be easy for an end user to find the job.

pkg/controller/infra/infra_controller.go
staebler: Can we just fail the sync and pick up the correct cluster object on the next retry? That seems much easier than trying to recreate the changes to the cluster status with the update version. Of course, in that case we would need to handle looking for the job and configmap to make sure that we don't accidentally create multiples. So that may make it more difficult.

staebler: Even if we did get the latest and retry inside a single sync call, we still have the potential to create multiple jobs, so doing it that way doesn't even buy us that simplification.

csrwng:
@staebler @dgoodwin  so I updated the controller to use a single queue and raciness got worse. The main problem is that first I go and create the job, then update the status of the cluster to reflect that it's in a provisioning state. Before the cache gets updated with the latest cluster status, I get the cluster queued again because I got an event for the recently created job. Because I can't tell that the cluster is already provisioning, I launch another job and update the cluster status again. It ends up creating about 3-4 jobs before the cache gets updated and the controller realizes that the job is in a provisioning state. With the 2-queue setup, the worse that can happen is that I miss events from the job to update the cluster. And even though this is very unlikely, I would eventually get a consistent state when the cluster is relisted.

For now I will keep the 2-queue solution and make the other updates to this PR. If you think of something we can do to make this work, please let me know.

staebler:
@csrwng The additional raciness is just exposing a deficiency in the job creation. We need to make a better effort to ensure that the job is not created multiple times. Given the asynchronous nature of kube, we cannot rely solely on the cluster status to determine whether there is an existing job. We must also look in the cache to see if a job exists that is owned by the cluster. This needs to be done regardless of whether we use a single queue or two queues. If that is done, then does that resolve the issues with using a single queue?

csrwng:
> If that is done, then does that resolve the issues with using a single queue?

So, get a list of jobs for the given cluster and determine if among those jobs there's one that's not in a completed state and if that's the case not launch a new job? I would say yes, it resolves the issues.

However, I'm not convinced that that's a better solution than the 2 queues.

csrwng:
> This needs to be done regardless of whether we use a single queue or two queues

Why?

staebler:
> > This needs to be done regardless of whether we use a single queue or two queues
> 
> Why?

Because you still need to guard against trying to reconcile an older version of a cluster object, where the job may not have been set in the status yet. Let's say that the infra controller is reconciling the cluster to create a job and that the cluster controller is simultaneously reconciling the cluster to create node groups.

1) Infra controller syncs cluster with revision 1. Infra controller sees no job in the status, so starts a job. Infra controlller updates status of cluster to indicate that a job has been started.
2) Prior to the API server getting the status update for the cluster from the infra controller, the cluster controller updates the cluster, changing its revision to 2.
3) Infra controller updates ends up changing the revision of the cluster to 3.
4) Infra controller syncs cluster with revision 2. Infra controller sees no job in the status, so starts a new job, which is a duplicate of an existing job.

staebler:
Another option is to use expectations, similar to what is done in the cluster controller. The job starting is then done in two phases. In the first phase, the jobs are started and expectations to see new job objects are set. In the second phase, the infra controller acknowledges that the jobs have been seen.

1) Infra controller syncs a cluster object.
2) Infra controller checks if there are pending expectations for jobs for the cluster object.
3) If there are no pending expectations, then create the job and set the expectation.
4) Look for jobs for the cluster. If one is found and running, set the cluster status to indicate that there is a job running.

csrwng:
> Let's say that the infra controller is reconciling the cluster to create a job and that the cluster controller is simultaneously reconciling the cluster to create node groups.

Thanks, I wasn't thinking of the multiple controllers.

csrwng:
> Another option is to use expectations

I think I like this better

csrwng:
@staebler @dgoodwin ready for another pass
I'm now using the generation field to determine whether a job should be generated or not. I'm also using the single queue with expectations.

pkg/apis/clusteroperator/types.go
```diff
@@ -151,6 +166,17 @@ type ClusterStatus struct {
 	// For machine set hardware, see the status of each machine set resource.
 	Provisioned bool
 
+	// ProvisioningJobGeneration is the generation of the cluster resource used to
+	// launch the last provisioning job.
+	// +optional
```
staebler: I am thinking that this should not be `+optional`. Even when a provisioning job has not been started yet, you could theoretically consider that the generation that was provisioned is the 0 generation. If it is optional, then the kube API conventions stipulate that the field should be a `*int64`.

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -151,6 +166,17 @@ type ClusterStatus struct {
 	// For machine set hardware, see the status of each machine set resource.
 	Provisioned bool `json:"provisioned"`
 
+	// ProvisioningJobGeneration is the generation of the cluster resource used to
+	// launch the last provisioning job.
+	// +optional
+	ProvisioningJobGeneration int64 `json:"provisioningJobGeneration,omitempty"`
```
staebler: If we remove `+optional`, remember to also remove `omitempty`.

pkg/controller/infra/infra_controller.go
```diff
@@ -201,18 +366,145 @@ func (c *InfraController) handleErr(err error, key interface{}) {
...
+func (c *InfraController) syncClusterStatusWithJob(cluster *clusteroperator.Cluster, job *v1batch.Job) error {
+	cluster = cluster.DeepCopy()
+	now := metav1.Now()
+
+	// Ensure that job status reflects current job
+	if cluster.Status.ProvisioningJob == nil {
+		cluster.Status.ProvisioningJob = &kapi.LocalObjectReference{
+			Name: job.Name,
+		}
+		cluster.Status.ProvisioningJobGeneration = jobClusterGeneration(job)
```
staebler: I think we need to set `ProvisioningJobGeneration` when the job is created and use comparison between the cluster generation stored in the job's annotations and the current generation of the cluster to determine whether a job matches up with the current state of the cluster for updating conditions.

pkg/controller/infra/infra_controller.go
```diff
@@ -234,27 +526,154 @@ func (c *InfraController) syncCluster(key string) error {
...
+		return nil
+	}
+
+	// Find provisioning job
+	var provisioningJob *v1batch.Job
+	jobs, err := c.jobsLister.Jobs(cluster.Namespace).List(labels.Everything())
+	if err != nil {
+		return err
+	}
+	for _, job := range jobs {
```
staebler: We need to do something about outstanding jobs after the cluster spec changes. I think that we don't want to have two jobs running simultaneously for a cluster, so we need to wait for outstanding jobs to finish before starting new jobs. We need to terminate old jobs so that they don't attempt to create more pods in case the old cluster spec is such that the job will never complete successfully.

pkg/controller/infra/infra_controller.go
```diff
@@ -201,18 +366,145 @@ func (c *InfraController) handleErr(err error, key interface{}) {
...
+				LastProbeTime:      now,
+				LastTransitionTime: now,
+				Reason:             "JobFailed",
+				Message:            fmt.Sprintf("Job %s/%s failed at %v, reason: %s", job.Namespace, job.Name, jobFailed.LastTransitionTime, jobFailed.Reason),
+			})
+		}
+	default:
+		clusterProvisioning := clusterCondition(cluster, clusteroperator.ClusterHardwareProvisioning)
+		reason := "JobRunning"
+		message := fmt.Sprintf("Job %s/%s is running since %v. Pod completions: %d, failures: %d", job.Namespace, job.Name, job.Status.StartTime, job.Status.Succeeded, job.Status.Failed)
```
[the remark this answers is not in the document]
csrwng: I will remove the pod info from the message. We'd said that we wouldn't update the running condition every time, since it would just cause a hot loop of updates.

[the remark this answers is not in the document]
staebler: If we only update it when the message changes, it won't hot loop. It would be nice to see this information in the condition. I'm fine with either way--removing the info or updating the info.

pkg/controller/infra/infra_controller.go
```diff
@@ -234,27 +526,154 @@ func (c *InfraController) syncCluster(key string) error {
...
+	}
+
+	// Find provisioning job
+	var provisioningJob *v1batch.Job
+	jobs, err := c.jobsLister.Jobs(cluster.Namespace).List(labels.Everything())
+	if err != nil {
+		return err
+	}
+	for _, job := range jobs {
+		if isInfraJob(job) && isController(cluster, job) && isActive(job) {
```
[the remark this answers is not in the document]
csrwng: So instead of checking whether the job is active, check whether its generation matches Status.ProvisioningJobGeneration ?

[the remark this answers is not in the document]
staebler: Yes, for determining whether a particular job should be considered as the "current" provisioning job.

pkg/controller/infra/infra_controller.go
```diff
@@ -233,28 +525,176 @@ func (c *InfraController) syncCluster(key string) error {
...
+	return nil
+}
+
+func (c *InfraController) deleteJobs(clusterKey string, jobs []*v1batch.Job) {
+	for _, job := range jobs {
+		err := c.kubeClient.BatchV1().Jobs(job.Namespace).Delete(job.Name, &metav1.DeleteOptions{})
+		if err != nil {
+			c.logger.WithField("job", jobKey(job)).Warningf("error deleting job: %v", err)
+			utilruntime.HandleError(err)
+			c.expectations.DeletionObserved(clusterKey, jobKey(job))
```
staebler: We are not tracking job deletes, so there is no reason to register observations of job deletes.

PR #41, opened 2018-01-08
Add scaffolding for separate cluster version resources.

dgoodwin:
Cluster versions define the set of OpenShift versions the operator can
install. They will be stored as independent resources, cluster specs
will reference them. A cluster's version reference is mutable, but
machine sets and potentially other objects will have immutable copies of
the data. Changing a cluster version should result in the creation of
new machine sets with the updated version info.

This change just adds the basic framework for defining cluster versions.

Cluster versions are mutable at this point, this may change in future.


pkg/apis/clusteroperator/types.go
```diff
@@ -48,6 +48,48 @@ type ClusterList struct {
...
+	metav1.TypeMeta
+	// +optional
+	metav1.ListMeta
+
+	Items []ClusterVersion
+}
+
+// VMImages contains required data to locate images in all supported cloud providers.
+type VMImages struct {
+	AWSImages AWSVMImages
```
csrwng: Should likely be a pointer and made optional

pkg/apis/clusteroperator/types.go
```diff
@@ -48,6 +48,48 @@ type ClusterList struct {
...
+	// +optional
+	metav1.ObjectMeta
+
+	// Possible future additions: RPMRepositories, openshift_release / image_tag / pkg_version.
+
+	// ImageFormat defines the pattern string for finding required OpenShift images.
+	// (i.e. "openshift/origin-${component}:v3.7.9", or
+	// "ci.openshift.org/ci-jobs/pr-2034834098243:${component}")
+	// Converts to oreg_url for openshift-ansible.
+	ImageFormat string
```
csrwng: I'm thinking that we should follow convention and have a Spec and Status structure. At some point we may want to have a controller that works on version objects to auto-discover things like AMI names.

csrwng:
Also, do we not want to track rpm repo as well?

pkg/apis/clusteroperator/types.go
dgoodwin: Ok good idea, I will add.

dgoodwin:
RPM repo seems likely but my plan was to just add it when it's needed, I'm not 100% sure if/when given our upcoming containers only focus and with golden images.

contrib/examples/cluster.yaml
```diff
@@ -20,6 +20,23 @@ objects:
...
+      imagesByRegion:
+        us-east-1: ami-10aed306a # TODO: not correct, need to find out how to lookup latest golden image.
+  yumRepositories:
+  # TODO: replace with a usable OpenShift repo
+  - id: dgoodwin-tito-copr
+    name: Tito Development Copr
+    baseurl: https://copr-be.cloud.fedoraproject.org/results/dgoodwin/tito/fedora-$releasever-$basearch/
+    gpgkey: https://copr-be.cloud.fedoraproject.org/results/dgoodwin/tito/pubkey.gpg
+    gpgcheck: 1
+    enabled: 0
```
dgoodwin: Note these keys are abnormal as far as kube is concerned, but match what we need to serialize too for openshift-ansible and ultimately yum.

pkg/registry/clusteroperator/clusterversion/strategy.go
```diff
@@ -0,0 +1,124 @@
...
+	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/registry/rest"
+	"k8s.io/apiserver/pkg/storage/names"
+
+	"github.com/golang/glog"
+	"github.com/openshift/cluster-operator/pkg/apis/clusteroperator"
+	"github.com/openshift/cluster-operator/pkg/apis/clusteroperator/validation"
+)
+
+// NewScopeStrategy returns a new NamespaceScopedStrategy for clusters
```
staebler: s/clusters/clusterversions/

pkg/registry/clusteroperator/clusterversion/strategy.go
```diff
@@ -0,0 +1,124 @@
...
+func NewScopeStrategy() rest.NamespaceScopedStrategy {
+	return clusterRESTStrategies
+}
+
+// implements interfaces RESTCreateStrategy, RESTUpdateStrategy, RESTDeleteStrategy,
+// NamespaceScopedStrategy
+type clusterRESTStrategy struct {
```
staebler: s/cluster/clusterversion/

pkg/registry/clusteroperator/clusterversion/strategy.go
```diff
@@ -0,0 +1,124 @@
...
+	names.NameGenerator // GenerateName method for CreateStrategy
+}
+
+// implements interface RESTUpdateStrategy
+type clusterVersionStatusRESTStrategy struct {
+	clusterRESTStrategy
+}
+
+var (
+	clusterRESTStrategies = clusterRESTStrategy{
```
staebler: s/cluster/clusterversion/g

pkg/registry/clusteroperator/clusterversion/strategy.go
```diff
@@ -0,0 +1,124 @@
...
+func (clusterRESTStrategy) Canonicalize(obj runtime.Object) {
+	_, ok := obj.(*clusteroperator.ClusterVersion)
+	if !ok {
+		glog.Fatal("received a non-cluster version object to create")
+	}
+}
+
+// NamespaceScoped returns true as cluster versions are scoped to a namespace.
+func (clusterRESTStrategy) NamespaceScoped() bool {
+	return true
```
staebler: Have we thought through the RBAC for accessing cluster versions? I am inclined to think that there will be cluster versions established by an administrator and used by users creating clusters in their own namespaces.

pkg/registry/clusteroperator/clusterversion/strategy.go
```diff
@@ -0,0 +1,124 @@
...
+	if !ok {
+		glog.Fatal("received a non-cluster version object to update to")
+	}
+	_, ok = old.(*clusteroperator.ClusterVersion)
+	if !ok {
+		glog.Fatal("received a non-cluster version object to update from")
+	}
+}
+
+func (clusterRESTStrategy) ValidateUpdate(ctx genericapirequest.Context, new, old runtime.Object) field.ErrorList {
```
staebler: Need to add a `PrepareForUpdate` and `ValidateUpdate` for `clusterVersionStatusRESTStrategy`.

pkg/registry/clusteroperator/clusterversion/strategy.go
```diff
@@ -0,0 +1,124 @@
...
+		glog.Fatal("received a non-cluster version object to create")
+	}
+}
+
+// NamespaceScoped returns true as cluster versions are scoped to a namespace.
+func (clusterRESTStrategy) NamespaceScoped() bool {
+	return true
+}
+
+func (clusterRESTStrategy) PrepareForCreate(ctx genericapirequest.Context, obj runtime.Object) {
```
staebler: Need to clear out the status. A user cannot set the status when creating an object.

pkg/registry/clusteroperator/clusterversion/strategy.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (clusterRESTStrategy) AllowCreateOnUpdate() bool {
+	return false
+}
+
+func (clusterRESTStrategy) AllowUnconditionalUpdate() bool {
+	return false
+}
+
+func (clusterRESTStrategy) PrepareForUpdate(ctx genericapirequest.Context, new, old runtime.Object) {
```
staebler: Need to clear out any changes made to the status. The user cannot update the status when doing a spec update.

pkg/registry/clusteroperator/rest/storage_cluster_operator_test.go
```diff
@@ -78,6 +78,11 @@ func TestV1Alpha1Storage(t *testing.T) {
 		t.Fatalf("no clusters/status storage found")
 	}
 
+	_, clusterVersionStorageExists := storageMap["clusterversions"]
```
staebler: Add check for "clusterversions/status".

pkg/apis/clusteroperator/types.go
```diff
@@ -48,6 +48,67 @@ type ClusterList struct {
...
+// ClusterVersion represents a version of OpenShift that can be, or is installed.
+type ClusterVersion struct {
+	// +optional
+	metav1.TypeMeta
+	// +optional
+	metav1.ObjectMeta
+
+	// Possible future additions: openshift_release / image_tag / pkg_version.
+
+	YumRepositories []YumRepositories
```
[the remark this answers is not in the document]
dgoodwin: It should definitely be optional, long term we shouldn't need this at all. I need to get some clarification on what // +optional does exactly because I would assume this is optional as it stands right now, it can be created without.

pkg/apis/clusteroperator/types.go
```diff
@@ -48,6 +48,67 @@ type ClusterList struct {
...
+
+	// ImageFormat defines the pattern string for finding required OpenShift images.
```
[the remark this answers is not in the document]
dgoodwin: The two would be unrelated, ImagesByRegion would be VM images, that is differentiated on a higher struct but perhaps it should be on ImagesByRegion as well.

As for the format of this I will add some clarification.

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -48,6 +48,70 @@ type ClusterList struct {
...
+}
+
+// YumRepositories represents optional yum repositories to deploy onto all systems in the cluster.
+type YumRepositories struct {
+	// Using lowercase for JSON as this matches the properties in the actual yum repo / openshift-ansible structure:
+	ID      string `json:"id"`
+	Name    string `json:"name"`
+	BaseURL string `json:"baseurl"`
+	GPGKey  string `json:"gpgkey"`
+	// Enabled is technically a boolean but represented as an int (0/1) to match the syntax we need to serialize too.
```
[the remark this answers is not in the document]
dgoodwin: My thought process was that you look at a cluster version serialized right now, and you see exactly a yum repo, the properties match perfectly, and anyone who is familiar with them will easily know what it is and how to define one. We're also very ready to serialize for openshift-ansible's properties.

We can change property names and the ints to bools and then do some serialization magic, but it didn't feel quite right to do so just for superficial reasons when the end result might actually be more confusing to a user creating one of these things.

pkg/apis/clusteroperator/validation/cluster.go
```diff
@@ -128,3 +131,18 @@ func ValidateClusterStatusUpdate(new *clusteroperator.Cluster, old *clusteropera
...
+
+// ValidateClusterVersion validates a cluster version being created.
+func ValidateClusterVersion(cv *clusteroperator.ClusterVersion) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	for _, msg := range ValidateClusterVersionName(cv.Name, false) {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("name"), cv.Name, msg))
+	}
+
+	if cv.ImageFormat == "" {
```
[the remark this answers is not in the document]
dgoodwin: Probably, I will look into it.

pkg/registry/clusteroperator/clusterversion/strategy.go
dgoodwin: No details on the RBAC yet other than that we wanted to do what you describe, define them in a known NS by an admin, other users get read-only access.

pkg/apis/clusteroperator/types.go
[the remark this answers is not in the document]
dgoodwin: We don't even really document this here: https://docs.openshift.com/container-platform/latest/install_config/install/advanced_install.html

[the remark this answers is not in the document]
dgoodwin: Going to go with AMIByRegion to clarify that var name.

pkg/apis/clusteroperator/types.go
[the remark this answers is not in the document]
staebler: If it can be created without it, then it is optional. See https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required.

pkg/apis/clusteroperator/v1alpha1/types.go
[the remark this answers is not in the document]
staebler: OK. I can accept that. Let's make sure that we are adding appropriate validation on these (and other) fields.

pkg/apis/clusteroperator/types.go
```diff
@@ -48,6 +48,67 @@ type ClusterList struct {
...
+	// (i.e. "openshift/origin-${component}:v3.7.9", or
+	// "ci.openshift.org/ci-jobs/pr-2034834098243:${component}")
+	// Converts to oreg_url for openshift-ansible.
+	ImageFormat string
+
+	VMImages VMImages
+	Status   ClusterVersionStatus
```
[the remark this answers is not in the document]
dgoodwin: Just the status I assume, vmimages assumed required for now seem reasonable?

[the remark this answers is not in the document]
staebler: Yes, the comment is referring just to `Status`. When you move the other stuff into a `Spec`, though, the `Spec` will also be optional.

pkg/apis/clusteroperator/validation/cluster.go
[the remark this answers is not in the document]
dgoodwin: Dug into this, we could go through the hassle of parsing out all the expansions and making sure that only component and version are there but I think we should pass on this for now, it's not particularly likely someone's just going to make something up, the real threat is that the images actually exist and we're not really a point where we want to validate that ourselves yet.

contrib/examples/cluster.yaml
```diff
@@ -20,6 +20,24 @@ objects:
...
+- apiVersion: clusteroperator.openshift.io/v1alpha1
+  kind: ClusterVersion
+  metadata:
+    name: v3-9
+  spec:
+    imageFormat: "openshift/origin-${component}:v3.9.0-alpha.1"
+    vmImages:
+      awsVMImages:
+        amiByRegion:
+          us-east-1: ami-10aed306a # TODO: not correct, need to find out how to lookup latest golden image.
```
csrwng: I found ami-9a0d95e0 to be the latest one

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -48,6 +48,78 @@ type ClusterList struct {
...
+	// +optional
+	Spec ClusterVersionSpec `json:"spec,omitempty"`
+	// +optional
+	Status ClusterVersionStatus `json:"status,omitempty"`
+}
+
+// ClusterVersionSpec is a specification of a cluster version that can be installed.
+type ClusterVersionSpec struct {
+	// +optional
+	YumRepositories []YumRepository `json:"yumRepositories,omitempty"`
```
staebler: In follow-on work, we need to add comments to all of these new fields.

pkg/apis/clusteroperator/validation/clusterversion.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/openshift/cluster-operator/pkg/apis/clusteroperator"
+)
+
+// ValidateClusterVersionName validates the name of a cluster version.
+var ValidateClusterVersionName = apivalidation.ValidateClusterName
```
staebler: Is there any reason why the ClusterVersion name needs the extra restrictions impose on Cluster names?

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -48,6 +48,78 @@ type ClusterList struct {
...
+}
+
+// AWSVMImages indicates which AMI to use in each supported AWS region for this OpenShift version.
+type AWSVMImages struct {
+	AMIByRegion map[string]string `json:"amiByRegion"`
+}
+
+// YumRepository represents optional yum repositories to deploy onto all systems in the cluster.
+type YumRepository struct {
+	// Using lowercase for JSON as this matches the properties in the actual yum repo / openshift-ansible structure:
```
staebler: This is probably not the best place for this comment as it will show up associated with the `ID` field. It seems better for it to be associated with the `YumRepository` type. But it might not be appropriate to show up for the end user at all.

pkg/apis/clusteroperator/validation/clusterversion.go
```diff
@@ -0,0 +1,89 @@
...
+
+func ValidateClusterVersionSpec(spec *clusteroperator.ClusterVersionSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if spec.ImageFormat == "" {
+		allErrs = append(allErrs, field.Required(field.NewPath("imageFormat"), "must define image format"))
+	}
+
+	reposPath := fldPath.Child("yumRepositories")
+	for i := range spec.YumRepositories {
+		r := spec.YumRepositories[i]
```
staebler: Why not include `r` in the left side of the `for` statement?

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -48,6 +48,78 @@ type ClusterList struct {
...
+	ID      string `json:"id"`
+	Name    string `json:"name"`
+	BaseURL string `json:"baseurl"`
+	// Enabled is technically a boolean but represented as an int (0/1) to match the syntax we need to serialize too.
+	// Enabled controls whether or not the repository should be enabled on the end system. Must be 0 or 1 to match yum.
+	Enabled int `json:"enabled"`
+	// GPGCheck controls whether or not the packages in the repository should have their GPG signatures validated. Must be 0 or 1 to match yum.
+	GPGCheck int `json:"gpgcheck"`
+	// +optional
+	GPGKey string `json:"gpgkey"`
```
staebler: Add `omitempty`.

pkg/apis/clusteroperator/validation/clusterversion.go
```diff
@@ -0,0 +1,89 @@
...
+// ValidateClusterVersionStatusUpdate validates an update to the status of a clusterversion.
+func ValidateClusterVersionStatusUpdate(new *clusteroperator.ClusterVersion, old *clusteroperator.ClusterVersion) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// No validation required yet as the status struct is currently just a placeholder.
+
+	return allErrs
+}
+
+func ValidateClusterVersionSpec(spec *clusteroperator.ClusterVersionSpec, fldPath *field.Path) field.ErrorList {
```
staebler: Add validation that at least one of the fields in `VMImages` is non-nil.

pkg/registry/clusteroperator/clusterversion/storage_test.go
```diff
@@ -0,0 +1,141 @@
...
+		ResourcePrefix:          "clustersversions",
+	}
+	clusterVersionStorage, _ := NewStorage(restOptions)
+	return clusterVersionStorage, server
+}
+
+func validClusterVersion(name string) *clusteroperatorapi.ClusterVersion {
+	return &clusteroperatorapi.ClusterVersion{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "helloworld",
```
staebler: Do you mean to use the `name` parameter as the name of the `ClusterVersion`? Was this changed to "helloworld" to overcome "v3.9" not being a valid name?

pkg/registry/clusteroperator/clusterversion/strategy.go
```diff
@@ -0,0 +1,156 @@
...
+	if !ok {
+		glog.Fatal("received a non-cluster version object to validate to")
+	}
+	_, ok = old.(*clusteroperator.ClusterVersion)
+	if !ok {
+		glog.Fatal("received a non-cluster version object to validate from")
+	}
+
+	return field.ErrorList{}
+	//return validation.ValidateClusterUpdate(newCluster, oldCluster)
```
staebler: Why are we not doing any validation on updates?

pkg/apis/clusteroperator/validation/clusterversion.go
dgoodwin: No probably not, just working off a bad example, I will remove this. Why do we do extra validation for clusters? I assume something dns related?

pkg/apis/clusteroperator/validation/clusterversion.go
dgoodwin: It's probably going to be all nil in an on-premise use case, but we can validate that it's there for now, might save someone some time.

pkg/registry/clusteroperator/clusterversion/strategy.go
dgoodwin: Will add this in a followup.

pkg/apis/clusteroperator/validation/clusterversion.go
staebler: That's a good question. The `ClusterName` field in `ObjectMeta` has that extra validation. I assumed that cluster-operator would reify a cluster with the same name as the name of the `Cluster` object, which would in turn populate the `ClusterName` field of objects in that cluster. However, it appears by the comments on `ClusterName`, that that field is not actually used.

pkg/apis/clusteroperator/validation/clusterversion.go
staebler: If it's going to be all nil, then the `VMImages` field itself should probably be optional. I would find it frustrating as a user to be forced to supply a `VMImages` field with nothing in it.

pkg/registry/clusteroperator/clusterversion/strategy.go
staebler: Can we at least call `ValidateClusterVersion` here, even if we are not going to do any validation of changes until a follow-up?

dgoodwin:
Updated based on feedback aside from a couple items that will arrive in followup PRs.

pkg/registry/clusteroperator/clusterversion/storage_test.go
```diff
@@ -0,0 +1,148 @@
...
+		ResourcePrefix:          "clustersversions",
+	}
+	clusterVersionStorage, _ := NewStorage(restOptions)
+	return clusterVersionStorage, server
+}
+
+func validClusterVersion(name string) *clusteroperatorapi.ClusterVersion {
+	return &clusteroperatorapi.ClusterVersion{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "v3-7",
```
staebler: The point of my previous comment here was not to change "helloworld" to some other constant value. My point was to change to using the `name` value passed into the function.

dgoodwin:
Couple more small updates, version name, and hooked up update to the basic create validation.

Issue #40, opened 2018-01-08
controller-manager not waiting for api registration


staebler:
With the move to kube 1.9, the controller-manager is no longer waiting for the cluster-operator api to be registered with the aggregation api server prior to attempting to start the controllers. Consequently, the controllers are skipped as their dependent resources do not exist.

PR #42, opened 2018-01-09
Delay starting controllers

staebler:
Wait for the cluster-operator.openshift.io resource group to be registered with the aggregation API server prior to starting controllers.


staebler:
Fixes #40.

PR #43, opened 2018-01-15
Create type for handling jobs for controllers

staebler:
`JobControl` can be re-used by other controllers (master, machineset) that need to start jobs to do work.

### DIFF
--- a/pkg/controller/infra/infra_controller.go
+++ b/pkg/controller/infra/infra_controller.go
@@ -496,9 +496,9 @@ func (c *InfraController) syncCluster(key string) error {
 
 	specChanged := cluster.Status.ProvisionedJobGeneration != cluster.Generation
 
-	jobFactory := c.getJobFactory(cluster, specChanged)
+	jobFactory := c.getJobFactory(cluster)
 
-	job, isJobNew, err := c.jobControl.ControlJobs(key, cluster, jobFactory, cLog)
+	job, isJobNew, err := c.jobControl.ControlJobs(key, cluster, specChanged, jobFactory, cLog)
 	if err != nil {
 		return err
 	}
@@ -604,10 +604,7 @@ func (f jobFactory) BuildJob(name string) (*v1batch.Job, *kapi.ConfigMap, error)
 	return f(name)
 }
 
-func (c *InfraController) getJobFactory(cluster *clusteroperator.Cluster, specChanged bool) controller.JobFactory {
-	if !specChanged {
-		return nil
-	}
+func (c *InfraController) getJobFactory(cluster *clusteroperator.Cluster) controller.JobFactory {
 	return jobFactory(func(name string) (*v1batch.Job, *kapi.ConfigMap, error) {
 		varsGenerator := ansible.NewVarsGenerator(cluster)
 		vars, err := varsGenerator.GenerateVars()
--- a/pkg/controller/infra/infra_controller.go
+++ b/pkg/controller/infra/infra_controller.go
@@ -18,15 +18,12 @@ package infra
 
 import (
 	"fmt"
-	"strconv"
-	"strings"
 	"time"
 
 	v1batch "k8s.io/api/batch/v1"
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	batchinformers "k8s.io/client-go/informers/batch/v1"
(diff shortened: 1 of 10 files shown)


pkg/controller/jobcontrol.go
```diff
@@ -0,0 +1,271 @@
...
+
+	return newJob, nil
+}
+
+func (c *jobControl) deleteOldJobs(ownerKey string, oldJobs []*kbatch.Job, logger log.FieldLogger) error {
+	keysToDelete := []string{}
+	for _, job := range oldJobs {
+		keysToDelete = append(keysToDelete, jobKey(job))
+	}
+	c.expectations.ExpectDeletions(ownerKey, keysToDelete)
```
csrwng: check return error

pkg/controller/jobcontrol.go
```diff
@@ -0,0 +1,271 @@
...
+
+	logger.Infof("Deleting old jobs: %v", keysToDelete)
+
+	var errCh chan error
+
+	var wg sync.WaitGroup
+	wg.Add(len(oldJobs))
+	for i, ng := range oldJobs {
+		go func(ix int, job *kbatch.Job) {
+			defer wg.Done()
```
csrwng: if there's an error, it doesn't look like we'd ever call this

pkg/controller/jobcontrol.go
```diff
@@ -0,0 +1,271 @@
...
+				logger.Infof("Failed to delete %v, decrementing expectations", jobKey)
+				c.expectations.DeletionObserved(ownerKey, jobKey)
+				errCh <- err
+			}
+		}(i, ng)
+	}
+	wg.Wait()
+
+	select {
+	case err := <-errCh:
```
csrwng: close errCh so we don't leak goroutines

pkg/controller/infra/infra_controller.go
```diff
@@ -525,101 +486,68 @@ func (c *InfraController) syncCluster(key string) error {
...
 	}
 
-	if !c.expectations.SatisfiedExpectations(key) {
-		// expectations have not been met, come back later
-		cLog.Debugln("cluster has not satisfied expectations yet")
-		return nil
-	}
+	specChanged := cluster.Status.ProvisionedJobGeneration != cluster.Generation
+
+	jobFactory := c.getJobFactory(cluster, specChanged)
```
csrwng: putting the logic in the job factory creation to decide whether we should create a job or not doesn't seem very explicit. I would rather pass a function to the JobControl interface that would decide, given an owner object, whether a new job is needed or not.

staebler: I am hesitant to supply a function when creating the `JobControl` that would be used to determine whether a new job is needed. I could easily see cases where the decision is based on more than just information in the owner. Would it be OK if I pass an `isJobNeeded` boolean to `ControlJobs`?

pkg/controller/jobcontrol.go
```diff
@@ -0,0 +1,271 @@
...
+	if configMap != nil {
+		configMap.OwnerReferences = append(configMap.OwnerReferences, *ownerRef)
+
+		configMap, err = c.kubeClient.CoreV1().ConfigMaps(owner.GetNamespace()).Create(configMap)
+		if err != nil {
+			return nil, err
+		}
+		defer func() {
+			if configMap != nil && cleanUpConfigMap {
+				if err := c.kubeClient.CoreV1().ConfigMaps(configMap.Namespace).Delete(configMap.Name, &metav1.DeleteOptions{}); err != nil {
```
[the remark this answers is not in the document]
staebler: We should make an effort to delete the configmap with the same name when we delete the job. I decided not to do that as part of this PR to limit the new functionality introduced by this PR. I will add an issue to make sure that this is tracked for future work.

I don't think that we need to *watch* configmaps. The point of watching job deletes is so that we know when it is safe to start a new job. We don't need to block anything else due to the existence of a configmap.

Note that this delete here is only to clean up the configmap in the case that there was an error that prevented the successful creation of the job.

pkg/controller/jobcontrol.go
staebler: I'm not following. Doesn't the defer *always* get called, error or not?

pkg/controller/jobcontrol.go
staebler: Why would we leak goroutines? There are no goroutines that are blocking on `errCh`.

pkg/controller/jobcontrol.go
staebler: Is your issue that the `errCh` is not being created? That is certainly a problem that I should resolve. It is also a problem in `clusterController`.

pkg/controller/jobcontrol.go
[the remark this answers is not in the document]
csrwng: ic, thx that's fine

pkg/controller/jobcontrol.go
csrwng: the problem is that sending to errCh will block until you read from it, but you won't get to the part that reads from it until all goroutines have called wg.Done(), essentially deadlocking

pkg/controller/infra/infra_controller.go
csrwng: sgtm

pkg/controller/jobcontrol.go
csrwng: you are sending to it, no?

pkg/controller/jobcontrol.go
staebler: Sending to `errCh` won't block unless `errCh` is already full. The size of `errCh` is now set to the length of `oldJobs`, so the sender will never block.

pkg/controller/jobcontrol.go
staebler: I am sending to it, but the send will never block.

pkg/controller/jobcontrol.go
csrwng: ahh buffered channel :) 👍

csrwng: I immediately assumed it was unbuffered, but I guess you were missing the make call

PR #45, opened 2018-01-17
Use latest tag for origin-ansible which now contains boto deps.

dgoodwin:
This change hasn't yet appeared in the v3.9 tag so latest will have to
do for now.


dgoodwin:
Tested this morning, working well doing actual AWS provisioning and will use for the demo.

PR #47, opened 2018-01-18
Immutable ClusterVersions, and references on cluster objects

dgoodwin:
Adds immutable cluster version specs.

Adds a version reference on ClusterSpec, should be mutable, changing it will eventually
trigger a cluster upgrade.


dgoodwin:
There's no guarantee here that the cluster version referenced exists, should we verify this before allowing it to be set?

Or should we verify later when trying to use, and what is the kube convention for indicating that kind of an error to the user? (events I assume?)

dgoodwin:
Expanded to include immutable cluster versions.

Working on adding the MachineSet reference now but may do a separate PR is this gets reviewed in meantime.

staebler:
> There's no guarantee here that the cluster version referenced exists, should we verify this before allowing it to be set?
> 
> Or should we verify later when trying to use, and what is the kube convention for indicating that kind of an error to the user? (events I assume?)

I do not think that we should verify that the ClusterVersion exists prior to allowing it to be set. We want to be able to add objects in an undetermined order and have everything reconcile correctly eventually. Any controller that is relying on the ClusterVersion reference pointing to an existing ClusterVersion should verify its existence. If the ClusterVersion does not exist at that point, then a Condition should be added to the Cluster stating as much. Additionally, an event should be added with that information. This behavior should be similar to what happens when one of the secrets in `AWSClusterSpec` does not exist. Which should be similar to what happens when you try to create a pod with a reference to a non-existent secret.

pkg/apis/clusteroperator/types.go
```diff
@@ -140,6 +140,9 @@ type ClusterSpec struct {
 
 	// MachineSets specifies the configuration of all machine sets for the cluster
 	MachineSets []ClusterMachineSet
+
+	// Version references the clusterversion that should be running.
+	Version corev1.ObjectReference
```
staebler: I don't think that we want this to be an `ObjectReference`. My understanding is that an `ObjectReference` is used when the actual object has been found. I think we want to use something akin to `LocalObjectReference`, where all we are supplying is the name and the namespace. The user should not have to supply the Kind for the object. And the user should not be confused by the presence of a UID field that does not have any effect.

dgoodwin: Looking at the source for the types ObjectReference was doc'd with: "ObjectReference contains enough information to let you inspect or modify the referred object", LocalObject states it's just for within one namespace and has a TODO to add Kind and UID.

Are you proposing adding our own ObjectReference type that has a namespace but no kind?

staebler: Yes, that is precisely what I am proposing. `ObjectReference` is great in a status, where a controller is filling out the information for an existing object. In a spec, it is sub-par because it either (1) requires the user to add a lot of information that is *always* and *required* to be the same or (2) includes fields that the user *cannot* utilize and *ignored* if filled out.

dgoodwin: Ok that sounds fine to me. Will add.

dgoodwin:
Ok this should be ready for review now, I've introduced a ClusterVersionReference,  added ClusterVersion to MachineSet and ensured the reference there is immutable, and changing it on the ClusterSpec will trigger creation of new machine sets.

dgoodwin:
Both items should now be addressed.

contrib/examples/cluster.yaml
```diff
@@ -8,6 +8,7 @@ objects:
   kind: ClusterVersion
   metadata:
     name: ocp-v3-9
+    namespace: cluster-operator
```
staebler: The namespaces for these ClusterVersions will be stripped out. To retain the namespace, it needs to come from a parameter.

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -149,8 +149,7 @@ type ClusterSpec struct {
 
 // ClusterVersionReference provides information to locate a cluster version to use.
 type ClusterVersionReference struct {
-	// Namespace of the clusterversion. Assumed to be in the same namespace if empty.
-	// +optional
+	// Namespace of the clusterversion.
 	Namespace string `json:"namespace,omitempty"`
```
staebler: Remove `omitempty`.

pkg/apis/clusteroperator/validation/cluster.go
```diff
@@ -76,6 +77,10 @@ func validateClusterSpec(spec *clusteroperator.ClusterSpec, fldPath *field.Path)
 		allErrs = append(allErrs, field.Invalid(machineSetsPath, &spec.MachineSets, "must have one machineset that hosts infra pods"))
 	}
 
+	if len(spec.Version.Name) == 0 {
```
staebler: Add validation that `Namespace` is not empty as well.

pkg/apis/clusteroperator/validation/cluster_test.go
```diff
@@ -197,125 +209,153 @@ func TestValidateClusterSpec(t *testing.T) {
...
 					getTestMachineSet(2, "three", false, false),
-				},
-			},
+				}
+				return &cs
+			}(),
+			valid: false,
+		},
+		{
+			name: "missing cluster version",
```
staebler: Replace with tests for missing name and namespace individually.

pkg/apis/clusteroperator/validation/machineset.go
```diff
@@ -58,7 +58,11 @@ func validateMachineSetClusterOwner(ownerRefs []metav1.OwnerReference, fldPath *
 
 // validateMachineSetSpec validates the spec of a machine set
 func validateMachineSetSpec(spec *clusteroperator.MachineSetSpec, fldPath *field.Path) field.ErrorList {
-	return validateMachineSetConfig(&spec.MachineSetConfig, fldPath)
+	allErrs := validateMachineSetConfig(&spec.MachineSetConfig, fldPath)
+	if len(spec.Version.Name) == 0 {
```
staebler: Validate non-empty `Namespace`.

pkg/apis/clusteroperator/validation/machineset.go
```diff
@@ -58,7 +58,11 @@ func validateMachineSetClusterOwner(ownerRefs []metav1.OwnerReference, fldPath *
...
+		allErrs = append(allErrs, field.Required(fldPath.Child("version"), "must specify version"))
```
staebler: The field path should point to the name not the version.

pkg/apis/clusteroperator/validation/machineset_test.go
```diff
@@ -271,31 +285,44 @@ func TestValidateMachineSetSpec(t *testing.T) {
...
-				clusteroperator.MachineSetConfig{
+				MachineSetConfig: clusteroperator.MachineSetConfig{
 					NodeType: clusteroperator.NodeTypeMaster,
 					Size:     0,
 				},
+				Version: getClusterVersionReference(),
 			},
 		},
+		{
+			name: "missing version",
```
staebler: Replace with tests for missing name and namespace individually.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -519,6 +520,12 @@ func (c *ClusterController) manageMachineSet(cluster *clusteroperator.Cluster, m
 		return machineSet, true, err
 	}
 
+	if !apiequality.Semantic.DeepEqual(machineSet.Spec.Version, cluster.Spec.Version) {
+		glog.V(2).Infof("The cluster version of the machine set %s has changed from %v to %v", machineSet.Name, machineSet.Spec.Version, cluster.Spec.Version)
```
staebler: This comment comes off as a bit misleading to me. It states that the cluster version of the machine set changed, when in actuality the cluster version of the cluster changed.

openshift-docker:
Object references should always have namespace be optional, by convention.
It's a convenience for users, and it's part of our api conventions.

openshift-docker:
Do not deviate from API conventions unless you have a really good reason.
If you have additional API questions, please ping aos-devel or the master
team specifically.

staebler:
> Object references should always have namespace be optional, by convention. It's a convenience for users, and it's part of our api conventions.

@openshift-docker Fair enough. If that is the convention, then by all means let's go with that. I don't see any reason to buck convention here.

Is that convention implied from https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#object-references? Does "contain a subset of the fields of the `ObjectReference` type" imply since the `Namespace` field in `ObjectReference` is optional then the `Namespace` field in every object reference that includes namespace should be optional as well? Or is this an openshift convention that is codified somewhere else?

pkg/apis/clusteroperator/types.go
```diff
@@ -348,6 +359,9 @@ type MachineSetConfig struct {
 type MachineSetSpec struct {
 	// MachineSetConfig is the configuration for the MachineSet
 	MachineSetConfig
+
+	// Version references the clusterversion the machine set is running.
+	Version ClusterVersionReference
```
staebler: According to https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#object-references, the name of this field should be `ClusterVersionRef`.

staebler: As I think more about this, this needs to be an `ObjectReference` that points to an actual distinct existing `ClusterVersion` object. `ClusterStatus` should also have an `ObjectReference` that points to the same `ClusterVersion` object. The way it is now, if a `ClusterVersion` object is deleted and a new one is created with the same name, then a `Cluster` and its `MachineSets` can be out of sync with each other.

dgoodwin: Should we keep ClusterVersionReference and add UID, or revert to ObjectReference? I'm assuming we keep and add UID.

When you mention ClusterStatus having a ref to same ClusterVersion, did you mean ClusterSpec? Or do you suggest ClusterVersionReference on ClusterSpec (without UID set), require that to resolve to a UID and get set again on ClusterStatus, which then unblocks machine set creation. What does that layer in the middle buy vs just requiring ClusterSpec.ClusterVerRef set with UID?

staebler: My suggestion is this.
- `ClusterSpec` has a `ClusterVersionRef` field that is a `ClusterVersionReference`, which includes a namespace and a name. This is set by the user to indicate which `ClusterVersion` object the user wants.
- `ClusterStatus` has a `ClusterVersionRef` field that is an `ObjectReference`, which points to a specific existing `ClusterVersion` object with the name and namespace specified in the `ClusterVersionRef` from the `ClusterSpec`. This is set by the ClusterController when the `ClusterVersion` object matching the requested namespace and name is found. This is cleared by ClusterController when the `ClusterVersion` object is deleted.
- `MachineSetSpec` has a `ClsuterVersionRef` field that is an `ObjectReference`, which points to a specific existing `ClusterVersion` object. This is set by the ClusterController when the MachineSet is created.

> What does that layer in the middle buy vs just requiring ClusterSpec.ClusterVerRef set with UID?
We cannot require `ClusterSpec.ClusterVersionRef` to be set with a UID. The user *must* be able to create the Cluster object prior to the ClusterVersion object existing. That eventual reconciliation is a core tenet of kubernetes. And a controller should not modify the spec of objects that it does not create.

staebler: I am also fine with the `ClusterVersionRef` in `ClusterStatus` being just a `UID`, as that is sufficient to uniquely identify the ClusterVersion object. The ClusterVersion object can be found using the `ClusterVersionRef` in `ClusterSpec`.

dgoodwin: Ok thanks this makes more sense to me now.

staebler: I would also be fine with the `ClusterVersionRef` in `MachineSetSpec` being an object that only includes the name, namespace, and UID, since the other fields of `ObjectReference` are unnecessary for our purposes. I don't know what the conventional way to do this is. We may need to get some outside guidance.

dgoodwin: I'm a bit worried about the fact that this appears to imply that whenever a cluster version is created, we have to iterate every cluster looking to see if it tried to reference that version and update it's status if necessary. I expect a CI deployment would be a lot of version creations but few clusters, and a real world deployment a lot of clusters but few creations of versions, so maybe this is ok.

dgoodwin:
The UID problems resolves on the assumption that we allow you to delete a cluster version that's in use, should we actually allow that however? Doing so would put a live cluster into a pretty dangerous state, we may not be able to scale it up/down as we've lost details for new machines to apply. 

I'm wondering if deletions should be blocked if still referenced on any cluster version or machine set. (this would be somewhat consistent with our decision to make them immutable for now) If so, does this UID problem get simpler? The Cluster and MachineSet then could not really get out of sync and a name+namespace should be sufficient.

If we allow deletion of live cluster versions, and potentially for them to be recreated, we should remember that probably implies a full cluster upgrade. If that UID changes we have no guarantee on what was setup in the old/new versions, we have to upgrade as if it were an entirely new version.

Alternatively I could return to looking at copying the entire ClusterVersion onto each MachineSet rather than just a reference. This would help us establish what should be used and in theory you could then delete cluster versions and still scale existing machine sets up/down, just not create new.

pkg/apis/clusteroperator/types.go
staebler: Yes, it was always going to be the case where we would have to iterate over all of a type of resource when a cluster version was created. With the original implementation, we would have had to iterate over all of the machine sets when a cluster version was created. Otherwise there would be no way for the machine set controller to know when it was OK to sync a machine set if the cluster version were created after the machine set.

staebler:
> I'm wondering if deletions should be blocked if still referenced on any cluster version or machine set.

That is intriguing. I like that. We would need to put a finalizer on a cluster version that is only removed when there are no longer any clusters (and maybe machine sets) that reference the cluster version. When a user deletes a cluster version, it is retained in storage but can no longer be used by any new clusters.

We would still need some way to track whether a cluster is actively using a cluster version: We cannot block a user from creating a cluster that references the name of a deleted cluster version. We may as well track it via the UID of the cluster version. In which case, we may as well specify the UID when creating the machine set. So, even if we block removal of cluster versions when they are in use, it still makes sense to me to use the API that I suggested.

dgoodwin:
Decision from scrum: we'll plan to block deletion of cluster versions that are currently in use via an admission controller to cover most all cases, we may also implement a finalizer as a backup in the event of an extremely rare transactional issue where we attepmt to delete as someone is trying to create a cluster using it. CV deletion will be filed as a separate trello card.

For this PR I will syncCluster with exponential backoff until the cluster version can be successfully resolved. The status reference will only change if we see the cluster version name/namespace change on the clusterspec.

dgoodwin:
Ok I think this is ready for another look.

Missing versions will now cause the cluster to requeue until it exists. Status gets updated with an objectref + uid once resolved, this is also carried on the machineset. Doesn't block infra provisioning.

pkg/controller/cluster/cluster_controller_test.go
```diff
@@ -865,6 +1139,79 @@ func TestSyncClusterMachineSetsMutated(t *testing.T) {
 	}
 }
 
+//  TestSyncClusterVersionMutated tests that changing a cluster spec version results in new machine sets being created with the new version.
+func TestSyncClusterVersionMutated(t *testing.T) {
```
staebler: This is already covered by `TestSyncClusterWithVersion`.

pkg/logging/logging.go
```diff
@@ -0,0 +1,33 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
staebler: The year is now 2018.

pkg/logging/logging.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+
+	clusteroperator "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func WithMachineSet(logger log.FieldLogger, machineSet *clusteroperator.MachineSet) log.FieldLogger {
```
staebler: Add a comment for the function.

pkg/logging/logging.go
```diff
@@ -0,0 +1,33 @@
...
+	return logger.WithField("machineset", fmt.Sprintf("%s/%s", machineSet.Namespace, machineSet.Name))
+}
+
+func WithCluster(logger log.FieldLogger, cluster *clusteroperator.Cluster) log.FieldLogger {
```
staebler: Add a comment for the function.

contrib/examples/cluster.yaml
```diff
@@ -4,14 +4,47 @@ metadata:
...
+- apiVersion: clusteroperator.openshift.io/v1alpha1
+  kind: ClusterVersion
+  metadata:
+    name: ocp-v3-9
+    namespace: cluster-operator
+  spec:
+    imageFormat: "openshift/origin-${component}:v3.9.0-alpha.1"
+    vmImages:
+      awsVMImages:
+        amiByRegion:
```
staebler: This is outside the scope of this PR, but `amiByRegion` should not be a map. See https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#lists-of-named-subobjects-preferred-over-maps.

dgoodwin: Oh good info, ok adding an issue for that one then.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -345,29 +371,64 @@ func (c *ClusterController) syncCluster(key string) error {
...
 	clusterKey, err := controller.KeyFunc(cluster)
 	if err != nil {
 		utilruntime.HandleError(fmt.Errorf("Couldn't get key for Cluster %#v: %v", cluster, err))
 		return nil
 	}
+	clusterLog.Debugf("managing machine sets")
+
+	// Safeguard against the cluster status version ref somehow getting cleared, but machine sets exist.
+	// This should not happen but if so, make sure we don't go wipe out all existing machine sets.
+	if clusterVersion == nil && len(machineSets) > 0 {
```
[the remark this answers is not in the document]
dgoodwin: The situation would be quite alarming so I wanted to make sure that if it ever happened, it logged a warning. Arguably just being called with nil indicates programmer error but it felt like seeing there were also active machine sets could be relevant. To be honest I'd kind of like to see a warning in both cases. I will clarify the logging a little with separate statements.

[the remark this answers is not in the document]
staebler: I wasn't suggesting to get rid of the warning. I was suggesting that you loosen the constraints on when to log the warning. Log **whenever** clusterVersion is nil regardless of what machineSets is.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -454,9 +515,11 @@ func (c *ClusterController) manageMachineSets(machineSets []*clusteroperator.Mac
 				}
 				if err != nil {
 					// Decrement the expected number of creates because the informer won't observe this machine set
-					glog.V(2).Infof("Failed creation, decrementing expectations for cluster %q/%q", cluster.Namespace, cluster.Name)
+					clusterLog.Errorf("failed creation, decrementing expectations for cluster: %v", err)
```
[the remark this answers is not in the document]
dgoodwin: This struck me when I saw it as well, it seems like it should be. I will modify.

pkg/logging/logging.go
dgoodwin: Not in Canada, typical American assumptions.

(will fix)

dgoodwin:
Updated from review, conflicts resolved but I resorted to merge rather than rebase this time around.

staebler:
This has been moved to a separate PR (#77) to be rebased.

PR #48, opened 2018-01-19
Machineset controller

csrwng:
Updates machineset controller to run the appropriate provisioning playbooks

### DIFF
--- a/pkg/ansible/generate.go
+++ b/pkg/ansible/generate.go
@@ -11,7 +11,8 @@ import (
 // varsTemplate contains some hardcoded variables that are constant for all clusters,
 // as well as the go template annotations to substitute in values we do need to change.
 // Today this is AWS specific but will need to be broken down for multi-provider.
-const varsTemplate = `---
+const (
+	clusterVarsTemplate = `---
 # Variables that are commented in this file are optional; uncommented variables
 # are mandatory.
 
@@ -30,7 +31,7 @@ ansible_ssh_user: centos
 # Ensure these variables are set for bootstrap
 ################################################################################
 # Deployment type must be specified. ('origin' or 'openshift-enterprise')
-openshift_deployment_type: openshift-enterprise
+openshift_deployment_type: origin
 
 openshift_master_bootstrap_enabled: True
 
@@ -46,9 +47,11 @@ openshift_master_console_port: 443
 
 # openshift_release must be specified.  Use whatever version of openshift
 # that is supported by openshift-ansible that you wish.
+# TODO: Parameterize
 openshift_release: "v3.7" # v3.7
 
 # This will be dependent on the version provided by the yum repository
+# TODO: Parameterize
 openshift_pkg_version: -3.7.0 # -3.7.0
 
 # TODO: development specific
@@ -58,6 +61,7 @@ openshift_repos_enable_testing: true
 # AWS region
 # This value will instruct the plays where all items should be created.
 # Multi-region deployments are not supported using these plays at this time.
+openshift_aws_region: [[ .Region ]]
 
 
 #openshift_aws_create_launch_config: true
@@ -90,19 +94,19 @@ openshift_aws_create_security_groups: true
 # openshift_aws_launch_config_security_groups specifies the security groups to
 # apply to the launch config.  The launch config security groups will be what
 # the cluster actually is deployed in.
-openshift_aws_launch_config_security_groups:
-  compute:
-  - cluster-engine
-  - cluster-engine_compute
-  - cluster-engine_compute_k8s
-  infra:
-  - cluster-engine
-  - cluster-engine_infra
-  - cluster-engine_infra_k8s
-  master:
-  - cluster-engine
-  - cluster-engine_master
-  - cluster-engine_master_k8s
+# openshift_aws_launch_config_security_groups:
+#   compute:
+#   - cluster-engine
+#   - cluster-engine_compute
+#   - cluster-engine_compute_k8s
+#   infra:
+#   - cluster-engine
+#   - cluster-engine_infra
+#   - cluster-engine_infra_k8s
+#   master:
+#   - cluster-engine
+#   - cluster-engine_master
+#   - cluster-engine_master_k8s
 
 # openshift_aws_node_security_groups are created when
 # openshift_aws_create_security_groups is set to true.
@@ -115,7 +119,7 @@ openshift_aws_launch_config_security_groups:
 # Specify the key pair name here to connect to the provisioned instances.  This
 # can be an existing key, or it can be one of the keys specified in
 # openshift_aws_users
-openshift_aws_ssh_key_name: libra # myuser_key
+openshift_aws_ssh_key_name: [[ .SSHKeyName ]]
 
 # This will ensure these user and public keys are created.
 #openshift_aws_users:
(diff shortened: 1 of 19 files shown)


pkg/apis/clusteroperator/types.go
```diff
@@ -378,6 +385,19 @@ type MachineSetStatus struct {
 
 	// Conditions includes more detailed status of the MachineSet
 	Conditions []MachineSetCondition
+
+	// Installed is true if the software required for this machine set is installed
+	// and running.
+	Installed bool
```
csrwng: @staebler I added this boolean to indicate that the install playbook had been run successfully on the masters. It needs to be reconciled with whatever you're doing for the master controller.

staebler: ack

pkg/ansible/generate.go
```diff
@@ -140,79 +144,243 @@ openshift_aws_create_s3: false
...
+	}
+
+	var buf bytes.Buffer
+	buf.WriteString(commonVars)
+
+	var varsTemplate string
+	switch {
+	case machineSet.Spec.NodeType == coapi.NodeTypeMaster:
+		varsTemplate = masterVarsTemplate
+	case machineSet.Spec.Infra:
```
staebler: How do we handle the case where the machineset is both master and infra? Overall, I am a little confused about how each of the vars templates work. I would have expected (1) that there some base vars that each machineset type would include, (2) that some vars would be added for master machinesets, (3) that some vars would be added for non-master machinesets, and (4) that some vars would be added for infra machinesets.

pkg/ansible/generate.go
```diff
@@ -140,79 +144,243 @@ openshift_aws_create_s3: false
...
+
+openshift_aws_node_groups:
+- name: "{{ openshift_aws_clusterid }} infra group"
+  group: infra
+  tags:
+    host-type: node
+    sub-host-type: infra
+    runtime: docker
+
+openshift_aws_ami_map:
```
staebler: This is duplicated from the beginning of `infraVarsTemplate`.

pkg/ansible/generate.go
```diff
@@ -140,79 +144,243 @@ openshift_aws_create_s3: false
...
+		varsTemplate = computeVarsTemplate
+	}
+
+	t, err := template.New("machinesetvars").Delims("[[", "]]").Parse(varsTemplate)
+	if err != nil {
+		return "", err
+	}
+
+	params := &machineSetParams{
+		Name:         machineSet.Name,
```
staebler: Do we want this to be the full name of the machine set, with the cluster name and the generated suffix? Or do we want this to be the simpler name from `ClusterMachineSet`? Or does it not matter because it is only used and seen by the ansible playbook?

pkg/controller/machineset/machineset_controller.go
```diff
@@ -207,48 +395,331 @@ func (c *MachineSetController) handleErr(err error, key interface{}) {
...
+	if cluster.UID != controllerRef.UID {
+		// The controller we found with this Name is not the same one that the
+		// ControllerRef points to.
+		return nil, nil
+	}
+	return cluster, nil
+}
+
+func (c *MachineSetController) shouldProvision(machineSet *clusteroperator.MachineSet) bool {
+	if machineSet.Status.ProvisionedJobGeneration == machineSet.Generation {
```
staebler: Do we also need to add a check for whether the cluster spec has changed? Or do you expect that any change to the cluster spec that would change the deployment of the machine set would be accomplished by a separate controller (ClusterController?) deleting and re-creating the machineset object or making some other modification to the spec of the machineset?

staebler:
I don't know enough about the ansible playbooks being run to give a worthwhile review of the ansible vars being generated. It would be good for someone with more knowledge about that (@abutcher, @joelddiaz, ???) to give a look over that part.

pkg/ansible/generate.go
csrwng: It is currently not possible with the playbooks we have to have the master be infra. Playbooks are designed to work with 3 distinct sets of nodes (based on my conversation with @abutcher)

staebler: Do we expect that to change in the future? Or do we need to adjust the API types to reflect that reality?

pkg/ansible/generate.go
csrwng: It does go beyond just the playbook because it goes on the labels on the AWS asg. I did think about using the shorter name from the ClusterMachineSet. However, it is conceivable that at some point (during an upgrade or whatever), you could end up with more than one ASG with the same name/label, which would be confusing.

pkg/controller/machineset/machineset_controller.go
csrwng: I would think (and prefer) the latter

pkg/ansible/generate.go
csrwng: I needed to duplicate some structs from the role defaults so that I could have a separate set of structs per machineset. As it is now, you're expected to create all non-master asg's at the same time.

pkg/ansible/generate.go
csrwng: thx, will remove

pkg/ansible/generate.go
csrwng: @staebler we need to have the conversation and make a call. It comes down to whether we care enough about the use case to justify the work required  to make it work.

pkg/ansible/generate.go
```diff
@@ -140,79 +144,240 @@ openshift_aws_create_s3: false
...
+`
+
+	computeVarsTemplate = `
+openshift_aws_ami_map:
+  [[ .Name ]]: [[ .AMIName ]]
+
+openshift_aws_node_groups:
+- name: "{{ openshift_aws_clusterid }} [[ .Name ]] group"
+  group: [[ .Name ]]
+  tags:
```
abutcher: Should we add a "Name" tag to make our lives easier in the dev account? Same would go with the infra group and `openshift_aws_master_group` if it were added.

```
openshift_aws_master_group:                                                                                                                                                                                        
- name: "{{ openshift_aws_clusterid }} master group"                                                                                                                                                               
  group: master                                                                                                                                                                                                    
  tags:                                                                                                                                                                                                            
    host-type: master                                                                                                                                                                                              
    sub-host-type: default                                                                                                                                                                                         
    runtime: docker                                                                                                                                                                                                
    Name: "<<< something >>>"
```

pkg/ansible/generate.go
```diff
@@ -140,79 +144,240 @@ openshift_aws_create_s3: false
...
+    max_size: [[ .Size ]]
+    desired_size: [[ .Size ]]
+    termination_policy: "{{ openshift_aws_node_group_termination_policy }}"
+    replace_all_instances: "{{ openshift_aws_node_group_replace_all_instances }}"
+    iam_role: "{{ openshift_aws_iam_role_name }}"
+    policy_name: "{{ openshift_aws_iam_role_policy_name }}"
+    policy_json: "{{ openshift_aws_iam_role_policy_json }}"
+    elbs: "{{ openshift_aws_elb_name_dict['infra'].keys()| map('extract', openshift_aws_elb_name_dict['infra']) | list }}"
+
+openshift_aws_node_groups:
```
abutcher: `openshift_aws_node_groups` is defined twice. I think both should be combined into a list containing infra and the group we're creating, right?

pkg/ansible/generate.go
csrwng: yup, makes sense

pkg/ansible/generate.go
abutcher: Or maybe just this one should be removed? The latter is taking precedence is what I mean :)

csrwng: It's defined in different templates. One's used for the infra machineset and the other for the compute machineset. It doesn't hurt to have both, but I thought it kept things more explicit if I only used the one I needed for the machineset I'm creating.

abutcher: Gotcha :+1:

pkg/ansible/generate.go
dgoodwin: I'm thinking we should do what we can to at least keep that door open. Spinning up single master all-in-one clusters  could we very desirable for kick the tires scenarios.

staebler:
This is ready for review now.

Note that only the master machineset will be provisioned. The compute machinesets are blocked from being provisioned on the master being installed. The master will not be installed until the MasterController is implemented.

Once this is merged in, extra caution needs to be taken when using AWS. The ELBs and ASGs will be provisioned for the master, but there is nothing yet to automatically deprovision them. It is the responsibility of the user to manually clean up the resource in AWS.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -207,48 +296,353 @@ func (c *MachineSetController) handleErr(err error, key interface{}) {
...
+	}
+	if !cluster.Status.Provisioned || cluster.Status.ProvisionedJobGeneration != cluster.Generation {
+		return false
+	}
+	switch machineSet.Spec.NodeType {
+	case clusteroperator.NodeTypeMaster:
+		return true
+	case clusteroperator.NodeTypeCompute:
+		masterMachineSet, err := c.machineSetsLister.MachineSets(machineSet.Namespace).Get(cluster.Status.MasterMachineSetName)
+		if err != nil {
```
dgoodwin: I'd like to see this logged as well.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -207,48 +296,353 @@ func (c *MachineSetController) handleErr(err error, key interface{}) {
...
+			return false
+		}
+		return masterMachineSet.Status.Installed && masterMachineSet.Status.InstalledJobGeneration == masterMachineSet.Generation
```
dgoodwin: Should the == be a >=? Or does an increase in generation imply re-installation and thus ==? (could use a comment either way)

pkg/controller/machineset/machineset_controller.go
```diff
@@ -207,48 +296,353 @@ func (c *MachineSetController) handleErr(err error, key interface{}) {
...
+	now := metav1.Now()
+
+	provisioning := machineSetCondition(machineSet, clusteroperator.MachineSetHardwareProvisioning)
+	if provisioning != nil &&
+		provisioning.Status == kapi.ConditionTrue {
+		provisioning.Status = kapi.ConditionFalse
+		provisioning.LastTransitionTime = now
+		provisioning.LastProbeTime = now
+		provisioning.Reason = "SpecChanged"
+		provisioning.Message = "Spec changed. New provisioning needed"
```
dgoodwin: Method could use a godoc, I'm a bit confused about the spec changed reason/message.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -207,48 +296,353 @@ func (c *MachineSetController) handleErr(err error, key interface{}) {
...
+	}
+
+	return c.updateMachineSetStatus(original, machineSet)
+}
+
+func (c *MachineSetController) syncMachineSetStatusWithJob(original *clusteroperator.MachineSet, job *v1batch.Job) error {
```
dgoodwin: Pretty important function, godoc might be good here as well.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -207,48 +296,353 @@ func (c *MachineSetController) handleErr(err error, key interface{}) {
...
+	jobFailed := jobCondition(job, v1batch.JobFailed)
+	switch {
+	case jobCompleted != nil && jobCompleted.Status == kapi.ConditionTrue:
+		provisioning := machineSetCondition(machineSet, clusteroperator.MachineSetHardwareProvisioning)
+		if provisioning != nil &&
+			provisioning.Status == kapi.ConditionTrue {
+			provisioning.Status = kapi.ConditionFalse
+			provisioning.LastTransitionTime = now
+			provisioning.LastProbeTime = now
+			provisioning.Reason = "JobCompleted"
```
dgoodwin: Should we do consts for all these Reasons?

pkg/controller/machineset/machineset_controller.go
```diff
@@ -207,48 +296,353 @@ func (c *MachineSetController) handleErr(err error, key interface{}) {
...
+}
+
+func (c *MachineSetController) syncMachineSetStatusWithJob(original *clusteroperator.MachineSet, job *v1batch.Job) error {
+	machineSet := original.DeepCopy()
+	now := metav1.Now()
+
+	jobCompleted := jobCondition(job, v1batch.JobComplete)
+	jobFailed := jobCondition(job, v1batch.JobFailed)
+	switch {
+	case jobCompleted != nil && jobCompleted.Status == kapi.ConditionTrue:
```
dgoodwin: Comments on the major cases/logic blocks would be nice for readability.

pkg/controller/client_utils.go
```diff
@@ -46,3 +46,37 @@ func preparePatchBytesforClusterStatus(oldCluster, newCluster *clusteroperator.C
...
 	return patchBytes, nil
 }
+
+func PatchMachineSetStatus(c clusteroperatorclientset.Interface, oldMachineSet, newMachineSet *clusteroperator.MachineSet) error {
+	logger := log.WithField("machineset", fmt.Sprintf("%s/%s", oldMachineSet.Namespace, oldMachineSet.Name))
+	patchBytes, err := preparePatchBytesForMachineSetStatus(oldMachineSet, newMachineSet)
+	if err != nil {
+		return err
+	}
+	logger.Debugf("about to patch machineset with %s", string(patchBytes))
```
dgoodwin: In manual testing this is showing up with some blank patch bytes:

```
$ oc logs cluster-operator-controller-manager-86f754b7d9-msbhq | grep "about to patch"
time="2018-01-29T14:37:12Z" level=debug msg="about to patch cluster with {\"status\":{\"conditions\":[{\"lastProbeTime\":\"2018-01-29T14:37:12Z\",\"lastTransitionTime\":\"2018-01-29T14:37:12Z\",\"message\":\"Job cluster-operator/job-infra-dgoodwin-cluster-t9fnm is running since \\u003cnil\\u003e. Pod completions: 0, failures: 0\",\"reason\":\"JobRunning\",\"status\":\"True\",\"type\":\"InfraProvisioning\"}]}}" cluster=cluster-operator/dgoodwin-cluster
time="2018-01-29T14:37:12Z" level=debug msg="about to patch cluster with {}" cluster=cluster-operator/dgoodwin-cluster
time="2018-01-29T14:37:13Z" level=debug msg="about to patch cluster with {}" cluster=cluster-operator/dgoodwin-cluster
time="2018-01-29T14:37:13Z" level=debug msg="about to patch cluster with {}" cluster=cluster-operator/dgoodwin-cluster
time="2018-01-29T14:37:13Z" level=debug msg="about to patch cluster with {}" cluster=cluster-operator/dgoodwin-cluster
time="2018-01-29T14:38:41Z" level=debug msg="about to patch cluster with {\"status\":{\"conditions\":[{\"lastProbeTime\":\"2018-01-29T14:38:41Z\",\"lastTransitionTime\":\"2018-01-29T14:38:41Z\",\"message\":\"Job cluster-operator/job-infra-dgoodwin-cluster-t9fnm completed at 2018-01-29 14:38:41 +0000 UTC\",\"reason\":\"JobCompleted\",\"status\":\"False\",\"type\":\"InfraProvisioning\"},{\"lastProbeTime\":\"2018-01-29T14:38:41Z\",\"lastTransitionTime\":\"2018-01-29T14:38:41Z\",\"message\":\"Job cluster-operator/job-infra-dgoodwin-cluster-t9fnm completed at 2018-01-29 14:38:41 +0000 UTC\",\"reason\":\"JobCompleted\",\"status\":\"True\",\"type\":\"InfraProvisioned\"}],\"provisioned\":true,\"provisionedJobGeneration\":1}}" cluster=cluster-operator/dgoodwin-cluster
time="2018-01-29T14:38:42Z" level=debug msg="about to patch machineset with {\"status\":{\"conditions\":[{\"lastProbeTime\":\"2018-01-29T14:38:42Z\",\"lastTransitionTime\":\"2018-01-29T14:38:42Z\",\"message\":\"Job cluster-operator/provision-machineset-dgoodwin-cluster-master-f7dlr-kj9wm is running since \\u003cnil\\u003e. Pod completions: 0, failures: 0\",\"reason\":\"JobRunning\",\"status\":\"True\",\"type\":\"HardwareProvisioning\"}]}}" machineset=cluster-operator/dgoodwin-cluster-master-f7dlr
time="2018-01-29T14:38:42Z" level=debug msg="about to patch machineset with {}" machineset=cluster-operator/dgoodwin-cluster-master-f7dlr
time="2018-01-29T14:38:42Z" level=debug msg="about to patch machineset with {}" machineset=cluster-operator/dgoodwin-cluster-master-f7dlr
```

staebler: I would like to hold off on addressing this until a later PR, to keep this PR from growing with additional stuff.

dgoodwin: Works for me.

pkg/controller/machineset/machineset_controller.go
staebler: It wouldn't make sense for the InstalledJobGeneration to be greater than the Generation. That would mean that the master controller set a InstalledJobGeneration to a Generation that has not yet existed.

I've added a comment.

dgoodwin: Sorry I should have put <=, I was wondering if the machine set gen could be further ahead than installed gen, but it looks like we can probably assume this will trigger when they're equal and after that we don't need to do anything.

staebler: This check is there *specifically* to prevent provisioning the compute nodes in the interregnum between when the master nodes were installed with the old spec and when they are installed with the new spec. The `masterMachineSet.Status.Installed` flag is invalidated by the spec change. But the machine set controller may try to sync the compute node before the master controller gets around to syncing the updated master machine set to clear the flag.

PR #50, opened 2018-01-24
Fix infra controller crashes when jobs exist for since deleted clusters.

dgoodwin:


dgoodwin:
Trying to figure out what to do for unit testing but here's the fix @staebler

PR #51, opened 2018-01-25
Move job event handling to jobcontrol from individual controllers

staebler:
This will remove boilerplate code that needs to be copied to all of the individual controllers that are using a jobcontrol.

### DIFF
--- a/pkg/controller/jobcontrol.go
+++ b/pkg/controller/jobcontrol.go
@@ -29,15 +29,24 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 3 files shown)


pkg/controller/jobcontrol.go
```diff
@@ -284,3 +424,16 @@ func jobOwnerGeneration(job *kbatch.Job) int64 {
...
 	return fmt.Sprintf("%s/%s", job.Namespace, job.Name)
 }
+
+func loggerForJob(logger log.FieldLogger, job *kbatch.Job) log.FieldLogger {
+	return logger.WithFields(log.Fields{"job": job.Name, "namespace": job.Namespace})
+}
+
+func loggerForOwner(logger log.FieldLogger, owner metav1.Object) log.FieldLogger {
+	return logger.WithFields(log.Fields{"owner": owner.GetName(), "namespace": owner.GetNamespace()})
+}
```
dgoodwin: 👍

dgoodwin:
LGTM, @joelddiaz can you take a pass and do the honors if everything looks ok to you as well.

Like the use of logging across code and tests both.

PR #57, opened 2018-01-26
cluster infra delete/deprovision

joelddiaz:
Add some handling in syncCluster() for when a cluster object has been marked for deletion

Fire off a non-JobControl(ed) job to run the 'uninstall_prerequisites.yml' playbook to remove the infra pieces

Update from the origin-ansible:v3.8 to 'latest' since that's where the uninstall/deprovision playbooks exist.

### DIFF
--- a/pkg/controller/infra/infra_controller.go
+++ b/pkg/controller/infra/infra_controller.go
@@ -24,6 +24,7 @@ import (
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	batchinformers "k8s.io/client-go/informers/batch/v1"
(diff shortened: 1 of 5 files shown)


pkg/registry/clusteroperator/cluster/strategy.go
```diff
@@ -91,6 +91,7 @@ func (clusterRESTStrategy) PrepareForCreate(ctx genericapirequest.Context, obj r
 	// Creating a brand new object, thus it must have no
 	// status. We can't fail here if they passed a status in, so
 	// we just wipe it clean.
+	cluster.Finalizers = []string{clusteroperator.FinalizerClusterOperator}
```
staebler: This is causing problems for the TestCreate function in pkg/registry/clusteroperator/cluster/storage_test.go. Thinking about this more, I think it is bad design to add the finalizer in the registry strategy. That requires that there is a controller that will come along and take responsibility for the finalizer. I think it is better to have that controller add the finalizer itself.

I recommend we move adding the finalizer to [here](https://github.com/joelddiaz/cluster-operator/blob/16c22dc5813813272759b0cbeda577f219342825/pkg/controller/infra/infra_controller.go#L527). Add the finalizer if it is not already there.

pkg/controller/infra/infra_controller.go
```diff
@@ -615,14 +634,40 @@ func (f jobFactory) BuildJob(name string) (*v1batch.Job, *kapi.ConfigMap, error)
 	return f(name)
 }
 
-func (c *InfraController) getJobFactory(cluster *clusteroperator.Cluster) controller.JobFactory {
+func (c *InfraController) getJobFactory(cluster *clusteroperator.Cluster, playbook string) controller.JobFactory {
+	c.logger.Infof("PLAYBOOK: %s", playbook)
```
staebler: Let's get rid of this info log. The ansible generator is already going to be writing an info log with the name of the playbook when the playbook job is generated.

Issue #46, opened 2018-01-17
Need to delete configmaps when no longer needed by job


staebler:
We are leaking configmaps currently because there is no code to delete them. A naive way to delete a configmap is to delete the configmap with the same name as the job being deleted.

staebler:
Resolved by #66.

PR #66, opened 2018-01-31
Delete configmaps after deleting the associated jobs

staebler:
This is not resilient to errors that happen when deleting the configmap. If the delete fails, it will not be retried (unless deleting the job failed as well).


staebler:
Fixes #46.

PR #68, opened 2018-01-31
Clean up etcd stores on API Server shutdown

staebler:
The integration tests will start and stop API Servers. This will keep those tests from leaking go routines.


csrwng:
lgtm

PR #70, opened 2018-02-01
JobSync tests

staebler:
This adds unit tests for jobSync, which is pending in #63.

Mocks are generated by gomock. I am struggling to figure out how to get the mocks generated in the makefile without causing the build to do `go build` again every time due to the generated files being touched. I will work on that as a follow-up.


staebler:
Created issue #72 to generate the mocks as part of the build/test process.

PR #75, opened 2018-02-02
Tweaks to make ec2 provisioning work

csrwng:
When we get more parameterization on the openshift-ansible side we'll be able to get rid of the complex structs that we're passing in vars. But for now, these changes make provisioning work, up to the point where we have a master and nodes running. The last remaining step is to approve the nodes.

### DIFF
--- a/pkg/controller/machineset/machineset_controller.go
+++ b/pkg/controller/machineset/machineset_controller.go
@@ -177,6 +177,25 @@ func (c *Controller) updateMachineSet(old, cur interface{}) {
 	ms := cur.(*clusteroperator.MachineSet)
 	loggerForMachineSet(c.logger, ms).Debugf("enqueueing updated machine set")
 	c.enqueueMachineSet(ms)
+	if ms.Spec.NodeType == clusteroperator.NodeTypeMaster && ms.Status.Installed {
+		cluster, err := controller.ClusterForMachineSet(ms, c.clustersLister)
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("cannot retrieve cluster for machineset %s/%s: %v", ms.Namespace, ms.Name, err))
+			return
+		}
+		machineSets, err := c.machineSetsForCluster(cluster)
+		if err != nil {
+			c.logger.Errorf("cannot retrieve machine sets for cluster: %v", err)
+			utilruntime.HandleError(err)
+			return
+		}
+		for _, machineSet := range machineSets {
+			if machineSet.Name != ms.Name {
+				loggerForMachineSet(c.logger, machineSet).Debugf("enqueueing machine set for installed master")
+				c.enqueueMachineSet(machineSet)
+			}
+		}
+	}
 }
 
 func (c *Controller) deleteMachineSet(obj interface{}) {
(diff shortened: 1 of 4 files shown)


csrwng:
@staebler @dgoodwin ptal

pkg/controller/master/master_controller.go
```diff
@@ -341,7 +341,7 @@ func (s *jobSyncStrategy) GetJobFactory(owner metav1.Object) (controller.JobFact
 		return nil, err
 	}
 	return jobFactory(func(name string) (*v1batch.Job, *kapi.ConfigMap, error) {
-		vars, err := ansible.GenerateMachineSetVars(cluster, machineSet)
+		vars, err := ansible.GenerateClusterVars(cluster)
```
staebler: Does using the machine set vars cause undesired behavior? Or does it simply include vars that are not needed?

csrwng: It causes undesired behavior, mainly because we're including whole structs in the vars.
For master install, the openshift_aws role is not used and some placeholders don't get resolved with actual values.

csrwng:
Added a change to enqueue other machinesets in the cluster when the master has been installed

pkg/controller/master/master_controller.go
staebler: OK. I think long term I would like something to be more resilient, whether it is our code or the playbooks. It doesn't make sense to me that when we are running a playbook against a machine set that we cannot use the machine set vars.

With that being said, I am fine with leaving it as you have it for now.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -175,8 +175,28 @@ func (c *MachineSetController) addMachineSet(obj interface{}) {
 
 func (c *MachineSetController) updateMachineSet(old, cur interface{}) {
 	ms := cur.(*clusteroperator.MachineSet)
+	oldms := cur.(*clusteroperator.MachineSet)
 	loggerForMachineSet(c.logger, ms).Debugf("enqueueing updated machine set")
 	c.enqueueMachineSet(ms)
+	if !oldms.Status.Installed && ms.Status.Installed {
```
staebler: Should this be done only for when the machine set is a master? Or maybe a master or an infra?

staebler: Is it fragile to only enqueue the other machinesets on the transition from not installed to installed? I think that I would rather just always enqueue all of the other machinesets when a master machineset is updated.

csrwng: I will switch it to check for master

csrwng:
@staebler updated to only check for master machineset, but thinking more about it, shouldn't we at least check that the machine set is installed?

staebler:
@csrwng That's reasonable. Master and installed.

pkg/controller/master/master_controller.go
csrwng: We should be able to fix things once we get additional parameters in the aws role default vars. (At least for infra and master). For compute, we'll need to figure out what's the best way to structure the role vars so we don't have to include chunks of the config.

csrwng:
Updated

PR #77, opened 2018-02-05
Immutable ClusterVersions, and references on cluster objects

staebler:
Add ClusterVersion reference to ClusterSpec.

Cluster version here should be mutable, changing it will eventually
trigger a cluster upgrade.

Add ClusterVersionReference to MachineSets and recreate on changes.

Resolve cluster versions before creating machine sets.

If a cluster version is requested that does not yet exist, we return and
re-try syncing the cluster with backoff until it does.

Once it exists we allow machine set processing to proceed. If the
version is being updated to one that does not exist, we just leave the
machine sets as they are.

Once resolved the cluster status is stamped with the resolve version,
including it's UID.


staebler:
This is a copy of #47 re-based to the latest master.

Closes #47.

PR #79, opened 2018-02-05
Controller to accept nodes in cluster master

csrwng:
Adds a new controller to run to the accept playbook for a cluster where all machinesets have been provisioned and the master machineset is installed.

### DIFF
--- a/cmd/cluster-operator-controller-manager/app/controllermanager.go
+++ b/cmd/cluster-operator-controller-manager/app/controllermanager.go
@@ -62,6 +62,7 @@ import (
 	"github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
 	clusteroperatorinformers "github.com/openshift/cluster-operator/pkg/client/informers_generated/externalversions"
 	"github.com/openshift/cluster-operator/pkg/controller"
+	"github.com/openshift/cluster-operator/pkg/controller/accept"
 	"github.com/openshift/cluster-operator/pkg/controller/cluster"
 	"github.com/openshift/cluster-operator/pkg/controller/infra"
 	"github.com/openshift/cluster-operator/pkg/controller/machine"
@@ -321,6 +322,7 @@ func NewControllerInitializers() map[string]InitFunc {
 	controllers["machineset"] = startMachineSetController
 	controllers["machine"] = startMachineController
 	controllers["master"] = startMasterController
+	controllers["accept"] = startAcceptController
 	return controllers
 }
 
@@ -444,7 +446,7 @@ func StartControllers(ctx ControllerContext, controllers map[string]InitFunc) er
 }
 
 func startClusterController(ctx ControllerContext) (bool, error) {
-	if !ctx.AvailableResources[schema.GroupVersionResource{Group: "clusteroperator.openshift.io", Version: "v1alpha1", Resource: "clusters"}] {
+	if !resourcesAvailable(ctx) {
 		return false, nil
 	}
 	go cluster.NewController(
@@ -458,7 +460,7 @@ func startClusterController(ctx ControllerContext) (bool, error) {
 }
 
 func startInfraController(ctx ControllerContext) (bool, error) {
-	if !ctx.AvailableResources[schema.GroupVersionResource{Group: "clusteroperator.openshift.io", Version: "v1alpha1", Resource: "clusters"}] {
+	if !resourcesAvailable(ctx) {
 		return false, nil
 	}
 	go infra.NewController(
@@ -473,7 +475,7 @@ func startInfraController(ctx ControllerContext) (bool, error) {
 }
 
 func startMachineSetController(ctx ControllerContext) (bool, error) {
-	if !ctx.AvailableResources[schema.GroupVersionResource{Group: "clusteroperator.openshift.io", Version: "v1alpha1", Resource: "machinesets"}] {
+	if !resourcesAvailable(ctx) {
 		return false, nil
 	}
 	go machineset.NewController(
@@ -489,7 +491,7 @@ func startMachineSetController(ctx ControllerContext) (bool, error) {
 }
 
 func startMachineController(ctx ControllerContext) (bool, error) {
-	if !ctx.AvailableResources[schema.GroupVersionResource{Group: "clusteroperator.openshift.io", Version: "v1alpha1", Resource: "machines"}] {
+	if !resourcesAvailable(ctx) {
 		return false, nil
 	}
 	go machine.NewController(
@@ -501,7 +503,7 @@ func startMachineController(ctx ControllerContext) (bool, error) {
 }
 
 func startMasterController(ctx ControllerContext) (bool, error) {
-	if !ctx.AvailableResources[schema.GroupVersionResource{Group: "clusteroperator.openshift.io", Version: "v1alpha1", Resource: "machinesets"}] {
+	if !resourcesAvailable(ctx) {
 		return false, nil
 	}
 	go master.NewController(
@@ -514,4 +516,35 @@ func startMasterController(ctx ControllerContext) (bool, error) {
 		v1.PullPolicy(ctx.Options.AnsibleImagePullPolicy),
 	).Run(int(ctx.Options.ConcurrentMasterSyncs), ctx.Stop)
 	return true, nil
+}
+
+func startAcceptController(ctx ControllerContext) (bool, error) {
+	if !resourcesAvailable(ctx) {
+		return false, nil
+	}
+	go accept.NewController(
+		ctx.InformerFactory.Clusteroperator().V1alpha1().Clusters(),
+		ctx.InformerFactory.Clusteroperator().V1alpha1().MachineSets(),
+		ctx.KubeInformerFactory.Batch().V1().Jobs(),
+		ctx.ClientBuilder.KubeClientOrDie("clusteroperator-accept-controller"),
+		ctx.ClientBuilder.ClientOrDie("clusteroperator-accept-controller"),
+		ctx.Options.AnsibleImage,
+		v1.PullPolicy(ctx.Options.AnsibleImagePullPolicy),
+	).Run(int(ctx.Options.ConcurrentAcceptSyncs), ctx.Stop)
+	return true, nil
+}
+
+func resourcesAvailable(ctx ControllerContext) bool {
+	gvrs := []schema.GroupVersionResource{
+		{Group: "clusteroperator.openshift.io", Version: "v1alpha1", Resource: "clusters"},
+		{Group: "clusteroperator.openshift.io", Version: "v1alpha1", Resource: "machinesets"},
+	}
+
+	for _, gvr := range gvrs {
+		if !ctx.AvailableResources[gvr] {
+			return false
+		}
+	}
+
+	return true
 }
--- a/cmd/cluster-operator-controller-manager/app/options/options.go
+++ b/cmd/cluster-operator-controller-manager/app/options/options.go
@@ -72,6 +72,7 @@ func NewCMServer() *CMServer {
 			ConcurrentMachineSetSyncs:     defaultConcurrentSyncs,
 			ConcurrentMachineSyncs:        defaultConcurrentSyncs,
 			ConcurrentMasterSyncs:         defaultConcurrentSyncs,
+			ConcurrentAcceptSyncs:         defaultConcurrentSyncs,
 			LeaderElection:                leaderelectionconfig.DefaultLeaderElectionConfiguration(),
 			LeaderElectionNamespace:       defaultLeaderElectionNamespace,
 			ControllerStartInterval:       metav1.Duration{Duration: 0 * time.Second},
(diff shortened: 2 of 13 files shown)


csrwng:
@staebler ptal

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+
+	// Machines that need to be synced
+	queue workqueue.RateLimitingInterface
+
+	logger log.FieldLogger
+}
+
+func (c *Controller) addMachineSet(obj interface{}) {
+	machineSet := obj.(*clusteroperator.MachineSet)
+	c.logger.Debugf("Enqueuing cluster for added machineset %s/%s", machineSet.Namespace, machineSet.Name)
```
dgoodwin: Checkout the logging pkg WithMachineSet and WithCluster functions, will keep all these standardized.

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+			vars,
+		)
+		return job, configMap, nil
+	}), nil
+}
+
+func (s *jobSyncStrategy) GetOwnerCurrentJob(owner metav1.Object) string {
+	cluster, ok := owner.(*clusteroperator.Cluster)
+	if !ok {
+		s.controller.logger.Warn("could not convert owner from JobSync into a cluster: %#v", owner)
```
dgoodwin: Mix of error and warn for these, error seems better.

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+	// Only process if all machinesets have been provisioned
+	totalCount := len(machineSets)
+	provisionedCount := 0
+	installedCount := 0
+	for _, ms := range machineSets {
+		if ms.Status.Provisioned {
+			provisionedCount++
+		}
+		if ms.Spec.NodeType == clusteroperator.NodeTypeMaster && ms.Status.Installed {
+			installedCount++
```
dgoodwin: Would this be clearer as a controlPlaneInstalled bool?

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+		}
+	}
+
+	// With at least one machine set, all machineSets in the cluster provisioned and a
+	// master that is installed, we are ready to accept nodes
+	return totalCount > 0 && provisionedCount == totalCount && installedCount > 0
```
dgoodwin: Do we have a guarantee that once a machineset reports provisioned the vms are fully up and listening? If not, what happens to the job? Will it fail and re-run until all nodes are successfully accepted?

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+	machineSetsSynced cache.InformerSynced
+
+	// clustersLister is able to list/get clusters and is populated by the shared informer passed to
+	// NewController.
+	clustersLister lister.ClusterLister
+	// clustersSynced returns true if the cluster shared informer has been synced at least once.
+	// Added as a member to the struct to allow injection for testing.
+	clustersSynced cache.InformerSynced
+
+	// jobsSynced returns true of the job shared informer has been synced at least once.
```
staebler: s/of/if/

I assume that this is copied over from other places, so it would be good to change in those places, too.

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
staebler: s/2017/2018/

cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -515,3 +517,19 @@ func startMasterController(ctx ControllerContext) (bool, error) {
 	).Run(int(ctx.Options.ConcurrentMasterSyncs), ctx.Stop)
 	return true, nil
 }
+
+func startAcceptController(ctx ControllerContext) (bool, error) {
+	if !ctx.AvailableResources[schema.GroupVersionResource{Group: "clusteroperator.openshift.io", Version: "v1alpha1", Resource: "machinesets"}] {
```
staebler: Since the controller is syncing clusters, this should check that clusters are available rather than machinesets. Or, arguably, all of these startXXXControllers should be verifying that all of the resources that they need are available, as opposed to just the ones that the controller is directly syncing.

pkg/controller/accept/accept_controller.go
csrwng: yup, will fix

pkg/controller/accept/accept_controller.go
csrwng: The approve playbook is meant to be run immediately after the node provisioning is done. But yes, if it fails, we keep on re-running it.

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+
+	for i := 0; i < workers; i++ {
+		go wait.Until(c.worker, time.Second, stopCh)
+	}
+
+	<-stopCh
+}
+
+func (c *Controller) enqueueClusterForMachineSet(machineSet *clusteroperator.MachineSet) {
+	cluster, err := controller.ClusterForMachineSet(machineSet, c.clustersLister)
```
staebler: Need to check that `cluster` is not nil. If the machine set is not owned by a cluster, then `ClusterForMachineSet` returns a nil cluster value and a nil error.

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+	// NewController.
+	clustersLister lister.ClusterLister
+	// clustersSynced returns true if the cluster shared informer has been synced at least once.
+	// Added as a member to the struct to allow injection for testing.
+	clustersSynced cache.InformerSynced
+
+	// jobsSynced returns true of the job shared informer has been synced at least once.
+	jobsSynced cache.InformerSynced
+
+	// Machines that need to be synced
```
staebler: s/Machines/Clusters/

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+		return false
+	}
+	if cluster.Status.Accepted && cluster.Status.AcceptedJobGeneration == cluster.Generation {
+		s.controller.logger.Debugf("cluster %s/%s is already accepted for its current generation, it will not be processed", cluster.Namespace, cluster.Name)
+		return false
+	}
+
+	machineSets, err := controller.MachineSetsForCluster(cluster, s.controller.machineSetsLister)
+	if err != nil {
+		s.controller.logger.Errorf("error retrieving machine sets for cluster %s/%s: %v", cluster.Namespace, cluster.Name, err)
```
staebler: Use a cluster field for the logger.

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+func (s *jobSyncStrategy) DoesOwnerNeedProcessing(owner metav1.Object) bool {
+	cluster, ok := owner.(*clusteroperator.Cluster)
+	if !ok {
+		s.controller.logger.Warn("could not convert owner from JobSync into a cluster: %#v", owner)
+		return false
+	}
+	if cluster.Status.Accepted && cluster.Status.AcceptedJobGeneration == cluster.Generation {
+		s.controller.logger.Debugf("cluster %s/%s is already accepted for its current generation, it will not be processed", cluster.Namespace, cluster.Name)
```
staebler: Use a cluster field for the logger rather than adding the cluster namespace and name directly. See `WithCluster` in pkg/logging.

pkg/controller/accept/accept_controller.go
staebler: This check is a bit more complex than it needs to be. Why don't we simply return false if we find a machine set that is not provisioned or a master machine set that is not installed. In other words, replace `provisionedCount++` and `installedCount++` with `return false`?

staebler: We also need a check for `len(machineSets) == len(cluster.Spec.MachineSets)` to ensure that all of the machine sets requested in the cluster spec have actually been created.

But as a think about this, I think we need an even stronger guarantee. We need to also ensure that there has not been a change in the cluster spec that is not going to invalidate the machine sets that currently exist. Let's say that the cluster is up and stable for generation 1 of the cluster object. The user changes the cluster version of the cluster, which will cause all of the machine sets to be re-created. The accept controller sees the change to the cluster, notices that the generation is different than the accepted-job generation and runs the accept job again. The job completes successfully, and the accepted-job generation is set to 2. Afterwards, the cluster controller sees the change to the cluster, deletes the machine set and creates new machine sets. The accept controller will not run any accept jobs for the new machine sets because the accepted-job generation makes the current generation of the cluster--which is 2.

I think that ultimately it would be better to have the accept controller sync the machinesets instead of the cluster. (1) That avoids having to keep track of whether the cluster has been updated. (2) We eventually want to be able to accept multiple machine sets anyway.

pkg/controller/accept/accept_controller.go
staebler: I think we should hold off on changing this unless we are going to change it in all of the controllers. I am not opposed to changing it in all of the controllers, but suggest we do them all at once.

csrwng:
@staebler this should be ready for another look. It's now syncing machinesets instead of clusters.

cmd/cluster-operator-controller-manager/app/options/options.go
```diff
@@ -105,6 +106,7 @@ func (s *CMServer) AddFlags(fs *pflag.FlagSet, allControllers []string, disabled
 	fs.Int32Var(&s.ConcurrentMachineSetSyncs, "concurrent-machine-set-syncs", s.ConcurrentMachineSetSyncs, "The number of machine set objects that are allowed to sync concurrently. Larger number = more responsive machine sets, but more CPU (and network) load")
 	fs.Int32Var(&s.ConcurrentMachineSyncs, "concurrent-machine-syncs", s.ConcurrentMachineSyncs, "The number of machine objects that are allowed to sync concurrently. Larger number = more responsive machines, but more CPU (and network) load")
 	fs.Int32Var(&s.ConcurrentMasterSyncs, "concurrent-master-syncs", s.ConcurrentMasterSyncs, "The number of master machine objects that are allowed to sync concurrently. Larger number = more responsive master machines, but more CPU (and network) load")
+	fs.Int32Var(&s.ConcurrentAcceptSyncs, "concurrent-accept-syncs", s.ConcurrentAcceptSyncs, "The number of cluster objects that are allowed to accept concurrently. Larger number = more responsive accept jobs, but more CPU (and network) load")
```
staebler: s/cluster/machine set/ since the accept controller is syncing on machine sets now.

pkg/apis/componentconfig/types.go
```diff
@@ -94,6 +94,11 @@ type ControllerManagerConfiguration struct {
 	// but more CPU (and network) load.
 	ConcurrentMasterSyncs int32
 
+	// concurrentAcceptSyncs is the number of cluster objects that are
```
staebler: s/cluster/machine set/

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,468 @@
...
+func (c *Controller) handleErr(err error, key interface{}) {
+	if err == nil {
+		c.queue.Forget(key)
+		return
+	}
+
+	logger := c.logger.WithField("machineSet", key)
+
+	if c.queue.NumRequeues(key) < maxRetries {
+		logger.Infof("Error syncing cluster: %v", err)
```
staebler: s/cluster/machine set/

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,468 @@
...
+		c.queue.AddRateLimited(key)
+		return
+	}
+
+	utilruntime.HandleError(err)
+	logger.Infof("Dropping cluster out of the queue: %v", err)
```
staebler: s/cluster/machine set/

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,468 @@
...
+type jobOwnerControl struct {
+	controller *Controller
+}
+
+func (c *jobOwnerControl) GetOwnerKey(owner metav1.Object) (string, error) {
+	return controller.KeyFunc(owner)
+}
+
+func (c *jobOwnerControl) GetOwner(namespace string, name string) (metav1.Object, error) {
+	return c.controller.clustersLister.Clusters(namespace).Get(name)
```
staebler: Owner of the accept job should be a machine set not a cluster.

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,468 @@
...
+
+func (s *jobSyncStrategy) GetOwner(key string) (metav1.Object, error) {
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return nil, err
+	}
+	if len(namespace) == 0 || len(name) == 0 {
+		return nil, fmt.Errorf("invalid key %q: either namespace or name is missing", key)
+	}
+	return s.controller.clustersLister.Clusters(namespace).Get(name)
```
staebler: Owner should be a machine set.

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+	if machineSet.Status.Accepted && machineSet.Status.AcceptedJobGeneration == machineSet.Generation {
+		logging.WithMachineSet(s.controller.logger, machineSet).Debugf("machine set is already accepted for its current generation, it will not be processed")
+		return false
+	}
+	return true
+}
+
+func (s *jobSyncStrategy) GetJobFactory(owner metav1.Object) (controller.JobFactory, error) {
+	cluster, ok := owner.(*clusteroperator.Cluster)
```
staebler: MachineSet, not Cluster

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,468 @@
...
+	if !ok {
+		return nil, fmt.Errorf("could not convert owner from JobSync into a cluster")
```
staebler: s/cluster/machineset/

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,468 @@
...
+
+func (s *jobSyncStrategy) OnJobFailure(owner metav1.Object) {
+	machineSet, ok := owner.(*clusteroperator.MachineSet)
+	if !ok {
+		s.controller.logger.Warn("could not convert owner from JobSync into a machine set: %#v", owner)
+		return
+	}
+	// AcceptedJobGeneration is set even when the job failed because we
+	// do not want to run the accept job again until there have been
+	// changes in the spec of the cluster.
```
staebler: s/cluster/machine set/

pkg/controller/accept/accept_controller.go
```diff
@@ -0,0 +1,468 @@
...
+		queue:      workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "master"),
+		logger:     logger,
+	}
+
+	machineSetInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.addMachineSet,
+		UpdateFunc: c.updateMachineSet,
+	})
+
+	clusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
staebler: I don't think that we need to sync machine sets in response to cluster changes.

PR #80, opened 2018-02-05
Use single compute label for compute nodes

csrwng:
Generate vars for a single compute group, instead of naming them after the machineset.
For now, multiple compute machinesets are not supported by the openshift-ansible playbooks (nodes will not be able to autojoin the cluster). This makes it so that we use the fixed "compute" group for compute machinesets. It means that for any given cluster you can only have a single compute machineset for now. Support for multiple machinesets will need to first be done on the openshift-ansible side.


csrwng:
@staebler ptal

PR #82, opened 2018-02-07
Move deployment type and version from cluster config to version.

dgoodwin:
Rename OpenshiftVersion to just Version to better accomodate possiblity
of installing base Kubernetes in the future.

Version will at worst, look something like "v3.9.0-alpha.4+e7d9503-18"
and at best, something like "v3.9.0". We may add validation on this
format once it's content is more solidified.

### DIFF
--- a/pkg/apis/clusteroperator/validation/clusterversion.go
+++ b/pkg/apis/clusteroperator/validation/clusterversion.go
@@ -23,6 +23,23 @@ import (
 	"github.com/openshift/cluster-operator/pkg/apis/clusteroperator"
 )
 
+// validDeploymentTypes is a map containing an entry for every valid DeploymentType value.
+var validDeploymentTypes = map[clusteroperator.ClusterDeploymentType]bool{
+	clusteroperator.ClusterDeploymentTypeOrigin:     true,
+	clusteroperator.ClusterDeploymentTypeEnterprise: true,
+}
+
+// validDeploymentTypeValues is an array of every valid DeploymentType value.
+var validDeploymentTypeValues = func() []string {
+	validValues := make([]string, len(validDeploymentTypes))
+	i := 0
+	for dt := range validDeploymentTypes {
+		validValues[i] = string(dt)
+		i++
+	}
+	return validValues
+}()
+
 // ValidateClusterVersion validates a cluster version being created.
 func ValidateClusterVersion(cv *clusteroperator.ClusterVersion) field.ErrorList {
 	allErrs := field.ErrorList{}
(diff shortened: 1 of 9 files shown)


pkg/apis/clusteroperator/validation/clusterversion.go
```diff
@@ -69,6 +69,16 @@ func ValidateClusterVersionSpec(spec *clusteroperator.ClusterVersionSpec, fldPat
 		allErrs = append(allErrs, field.Required(vmImagesPath, "must define VM images"))
 	}
 
+	deploymentTypePath := fldPath.Child("deploymentType")
+	if spec.DeploymentType == "" {
+		allErrs = append(allErrs, field.Required(deploymentTypePath, "must define deployment type"))
```
staebler: I would like for this validation to be stronger than just empty or not empty. We should validate that the deployment type is one of the valid enumerated values.

pkg/openapi/openapi_generated.go
```diff
@@ -10122,15 +10122,7 @@ func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenA
 			Schema: spec.Schema{
 				SchemaProps: spec.SchemaProps{
 					Description: "Duration is a wrapper around time.Duration which supports correct marshaling to YAML and JSON. In particular, it marshals into strings, which can be used as map keys in json.",
-					Properties: map[string]spec.Schema{
```
staebler: This is outside of the scope of this PR, but there seems to be a problem where this part of the schema varies depending upon whether the build is done in docker or outside of docker.

pkg/apis/clusteroperator/validation/clusterversion.go
dgoodwin: My mistake I thought the use of ClusterDeploymentType for that handled that for us like a java enum, will get this fixed.

staebler: When in doubt, assume that go doesn't handle anything for us.  ;)

dgoodwin:
Updated with a test fix as well.

pkg/apis/clusteroperator/validation/clusterversion.go
```diff
@@ -69,6 +69,18 @@ func ValidateClusterVersionSpec(spec *clusteroperator.ClusterVersionSpec, fldPat
...
+	} else if spec.DeploymentType != clusteroperator.ClusterDeploymentTypeOrigin && spec.DeploymentType != clusteroperator.ClusterDeploymentTypeEnterprise {
+		allErrs = append(allErrs, field.Invalid(deploymentTypePath, spec.DeploymentType, "unknown deployment type"))
```
staebler: `field.NotSupported` is a better choice than `field.Invalid` for enumerated fields. See [here](https://github.com/openshift/cluster-operator/blob/551c8e210cb9349d6f709aa303909bc665e09280/pkg/apis/clusteroperator/validation/validation.go#L48) for an example.

dgoodwin: Fixed.

PR #84, opened 2018-02-07
Perform containerized installation of master

csrwng:
Adds the containerized=True variable to the variables passed to ansible playbooks.


csrwng:
@abutcher ptal

butcher:
@abutcher should be mentioned :point_up: here. I was very surprised :smile:

PR #85, opened 2018-02-09
Allow machineset scaling

csrwng:
Modifies cluster controller to ignore size when deciding whether to recreate a machineset due to a spec change.

Updates unit tests to use instance type instead of size to test machineset mutation.

### DIFF
--- a/pkg/controller/cluster/cluster_controller.go
+++ b/pkg/controller/cluster/cluster_controller.go
@@ -488,19 +488,21 @@ func (c *Controller) manageMachineSets(machineSets []*clusteroperator.MachineSet
 		}
 		machineSetConfig.Hardware = mergedHardwareSpec
 
-		machineSetToCreate, updateMachineSet, deleteMachineSet, err := c.manageMachineSet(cluster, clusterVersion, clusterMachineSets[i], machineSetConfig, machineSetPrefixes[i])
+		machineSet, updateMachineSet, deleteMachineSet, err := c.manageMachineSet(cluster, clusterVersion, clusterMachineSets[i], machineSetConfig, machineSetPrefixes[i])
 		if err != nil {
 			errCh <- err
 			continue
 		}
 
-		if machineSetToCreate != nil {
-			machineSetsToCreate = append(machineSetsToCreate, machineSetToCreate)
+		if machineSet != nil {
+			if updateMachineSet {
+				machineSetsToUpdate = append(machineSetsToUpdate, machineSet)
+			} else {
+				machineSetsToCreate = append(machineSetsToCreate, machineSet)
+			}
 		}
 		if deleteMachineSet {
 			machineSetsToDelete = append(machineSetsToDelete, clusterMachineSets[i])
-		} else if updateMachineSet {
-			machineSetsToUpdate = append(machineSetsToUpdate, clusterMachineSets[i])
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


csrwng:
@staebler ptal

csrwng:
I have now updated the cluster controller to update machine set size based on what's in the cluster spec.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -498,6 +499,8 @@ func (c *Controller) manageMachineSets(machineSets []*clusteroperator.MachineSet
 		}
 		if deleteMachineSet {
 			machineSetsToDelete = append(machineSetsToDelete, clusterMachineSets[i])
+		} else if updateMachineSet {
+			machineSetsToUpdate = append(machineSetsToUpdate, clusterMachineSets[i])
```
staebler: Replace `clusterMachineSets[i]` with `machineSetToCreate`. The former is what is currently in storage; the latter is what we want to update to. It would probably behoove us to rename `machineSetToCreate`, too, since it is not used only for creating a machine set.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -543,10 +546,26 @@ func (c *Controller) manageMachineSets(machineSets []*clusteroperator.MachineSet
 		wg.Wait()
 	}
 
+	if len(machineSetsToUpdate) > 0 {
```
staebler: It would be helpful to add unit tests that exercise updating the machine set.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -575,32 +594,52 @@ func (c *Controller) manageMachineSets(machineSets []*clusteroperator.MachineSet
...
 			machineSet.Spec.ClusterVersionRef, clusterVersion.Namespace, clusterVersion.Name)
 		machineSet, err := buildNewMachineSet(cluster, clusterVersion, clusterMachineSetConfig, machineSetNamePrefix)
-		return machineSet, true, err
+		return machineSet, false, true, err
+	}
+
+	if machineSet.Spec.Size != clusterMachineSetConfig.Size {
+		machineSet.Spec.Size = clusterMachineSetConfig.Size
+		msLog.Infof("machine set size has changed from %d to %d", clusterMachineSetConfig.Size, machineSet.Spec.Size)
+		return nil, true, false, nil
```
staebler: We need to return the updated machine set so that the controller can use it to send the update machine set to the api server.

staebler:
@csrwng After looking into what happens when a machine set is re-created as part of my efforts on #87, I am not convinced that there is a benefit to updating the machineset over re-creating it even for scaling. The ansible playbook that is run is exactly the same regardless of whether the machineset is new or updated. The end result on AWS is the same either way. What have you seen different between updating and re-creating? Maybe there is some other thing that has changed with #87 that obviates needing to update instead of re-create.

csrwng:
@staebler the reason I opted for an update and not a recreate is that my understanding is that at some point a delete of a machineset would result in it getting deprovisioned. Are we thinking that won't be the case?

staebler:
@csrwng Ah, that makes sense. So any field changes to the machine set that should result in a re-creation of the ASG should correspond to a re-creation of the machine set. It would be helpful to call that out in the comment of `shouldRecreateMachineSet`, assuming that there will eventually be more fields that can change without needing to re-create the ASG.

pkg/controller/cluster/cluster_controller.go
csrwng: Yes, I broke this with a change after I had tested it. Will fix and add the relevant unit test.

csrwng:
I'll add the comment

PR #86, opened 2018-02-09
Convert region AMIs to a list.

dgoodwin:
Per API guidelines, convert the map of regions to AMI to use, to a list of
structs which now also allow defining an optional master AMI to be used
if set. If none is set, masters will fall back to using the compute AMI
as we do today.

Logic to utilize when provisioning to follow.


pkg/apis/clusteroperator/types.go
```diff
@@ -279,6 +289,7 @@ type ClusterStatus struct {
 	Conditions []ClusterCondition
 
 	// ClusterVersionRef references the resolved clusterversion the cluster should be running.
+
```
staebler: Did you intentionally add this blank line?

pkg/apis/clusteroperator/types.go
```diff
@@ -121,7 +121,17 @@ type VMImages struct {
...
 type AWSVMImages struct {
-	AMIByRegion map[string]string
+	RegionAMIs []AWSRegionAMIs
+}
+
+// AWSRegionAMIs defines which AMI to use for a node type in a given region.
+type AWSRegionAMIs struct {
+	Region string
+	// ComputeAMI is the ID of the AMI to use for compute nodes in the cluster. If no MasterAMI is defined, it will be used for masters as well.
+	ComputeAMI string
```
staebler: What are your thoughts on calling this AMI instead? I think that since it is the default AMI for all nodes, it makes more sense to call it AMI.

pkg/apis/clusteroperator/validation/clusterversion.go
```diff
@@ -126,3 +126,47 @@ func ValidateYumRepository(repo *clusteroperator.YumRepository, fldPath *field.P
...
+
+	regionsSeen := map[string]bool{}
+	amisPath := fldPath.Child("awsVMImages").Child("regionAMIs")
+	if len(vmImages.AWSImages.RegionAMIs) == 0 {
+		allErrs = append(allErrs, field.Required(amisPath, "must define AMIs for at least one region"))
+	} else {
+		for i, ami := range vmImages.AWSImages.RegionAMIs {
+			allErrs = append(allErrs, ValidateRegionAMIs(&ami, amisPath.Index(i))...)
+			if _, previouslySeen := regionsSeen[ami.Region]; previouslySeen {
+				allErrs = append(allErrs, field.Invalid(amisPath.Index(i), ami, fmt.Sprintf("duplicated AMIs found for region: %s", ami.Region)))
```
staebler: Use `field.Duplicate` and use `amisPath.Index(i).Child("region")` as the path.

pkg/apis/clusteroperator/validation/clusterversion.go
```diff
@@ -126,3 +126,47 @@ func ValidateYumRepository(repo *clusteroperator.YumRepository, fldPath *field.P
...
+	allErrs := field.ErrorList{}
+
+	if regionAMIs.Region == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("region"), "must define region"))
+	}
+
+	if regionAMIs.ComputeAMI == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("computeAMI"), "must define compute AMI ID"))
+	}
+
```
staebler: Validate that if MasterAMI is defined it is not an empty string.

pkg/controller/infra/infra_controller.go
```diff
@@ -459,7 +459,7 @@ func (s *jobSyncStrategy) ProcessDeletedOwner(owner metav1.Object) error {
 
 	// Clear the finalizer for the cluster
 	finalizers.Delete(clusteroperator.FinalizerClusterOperator)
-	cluster.ObjectMeta.Finalizers = finalizers.List()
```
staebler: Why is this being removed?

dgoodwin: My bad, been trying to adjust to vscode and I'm doing something that kills lines without noticing.

dgoodwin:
Updated after review, let me know when to squash.

Moved computeAMI to ami.

staebler:
@dgoodwin Looks good. Go ahead and squash it.

dgoodwin:
All set.

PR #87, opened 2018-02-10
Run all playbook jobs in parallel

staebler:
* Each controller runs its playbook without worrying about whether other playbooks that set up dependencies have been run yet. If those dependencies are have not been set up yet, then the dependent playbook will fail and be retried. This greatly simplifies the controllers because they do not have to worry about checking on the state into which other controllers have put the kube objects.
* A MachineSet object contains all of the information that is needed to run playbooks for the MachineSet. Controllers syncing a MachineSet no longer need to fetch the owning Cluster.

In my experience, running all the ansible playbooks in parallel has actually slowed the total time that it takes to go from cold start to a working cluster. This may be due to the exponential backoffs of the jobs making it so that the successful run of each job starts later than it would have if the controller started it as soon as its dependent job completed. We may be able to improve upon this by tweaking the settings for the jobs.

### DIFF
--- a/pkg/apis/clusteroperator/types.go
+++ b/pkg/apis/clusteroperator/types.go
@@ -394,6 +394,9 @@ type MachineSetSpec struct {
 	// MachineSetConfig is the configuration for the MachineSet
 	MachineSetConfig
 
+	// ClusterHardware specifies the hardware that the cluster will run on
+	ClusterHardware ClusterHardwareSpec
+
 	// ClusterVersionRef references the clusterversion the machine set is running.
 	ClusterVersionRef corev1.ObjectReference
 }
--- a/cmd/cluster-operator-controller-manager/app/controllermanager.go
+++ b/cmd/cluster-operator-controller-manager/app/controllermanager.go
@@ -479,7 +479,6 @@ func startMachineSetController(ctx ControllerContext) (bool, error) {
 		return false, nil
 	}
 	go machineset.NewController(
-		ctx.InformerFactory.Clusteroperator().V1alpha1().Clusters(),
 		ctx.InformerFactory.Clusteroperator().V1alpha1().MachineSets(),
 		ctx.KubeInformerFactory.Batch().V1().Jobs(),
 		ctx.ClientBuilder.KubeClientOrDie("clusteroperator-machine-set-controller"),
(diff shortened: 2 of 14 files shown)


pkg/apis/clusteroperator/types.go
```diff
@@ -384,6 +384,9 @@ type MachineSetSpec struct {
 	// MachineSetConfig is the configuration for the MachineSet
 	MachineSetConfig
 
+	// ClusterHardware specifies the hardware that the cluster will run on
+	ClusterHardware ClusterHardwareSpec
```
csrwng: Do we foresee having other settings from the cluster (that are not part of the hardware spec) that should also be included in the machineset?

dgoodwin:
Haven't read through the code yet but wondering initially, could we hit a scenario where 2/3 master VMs are booted, the job succeeds because two hosts are live, leaving a third unconfigured for some time? Or does the in-job hosts lookup know that we're still missing a vm and thus report an error via ansible's return code as the host is unreachable.

staebler:
@dgoodwin I don't definitively know the answer to that question. However, there is a part of the install.yml playbook that waits to be able to ssh into the master machine. I have only run it with one master, so I am not certain whether it would wait to be able to ssh into *all* the masters.

pkg/apis/clusteroperator/types.go
staebler: I don't know. Currently, the only thing in `ClusterSpec` that is not passed along to `MachineSetSpec` in some way or another is the `ClusterConfigSpec`. It shouldn't be too difficult to adjust what we pass along to `MachineSetSpec` in the future if we need something more. Are you suggesting that it might be better to have an explicit struct containing all of the information passed along from the cluster rather than using the `ClusterHardwareSpec` at the top-level in the `MachineSetSpec`?

csrwng: > Are you suggesting that it might be better to have an explicit struct containing all of the information passed along from the cluster rather than using the ClusterHardwareSpec at the top-level in the MachineSetSpec?

Not sure that's better. So, no :) 
I guess we can add other fields to the machine set spec as needed if we need something else from the cluster.

staebler:
@csrwng It appears that there is not a way to tinker with the rate of backoffs for the job. We may find that it is better for us to create the pods directly rather than going through a job, if we need tighter control over the backoff characteristics.

We may also find that there is benefit in (1) breaking the playbooks into smaller pieces that have no dependencies with each other and can succeed in parallel and (2) having actual dependencies where we know that one playbook absolutely cannot do any work before another playbook succeeds. For example, there may be pieces of cluster infra provisioning that can be run in parallel. And there are certainly parts of master install that cannot succeed prior to the master machines being provisioned.

PR #88, opened 2018-02-12
Deprovision clusters and machine sets

staebler:
When a cluster or a machine set is deleted, a job is created to deprovision. The cluster or machine set is blocked from being fully deleted until the playbook completes successfully. Note that if the job fails, then the user will have to manually delete the cluster or machine set by removing the appropriate finalizer. This gives the user an opportunity to see that there was a failure and do any manual deprovisioning needed.

By default, when a cluster is deleted, the machine sets that it owns are deleted as well, since the owner reference on the machine set for the cluster has the `blockDeletion` field set to true.

There is not *yet* a playbook to deprovision the machine set. Consequently, a job to deprovision a machine set will never succeed, and a machine set will never be fully deleted with manual intervention.

### DIFF
--- a/pkg/controller/jobsync.go
+++ b/pkg/controller/jobsync.go
@@ -26,6 +26,7 @@ import (
 	kapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // JobSync is used by a controller to sync an object that uses a job to do
@@ -47,6 +48,10 @@ const (
 	JobSyncProcessed JobSyncConditionType = "Processed"
 	// JobSyncProcessingFailed indicates that the processing job has failed.
 	JobSyncProcessingFailed JobSyncConditionType = "ProcessingFailed"
+	// JobSyncUndoing indicates that the undoing job is in progress.
+	JobSyncUndoing JobSyncConditionType = "Undoing"
+	// JobSyncUndoFailed indicates that the undoing job has failed.
+	JobSyncUndoFailed JobSyncConditionType = "UndoFailed"
 )
 
 const (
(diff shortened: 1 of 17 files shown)


pkg/controller/jobsync.go
```diff
@@ -229,6 +284,35 @@ func (s *jobSync) setOwnerStatusForInProgressJob(original metav1.Object, job *v1
...
+	owner := s.strategy.DeepCopyOwner(original)
+	finalizers := sets.NewString(owner.GetFinalizers()...)
+	finalizers.Insert(s.getFinalizerName())
+	owner.SetFinalizers(finalizers.List())
+	return s.strategy.UpdateOwnerStatus(original, owner)
+}
+
+func (s *jobSync) getFinalizerName() string {
+	finalizerName := "openshift/cluster-operator-" + s.jobControl.GetJobPrefix()
+	if !validateFinalizer.MatchString(finalizerName) {
```
csrwng: what is the purpose of this condition? it looks like the finalizerName would never match here.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -62,6 +62,8 @@ const (
 	masterProvisioningPlaybook = "playbooks/aws/openshift-cluster/provision.yml"
 
 	computeProvisioningPlaybook = "playbooks/aws/openshift-cluster/provision_nodes.yml"
+
+	deprovisioningPlaybook = "playbooks/aws/openshift-cluster/deprovision.yml"
```
csrwng: looks like the playbook name is `playbooks/aws/openshift-cluster/uninstall_nodes.yml` (from https://github.com/openshift/openshift-ansible/pull/7114). Not sure if we can tell it to deprovision one asg at a time. Will need to coordinate with @joelddiaz

pkg/controller/jobsync.go
staebler: The purpose of this is to ensure that the finalizer name is valid. The finalizer name cannot end in a hyphen, which it would for a job prefix that ends in a hyphen. Given that all of the job prefixes end in hyphens, I suppose it would be better to simply append a valid character to the end in all cases, even when the finalizer may have already been valid to begin with. Is that what you are suggesting?

pkg/controller/machineset/machineset_controller.go
staebler: I didn't want to use an actual playbook here until the deprovision playbook work was complete.

pkg/controller/jobsync.go
csrwng: Sorry no, my bad, I was thinking that MatchString would require that the regexp match the entire string, which would require the string to only be a single char :) 
Makes sense what it's doing

pkg/controller/machineset/machineset_controller.go
csrwng: Cool we can update when that merges

PR #89, opened 2018-02-12
Remove manual deletion of configmaps from jobControl

staebler:
The manual deletion of configmaps is not necessary. The garbage collector will delete the configmaps automatically when the job is deleted since the configmaps are owned by the jobs and are set to block deletion of the jobs.

I have explicitly set the propagation policy when deleting the job even though it is not necessary as that is the default behavior. If for whatever reason the job resource is changed in the future in such a way that background deletion of dependents is now longer the default, then clust-op will still be cleaning up the configmaps (and pods) that clust-op creates.

### DIFF
--- a/pkg/controller/jobcontrol.go
+++ b/pkg/controller/jobcontrol.go
@@ -433,7 +433,11 @@ func (c *jobControl) deleteOldJobs(ownerKey string, oldJobs []*kbatch.Job, logge
 	for i, ng := range oldJobs {
 		go func(ix int, job *kbatch.Job) {
 			defer wg.Done()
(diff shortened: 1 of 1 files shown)


dgoodwin:
It looks like infra provision job config map is owned by the cluster itself, and most others (machine set provision, master install) are owned by their respective machinesets, I haven't spotted any owned but their jobs yet.

Should any/all of these be owned by jobs?

staebler:
@dgoodwin Ah, you're right. I am confused about what I tested then, because I definitely tested that the config maps were removed when the job was deleted. I must have been using out-of-date images.

PR #90, opened 2018-02-12
Generate unit test mocks as part of make test-unit

### LINKED ISSUE #72: Generate mocks from make
The mocks used in unit testing need to be generated via `make`. Currently, they need to be generated manually when there are relevant changes in the mocked interfaces.

```
go get github.com/golang/mock/gomock
go get github.com/golang/mock/mockgen

go generate pkg/controller
```

staebler:
Closes #72.

* The `go build` commands will execute regardless of whether files have changed. We will now rely on the go builder to determine what code needs to be re-compiled.
* Because we are relying on the go builder to determine what needs to be rebuilt, we no longer need to clean out the bin directories before building for each architecture.
* If you are not building inside docker, you will need to run `go get github.com/golang/mock/gomock` and `go get github.com/golang/mock/mockgen` in your local environment.


dgoodwin:
Really nice. I'd be tempted to doc the mock requirements but if we default to dockerized, there's not really any need or good place for it.

staebler:
@dgoodwin Eventually we should have docs for how to make builds and run tests. That would be the place to include these dependencies. We already have others like golint.

dgoodwin:
Getting make verify failures with NO_DOCKER, trying to debug:

```
Running gofmt:              
Running golint and go vet:  
go vet ./cmd/... ./contrib/... ./pkg/... ./test/...      
pkg/client/clientset_generated/clientset/fake/clientset_generated.go:46: literal copies lock value from fakePtr: github.com/openshift/cluster-operator/vendor/k8s.io/client-go/testing.Fake
exit status 1               
pkg/client/clientset_generated/internalclientset/fake/clientset_generated.go:46: literal copies lock value from fakePtr: github.com/openshift/cluster-operator/vendor/k8s.io/client-go/testing.Fake
exit status 1               
pkg/controller/cluster/cluster_controller_test.go:462: arg machineSet for printf verb %t of wrong type: *github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1.MachineSet
exit status 1               
test/integration/controller_test.go:297: literal copies lock value from fakePtr: github.com/openshift/cluster-operator/vendor/k8s.io/client-go/testing.Fake
exit status 1               
make: *** [Makefile:218: verify] Error 1               
```

staebler:
@dgoodwin No. The reason for defaulting to in docker is so that jenkins build in a container in which we can include golint, gomock, and the other dependencies that we use. `make verify` should not fail in or out of docker.

PR #91, opened 2018-02-13
Add SSHUser to cluster hardware info.

dgoodwin:
This is added to sit along-side the SSHKey.

Today this setting is largely dependent on the AMI and we require the
user to exist already.

Long term this setting will probably die, being replaced by global
cluster operator configuration for a user that we will ensure is created
on every host via the upcoming config DaemonSet work from kwoodson.
For now this gives us flexibility to adjust per cluster as ami's come
and go.

### DIFF
--- a/pkg/apis/clusteroperator/v1alpha1/defaults.go
+++ b/pkg/apis/clusteroperator/v1alpha1/defaults.go
@@ -20,12 +20,16 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const defaultSSHUser = "clusteroperator"
+
 func addDefaultingFuncs(scheme *runtime.Scheme) error {
 	return RegisterDefaults(scheme)
 }
 
 func SetDefaults_ClusterSpec(spec *ClusterSpec) {
-	// No defaults
+	if spec.Hardware.AWS != nil && spec.Hardware.AWS.SSHUser == "" {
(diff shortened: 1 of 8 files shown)


dgoodwin:
CC @abutcher @kwoodson

dgoodwin:
Alternatively I could throw this onto ClusterVersion to live with the AMI ID, but this separates it from the ssh key.

pkg/apis/clusteroperator/v1alpha1/defaults.go
```diff
@@ -20,12 +20,16 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const defaultSSHUser = "root"
```
staebler: Do we really want the default SSH user to be root? That seems unnecessarily dangerous.

dgoodwin: Via Kenny this was what they're using today, but we could go without a default instead and make it a required field, and you have the thus opt into using root.

dgoodwin: Switching it to clusteroperator, as that's what I'd like to use eventually when we can create/config the user.

dgoodwin: Updated.

PR #93, opened 2018-02-14
Add controller to install components on master

csrwng:
Adds a controller to install components on a cluster that has accepted compute/infra nodes.

### DIFF
--- a/pkg/apis/clusteroperator/types.go
+++ b/pkg/apis/clusteroperator/types.go
@@ -434,6 +434,20 @@ type MachineSetStatus struct {
 	// Conditions includes more detailed status of the MachineSet
 	Conditions []MachineSetCondition
 
+	// ComponentsInstalled is true if the additional components needed for the cluster
+	// have been installed
+	ComponentsInstalled bool
+
+	// ComponentsInstalledJobGeneration is the generation of the machine set resource used to
+	// to generate the latest completed component installation job. The value will be set
+	// regardless of the job having succeeded or failed.
+	ComponentsInstalledJobGeneration int64
+
+	// ComponentInstallationJob is the job that is actively performing installation
+	// of components on the machine set.
+	// +optional
+	ComponentInstallationJob *corev1.LocalObjectReference
+
 	// Installed is true if the software required for this machine set is installed
 	// and running.
 	Installed bool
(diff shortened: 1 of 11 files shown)


csrwng:
@staebler @dgoodwin this is functional in that it will run the job, but the job is still failing due to the current playbooks checking for nodes with a particular label which assumes a static inventory. That's something that will need to be fixed on the openshift-ansible side.

pkg/apis/clusteroperator/types.go
```diff
@@ -525,6 +539,18 @@ const (
 	// on this machine set.
 	MachineSetInstalled MachineSetConditionType = "Installed"
 
+	// MachineSetComponentsInstalling is true if OpenShift components are being installed on
+	// this machine set.
+	MachineSetComponentsInstalling MachineSetConditionType = "Installing"
```
staebler: I don't like that we have two different `MachineSetConditionType` constants that have the same value. Can we either (1) use the same const in both cases or (2) change the value for one of the consts. My preference would be (2) where, for example, "Installing" becomes "ComponentsInstalling" for `MachineSetComponentsInstalling`.

pkg/controller/components/components_controller.go
```diff
@@ -0,0 +1,428 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
staebler: s/2107/2018/

pkg/controller/components/components_controller.go
```diff
@@ -0,0 +1,428 @@
...
+	case controller.JobSyncProcessed:
+		return clusteroperator.MachineSetComponentsInstalled
+	case controller.JobSyncProcessingFailed:
+		return clusteroperator.MachineSetComponentsInstallationFailed
+	default:
+		return clusteroperator.MachineSetConditionType("")
+	}
+}
+
+func loggerForMachineSet(logger log.FieldLogger, machineSet *clusteroperator.MachineSet) log.FieldLogger {
```
staebler: Use `WithMachineSet` from pkg/logging/logging.go.

pkg/apis/clusteroperator/types.go
csrwng: @staebler thx, that was not intentional, rather careless copy/paste. Will fix.

csrwng:
comments addressed

PR #96, opened 2018-02-16
Complete functional cluster version installs.

dgoodwin:
Clusters will now install the version specified.

YumRepositories and the previous specification of the AMI in the API are
both removed.

Upgrade (changing the cluster version for a cluster/machine set) is not
yet supported.


staebler:
@dgoodwin Delete bin/openapi-gen, and run `make build` or `make .generate_files` again.

dgoodwin:
Doesn't seem to register any changes I can commit, investigating.

staebler:
Is it rebuilding the generated files? If not, then you can touch the types.go file to provoke re-generation.

dgoodwin:
Found it, should be good now.

pkg/ansible/generate.go
```diff
@@ -332,21 +338,77 @@ func GenerateClusterWideVars(name string, hardwareSpec *coapi.ClusterHardwareSpe
...
+func lookupAMIForMachineSet(machineSet *coapi.MachineSet, clusterVersion *coapi.ClusterVersion) (string, error) {
+	for _, regionAMI := range clusterVersion.Spec.VMImages.AWSImages.RegionAMIs {
+		if regionAMI.Region == machineSet.Spec.ClusterHardware.AWS.Region {
+			if machineSet.Spec.MachineSetConfig.NodeType == coapi.NodeTypeMaster && regionAMI.MasterAMI != nil {
+				return *regionAMI.MasterAMI, nil
+			}
+			return regionAMI.AMI, nil
+		}
+	}
+	return "", fmt.Errorf("no AMI defined for cluster version %s/%s in region %v", clusterVersion.Namespace, clusterVersion.Name, machineSet.Spec.ClusterHardware.AWS.Region)
```
staebler: This is a good error here. But in addition to this, we need to have cluster controller block creation of machine sets when the cluster is for a region that is not included in the cluster version. The cluster should also have a condition added in that case so that the user can see that there is a problem that they need to address.

dgoodwin:
Intermittent failures on master, filed as https://github.com/openshift/cluster-operator/issues/106

dgoodwin:
Ready for re-review I think.

dgoodwin:
Tests clear and ready for re-review.

csrwng:
lgtm (also tested manually and worked fine). Will let @staebler have a chance to take a look

dgoodwin:
I think I've got those addressed now. Condition added/cleared if there's a mismatch, regex updated, couple new tests added.

PR #97, opened 2018-02-16
Circumvent long delays in job completion after pod success

staebler:
The job controller uses longer and longer delays as the pods of a job fail. For the compute provision jobs, there can be large delays, in excess of 5 minutes, between when a pod completes successfully and when a job is updated to have a completed status. For clust-op controllers that wait for other jobs to complete before starting their jobs, this can significantly increase the time that it takes to stand up a cluster.

As part of a circumvention to this problem, this PR limits the number of failures for our jobs to 3 and re-creates the job when that limit is reached. As future work, it would be good to also (1) include some backoff wait between job failure and job creation and (2) stop the job re-creation after a sufficiently long time of failures.

Note: In the interest of avoiding a rebase, this is built on top of #88. If needed, I can separate it into its own PR.

### DIFF
--- a/pkg/controller/jobcontrol.go
+++ b/pkg/controller/jobcontrol.go
@@ -54,6 +54,12 @@ const (
 	// JobControlJobWorking indicates that there is an outstanding job
 	// processing the current generation of the owner.
 	JobControlJobWorking JobControlResult = "JobWorking"
+	// JobControlJobSucceeded indicates that the job for current
+	// generation of the owner has completed successfully.
+	JobControlJobSucceeded JobControlResult = "JobSucceeded"
+	// JobControlJobFailed indicates that the job for current
+	// generation of the owner has failed.
+	JobControlJobFailed JobControlResult = "JobFailed"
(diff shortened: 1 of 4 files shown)


staebler:
Here are the rough times that it takes on my development machine to go from 0 to complete cluster with accepted compute nodes.

Without changes: > 20 minutes
With changes: 10 minutes

pkg/controller/jobcontrol.go
```diff
@@ -54,6 +54,12 @@ const (
 	// JobControlJobWorking indicates that there is an outstanding job
 	// processing the current generation of the owner.
 	JobControlJobWorking JobControlResult = "JobWorking"
+	// JobControlJobSucceeded indicates that the job for current
+	// generation of the owner has completed succeessfully.
```
csrwng: s/succeessfully/successfully/

staebler:
Rebased.
Added issue #98.

dgoodwin:
Very nice, 10min is a lot further along at this point than I would have thought we'd be.

PR #102, opened 2018-02-20
Restart jobs when they fail

csrwng:
Modifies job sync and job control to keep retrying when a job fails. 

The job generation field in the cluster and machineset status structs now only gets a value if a job has completed successfully for a given generation.

### DIFF
--- a/pkg/apis/clusteroperator/types.go
+++ b/pkg/apis/clusteroperator/types.go
@@ -276,8 +276,7 @@ type ClusterStatus struct {
 	Provisioned bool
 
 	// ProvisionedJobGeneration is the generation of the cluster resource used to
-	// to generate the latest completed infra provisioning job. The value will be set
-	// regardless of the job having succeeded or failed.
+	// generate the latest successful infra provisioning job.
 	ProvisionedJobGeneration int64
 
 	// ProvisionJob is the job that is actively performing provisioning
@@ -443,8 +442,7 @@ type MachineSetStatus struct {
 	ComponentsInstalled bool
 
 	// ComponentsInstalledJobGeneration is the generation of the machine set resource used to
-	// to generate the latest completed component installation job. The value will be set
-	// regardless of the job having succeeded or failed.
+	// generate the latest successful component installation job.
 	ComponentsInstalledJobGeneration int64
 
 	// ComponentInstallationJob is the job that is actively performing installation
@@ -457,8 +455,7 @@ type MachineSetStatus struct {
 	Installed bool
 
 	// InstalledJobGeneration is the generation of the machine set resource used to
-	// to generate the latest completed installation job. The value will be set
-	// regardless of the job having succeeded or failed.
+	// generate the latest successful installation job.
 	InstalledJobGeneration int64
 
 	// InstallationJob is the job that is actively performing installation
@@ -471,8 +468,7 @@ type MachineSetStatus struct {
 	Provisioned bool
 
 	// ProvisionedJobGeneration is the generation of the machine set resource used to
-	// to generate the latest completed hardware provisioning job. The value will be set
-	// regardless of the job having succeeded or failed.
+	// to generate the latest successful hardware provisioning job.
 	ProvisionedJobGeneration int64
 
 	// ProvisionJob is the job that is actively performing provisioning
@@ -484,8 +480,7 @@ type MachineSetStatus struct {
 	Accepted bool
 
 	// AcceptedJobGeneration is the generation of the machine set resource used to
-	// run the latest completed accept job. The value will be set regardless of
-	// the job having succceeded or failed.
+	// run the latest successful accept job.
 	AcceptedJobGeneration int64
 
 	// AcceptJob is the job that is actively running to accept nodes from this machine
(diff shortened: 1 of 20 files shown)


csrwng:
@staebler @dgoodwin ptal

pkg/controller/jobcontrol.go
```diff
@@ -207,34 +207,16 @@ func (c *jobControl) ControlJobs(
 			currentJob = job
 		}
 		if jobOwnerGeneration(job) == owner.GetGeneration() {
-			generationJob = job
+			if isActive(job) {
+				activeJob = job
+			} else if isSuccessful(job) {
+				successfulJob = job
+			}
```
staebler: Are we intentionally not deleting failed jobs for the current generation? All jobs for a given generation will persist until the generation is updated.

pkg/apis/clusteroperator/types.go
```diff
@@ -276,8 +276,7 @@ type ClusterStatus struct {
 	Provisioned bool
 
 	// ProvisionedJobGeneration is the generation of the cluster resource used to
-	// to generate the latest completed infra provisioning job. The value will be set
-	// regardless of the job having succeeded or failed.
+	// generate the latest successful infra provisioning job.
```
staebler: Do we still have plans to cease starting new jobs after they fail for a long amount of time?

pkg/controller/jobcontrol.go
```diff
@@ -244,6 +226,37 @@ func (c *jobControl) ControlJobs(
 		return JobControlDeletingJobs, nil, err
 	}
 
+	if currentJobName != "" {
```
staebler: Do we even care about tracking current job any more? The only reason we tracked it previously was so that we could tell when a job was lost. If we are going to re-create the job even when it was lost, then I don't see much point in distinguish a lost job from a failed job.

pkg/controller/jobcontrol.go
csrwng: I was thinking that we wouldn't for the current generation so you could get an idea of why they're failing. Since we create a new job every 3 tries, and suppose that you have a job that takes 1 hr to fail ... and you're in the middle of a new job you just started. It'd be nice to have the previous job so you can see what happened.

pkg/apis/clusteroperator/types.go
csrwng: Yes but would we not track that explicitly in the owner's status instead of here?

pkg/controller/jobcontrol.go
csrwng: Yes, I agree we don't need it. However, it felt like a lot of changes all at once, but if you're fine with me removing it, I can proceed.

pkg/controller/jobcontrol.go
staebler: That makes sense. I was just verifying that it was intentional. At some point, we'll need to clean up old jobs (failed and succeeded), but that is a problem that we can address later.

I guess one concern that I have is that if there is a job that is perpetually failing, then we will use up a lot of resources keeping around the failed jobs and their pods.

pkg/apis/clusteroperator/types.go
staebler: This is the owner's status. I don't explicitly have an objection to tracking it elsewhere. Do you have a suggestion about what that would look like?

pkg/controller/jobcontrol.go
staebler: I vote for removing it now. It is only going to make the code simpler.

pkg/apis/clusteroperator/types.go
csrwng: I would repurpose the Failed condition to label the provisioning/install as failed. It really serves no purpose right now since we're starting a new job every time the job fails anyway.

pkg/controller/jobcontrol.go
csrwng: We could trim the number of jobs we keep around. Is it ok if we do it as a follow up ?

staebler: Yes.

pkg/apis/clusteroperator/types.go
staebler: Would you expect the controller to clear the condition when it notices a change in the generation? If not, then we would not have a way to differentiate between a failed state from a previous generation as opposed to a failed state from the current generation.

If we did that, then I suggest that we change the semantics of the job generation to be an observed generation that is set as soon as the controller sees the new generation--as opposed to setting the generation when the job succeeds.

In any event, we can address that later when we do work for stopping jobs.

staebler: Here are some more details about my proposal.

- If the controller sees generation != observedGeneration, the controller clears out job succeeded/failed conditions and records the observedGeneration as the current generation.
- If the controller sees generation == observedGeneration, the controller also looks at job succeeded/failed conditions to determine whether there is work that needs to be done. If neither succeeded/failed conditions are true, then there is work to do.

csrwng: Sounds good to me.

staebler:
It is not a straight forward fix for the integration tests. I would be fine with deleting the integration tests in the very short term to get this PR merged. I can rework the integration tests.

csrwng:
@staebler I attempted to fix them anyway. If you don't think it's a good fix, I can just remove them.

staebler:
@csrwng That's a good fix. I didn't have a good feel for what the change was going to entail to fix the integration tests and didn't want to have us wait to get this bug fixed. Changes all look good to me.

PR #105, opened 2018-02-22
Hookup deprovisioning playbook

csrwng:
Updates the playbook called for machineset deprovisioning
Also updates the ami used in the sample cluster

### DIFF
--- a/contrib/examples/cluster.yaml
+++ b/contrib/examples/cluster.yaml
@@ -15,7 +15,7 @@ objects:
       awsVMImages:
         regionAMIs:
         - region: us-east-1
-          ami: ami-10aed306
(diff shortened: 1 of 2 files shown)


csrwng:
@dgoodwin @staebler ptal
@staebler after deleting the cluster, the infra deprovisioning playbook is run but not the one for the machinesets. I have to delete the machinesets explicitly as well if I want to start the deprovisioning for those.

staebler:
@csrwng I noticed that this morning too. It looks like the garbage collector does not delete the machine set until all the finalizers are removed from the cluster. So it looks like we'll need to have the cluster controller delete the machine sets explicitly.

PR #108, opened 2018-02-26
Delete machinesets in cluster controller

csrwng:
if a DeletionTimestamp is set on a cluster, it ensures that all MachineSets belonging to that cluster get deleted (or already have a DeletionTimestamp).


csrwng:
@dgoodwin @staebler ptal
I tested this manually and things work great for me. The one resource that I don't see going away is pods. I would have thought that those would get deleted by their owner jobs, but that doesn't seem the case. However, that doesn't seem something we can control.

staebler:
@csrwng I'll create an issue for the pods not getting deleted. What I see running locally without AWS is that the pods created for the deprovision jobs do not have their jobs as owners.

csrwng:
Thx @staebler. I added a check for NotFound. However, it looks like an internal error is returned for AlreadyDeleted. I didn't think it'd be a good idea to add a less than robust check esp. if it's only a temporary condition.

PR #112, opened 2018-03-02
Accept name parameter in example cluster script

csrwng:
Creates a cluster with a different name if specified as a parameter to the create-cluster shell script

### DIFF
--- a/contrib/examples/create-cluster.sh
+++ b/contrib/examples/create-cluster.sh
@@ -20,12 +20,14 @@ oc get secret aws-credentials -n cluster-operator -o yaml | sed -E '/(namespace:
 oc get secret ssh-private-key -n cluster-operator -o yaml | sed -E '/(namespace:|annotations|last-applied-configuration:|selfLink|uid:|resourceVersion:)/d' | oc apply -f -
 oc get secret ssl-cert -n cluster-operator -o yaml | sed -E '/(namespace:|annotations|last-applied-configuration:|selfLink|uid:|resourceVersion:)/d' | oc apply -f -
 
+default_name="$(whoami)-cluster"
+name="${1:-$default_name}"
 
-if [ -e contrib/examples/$(whoami)-cluster.yaml ]
+if [ -e contrib/examples/${name}.yaml ]
 then
-	CLUSTER_YAML="$(whoami)-cluster.yaml"
+	CLUSTER_YAML="${name}.yaml"
 else
 	CLUSTER_YAML="cluster.yaml"
 fi
 
-oc process -f contrib/examples/${CLUSTER_YAML} -p CLUSTER_NAME=$(whoami)-cluster | oc apply -f -
+oc process -f contrib/examples/${CLUSTER_YAML} -p CLUSTER_NAME=${name} | oc apply -f -


contrib/examples/create-cluster.sh
```diff
@@ -20,12 +20,14 @@ oc get secret aws-credentials -n cluster-operator -o yaml | sed -E '/(namespace:
 oc get secret ssh-private-key -n cluster-operator -o yaml | sed -E '/(namespace:|annotations|last-applied-configuration:|selfLink|uid:|resourceVersion:)/d' | oc apply -f -
 oc get secret ssl-cert -n cluster-operator -o yaml | sed -E '/(namespace:|annotations|last-applied-configuration:|selfLink|uid:|resourceVersion:)/d' | oc apply -f -
 
+default_name="$(whoami)-cluster"
```
staebler: I would rather not check for ${whoam}-cluster.yaml. I think it makes the logic of which yaml file is used a bit convoluted where someone might unknowingly use the wrong file.

@joelddiaz How would you feel if you had to specify the name of the yaml file that you use instead of relying on a default ${whoami}-cluster.yaml?

contrib/examples/create-cluster.sh
```diff
@@ -20,12 +20,14 @@ oc get secret aws-credentials -n cluster-operator -o yaml | sed -E '/(namespace:
...
+name="${1:-$default_name}"
 
-if [ -e contrib/examples/$(whoami)-cluster.yaml ]
+if [ -e contrib/examples/${name}.yaml ]
```
staebler: I would rather that the user specified the file path rather than inferring the yaml extension and the location. I don't like that the name of the cluster created is inextricably tied to the name of the yaml file. I would rather continue to use `${whoami}-cluster` as the value passed to CLUSTER_NAME, where the user can choose to ignore that parameter in the custom template that they are using.

csrwng:
@staebler are you ok with the script taking a second parameter with the path to the yaml file but defaulting to the default (cluster.yaml) if not specified?

staebler:
@csrwng Is this what you are proposing?
First paramter specifies the name of the cluster.
Second parameter specifies the yaml file to use.

Is your motivation for this change to be able to more easily use a different name for the cluster? This may make @joelddiaz workflow more cumbersome. But if he's fine with it then I don't mind.

csrwng:
@staebler yes and yes :)

staebler:
@csrwng What if we used an externally defined variable for the cluster name instead of passing a parameter? So you could call it with `CLUSTER_NAME=myCluster contrib/example/create-cluster.sh`.

csrwng:
that works for me, if ok with you and @joelddiaz

contrib/examples/create-cluster.sh
joelddiaz: @staebler fine by me to have the user specify the yaml file

joelddiaz:
as long as there is a way to specify/override files/cluster-names without having to modify the files hosted under SCM, i'm okay

staebler:
@joelddiaz You can always call `oc apply` directly with whatever yaml file that you have. The script is only meant to be a convenience for running the example.

joelddiaz:
something as simple as "CLUSTER_YAML='<path>' ./contrib/examples/create-cluster.sh" would work great for my workflow (probably what i should have done in the first place...)

csrwng:
closing for now, will open a new one that uses env vars

PR #114, opened 2018-03-02
#98. Restart job pods on failure

### LINKED ISSUE #98: Make failing jobs less obstrusive
1) Do a backoff delay between failed jobs. Currently the next job starts right away after the previous job fails.
2) Limit the length of time to keep trying jobs that are failing. Currently failed jobs will be re-start indefinitely.

staebler:
Closes #98.

Instead of having a job recreate a pod when its container fails, let's change the pod spec so that its container is restarted when it fails. Since the pod will not enter the Failed phase, the job controller will not increase its backoff when the pod's container fails. This will allow the job controller to monitor the pod status in a timely fashion. As a bonus, etcd storage will be reduced since we won't be created an ever-increasing number of pods.

We should consider whether to skip the job completely and create and monitor pods directly instead.


dgoodwin:
How does the job ever reach failed state if the pod never goes to failed and continuously restarts?

Are we still lacking a way to delay retries a little without going exponential? Seems like we kind of want to re-try often for awhile but eventually slow down.

staebler:
@dgoodwin The job fails after 24 hours.

dgoodwin:
And are they immediate retries for that 24h?

staebler:
@dgoodwin Exponential backoff capped at 5 minutes.

dgoodwin:
This is really nice on the pod list. I'm a little worried how that 5 min cap will play out for total install time with things that will fail until the control plane is up, but maybe we shouldn't be starting those so early anyhow.

PR #115, opened 2018-03-08
WIP: first pass at machine set controller

joelddiaz:
high level notes:
1) ignoring master-type machinesets since those can't be easily scaled right now
2) doing dumb master object deletions
3) still need to integrate expectations (to keep from creating too many objects)

### DIFF
--- a/pkg/controller/machineset/machineset_controller.go
+++ b/pkg/controller/machineset/machineset_controller.go
@@ -18,11 +18,14 @@ package machineset
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	v1batch "k8s.io/api/batch/v1"
 	kapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	batchinformers "k8s.io/client-go/informers/batch/v1"
@@ -44,6 +47,8 @@ import (
 	"github.com/openshift/cluster-operator/pkg/kubernetes/pkg/util/metrics"
 	colog "github.com/openshift/cluster-operator/pkg/logging"
 )
+
+var controllerKind = clusteroperator.SchemeGroupVersion.WithKind("MachineSet")
 
 const (
 	// maxRetries is the number of times a service will be retried before it is dropped out of the queue.
@@ -75,6 +80,7 @@ var (
 // NewController returns a new *Controller.
 func NewController(
 	machineSetInformer informers.MachineSetInformer,
+	machineInformer informers.MachineInformer,
 	jobInformer batchinformers.JobInformer,
 	kubeClient kubeclientset.Interface,
 	clusteroperatorClient clusteroperatorclientset.Interface,
@@ -99,6 +105,8 @@ func NewController(
 		logger:     logger,
 	}
 
+	c.machineLister = machineInformer.Lister()
+
 	machineSetInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    c.addMachineSet,
 		UpdateFunc: c.updateMachineSet,
@@ -115,6 +123,7 @@ func NewController(
 	c.jobSync = controller.NewJobSync(c.jobControl, &jobSyncStrategy{controller: c}, true, logger)
 
 	c.syncHandler = c.jobSync.Sync
+	c.syncHandler2 = c.syncMachineSet
 	c.enqueueMachineSet = c.enqueue
 	c.ansibleGenerator = ansible.NewJobGenerator(ansibleImage, ansibleImagePullPolicy)
 
@@ -129,6 +138,9 @@ type Controller struct {
 	// To allow injection of syncMachineSet for testing.
 	syncHandler func(hKey string) error
 
+	// Handle machineset->machine object lifecycle
+	syncHandler2 func(hKey string) error
+
 	// To allow injection of mock ansible generator for testing
 	ansibleGenerator ansible.JobGenerator
 
@@ -138,6 +150,9 @@ type Controller struct {
 
 	// used for unit testing
 	enqueueMachineSet func(machineSet *clusteroperator.MachineSet)
+
+	// fetch machine objects
+	machineLister lister.MachineLister
 
 	// machineSetsLister is able to list/get machine sets and is populated by the shared informer passed to
 	// NewController.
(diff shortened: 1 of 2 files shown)


joelddiaz:
@dgoodwin @csrwng or @staebler i've made it far enough where some useful things happen with the machienset controller. would like a review to make sure i'm heading in the right direction and following conventions

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
...
+}
+
+func (c *Controller) manageMachines(machines []*clusteroperator.Machine, ms *clusteroperator.MachineSet) error {
+	/*machineSetKey, err := controller.KeyFunc(ms)
+	if err != nil {
+		// FIXME
+		return nil
+	}*/
+
+	machinePrefix := ms.ObjectMeta.Name + "-"
```
staebler: You can use `ms.Name` instead of `ms.ObjectMeta.Name`.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
...
+	// caculate how many to create/delete
+	if len(machines) > ms.Spec.Size {
+		deleteCount = len(machines) - ms.Spec.Size
+	}
+	if len(machines) < ms.Spec.Size {
+		createCount = ms.Spec.Size - len(machines)
+	}
+
+	for _, machine := range machines {
+		if ! strings.HasPrefix(machine.ObjectMeta.Name, machinePrefix) {
```
staebler: I don't think we need to worry about validating the machine name. We are using the owner reference as the determination of to which machine set the machine belongs. The name is just a convenience for the human user.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
...
+		c.logger.WithField("machineset", ms.ObjectMeta.Name).Debugf("JDIAZ deleting machine %v", m.Name)
+		c.client.ClusteroperatorV1alpha1().Machines(ms.ObjectMeta.Namespace).Delete(m.Name, &metav1.DeleteOptions{})
+		// catch error ^^^
+		deleteCount--
+	}
+
+	var machineToDelete *clusteroperator.Machine
+	c.logger.WithField("machineset", ms.ObjectMeta.Name).Debugf("JDIAZ deleting %v more machines", deleteCount)
+	for i := 0 ; i < deleteCount ; i++ {
+		machineToDelete, machineSetMachines = machineSetMachines[0], machineSetMachines[1:]
```
staebler: I don't see the point of overwriting the slice each time. Can't we simply use `machineSetMachines[i]` in each iteration?

`c.client.ClusteroperatorV1alpha1().Machines(ms.ObjectMeta.Namespace).Delete(machineSetMachines[i].Name, &metav1.DeleteOptions{})`

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
...
+		machineSetMachines = append(machineSetMachines, machine)
+
+	}
+
+	// Sync machines
+	machineConfig := buildNewMachine(ms, machinePrefix)
+	c.logger.WithField("machineset", ms.ObjectMeta.Name).Debugf("JDIAZ machine template: %+v", machineConfig)
+
+	for i := 0 ; i < createCount ; i++ {
+		_, err := c.client.ClusteroperatorV1alpha1().Machines(ms.ObjectMeta.Namespace).Create(machineConfig)
```
staebler: You are going to want to make these calls in a goroutine so that they can be performed in parallel, but that can be done later.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
...
+		c.client.ClusteroperatorV1alpha1().Machines(ms.ObjectMeta.Namespace).Delete(machineToDelete.Name, &metav1.DeleteOptions{})
+		// catch error ^^^
+	}
+
+	return nil
+}
+
+func currentMachineNeedsUpdate(machine *clusteroperator.Machine, ms *clusteroperator.MachineSet) bool {
+	// compare machine settings with machineset settings to see if we have an out-of-date machine
+	if machine.Spec.NodeType != ms.Spec.NodeType {
```
staebler: I would hope that this doesn't happen. That would imply that either a machine got attached to the wrong machineset or the node type of a machineset changed.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
...
+	return false
+}
+func buildNewMachine(machineSet *clusteroperator.MachineSet, machineNamePrefix string) *clusteroperator.Machine {
+	machineConfig := &clusteroperator.Machine{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName:	machineNamePrefix,
+			OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(machineSet, controllerKind)},
+		},
+		Spec: clusteroperator.MachineSpec{
+			NodeType: "Compute",
```
staebler: This should be the node type from machineset spec.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
...
+	}
+	//c.logger.WithField("machineset", name).Debugf("JDIAZ children machines: %v", filteredMachines)
+
+	var manageMachinesErr error
+
+	//if machineSetNeedsSync && ms.DeletionTimestamp == nil {
+	if true && ms.DeletionTimestamp == nil {
+		manageMachinesErr = c.manageMachines(filteredMachines, ms)
+	}
+
```
staebler: We'll want to add a status update to the machineset to track how many machines currently exist.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
...
+	ms, err := c.machineSetsLister.MachineSets(namespace).Get(name)
+	if errors.IsNotFound(err) {
+		c.logger.WithField("key", key).Errorf("JDIAZ couldn't find machineset %v", key)
+		return err
+	}
+	if err != nil {
+		c.logger.Errorf("JDIAZ problem fetching machineset %v", key)
+		return err
+	}
+	if ms.Spec.NodeType == "Master" {
```
staebler: This should be ignored in `addMachineSet`, `updateMachineSet`, and `deleteMachineSet`.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -230,12 +245,188 @@ func (c *Controller) processNextWorkItem() bool {
 	}
 	defer c.queue.Done(key)
 
+	// create/delete ansible jobs
 	err := c.syncHandler(key.(string))
 	c.handleErr(err, key)
 
+	// create/delete Machine objects
+	err = c.syncHandler2(key.(string))
```
staebler: We don't want to run two different sync handlers in the same controller. That is going to get very complicated. If we are going to continue to use ansible to create the master machines, then we should have a separate controller for master machine sets.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -99,6 +105,8 @@ func NewController(
 		logger:     logger,
 	}
 
+	c.machineLister = machineInformer.Lister()
```
staebler: We need to add an event handler for machines. That will be used by the expectations to track when a machine is added or deleted.

pkg/controller/machineset/machineset_controller.go
joelddiaz: is the nodetype field on a machineset expected to be mutable? if 'no', then perhaps there is no need for currentMachineNeedsUpdate() at all.

dgoodwin: Blocking that in validation seems fair to me, you shouldn't be able to change the node type of a machineset.

PR #116, opened 2018-03-09
Run each components playbook in parallel in a separate container

staebler:
Requires https://github.com/openshift/openshift-ansible/pull/7473.

Time to install from clean: ~12 minutes.

### DIFF
--- a/pkg/ansible/jobgenerator.go
+++ b/pkg/ansible/jobgenerator.go
@@ -18,6 +18,7 @@ package ansible
 
 import (
 	"path"
+	"strings"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


pkg/ansible/jobgenerator.go
```diff
@@ -231,3 +253,27 @@ func (r *jobGenerator) GeneratePlaybookJob(name string, hardware *clusteroperato
...
+		case '0' <= r && r <= '9':
+			return r
+		case r == '-':
+			return r
+		default:
+			return '-'
+		}
+	}
+	return strings.Map(replaceInvalidRunes, containerName)
+}
```
dgoodwin: Could we throw some simple unit tests at this func.

pkg/ansible/jobgenerator.go
```diff
@@ -208,6 +205,31 @@ func (r *jobGenerator) GeneratePlaybookJob(name string, hardware *clusteroperato
...
+			ImagePullPolicy: r.imagePullPolicy,
+			Env:             envForPlaybook,
+			VolumeMounts:    volumeMounts,
+		}
+	}
+
+	podSpec := kapi.PodSpec{
+		DNSPolicy:     kapi.DNSClusterFirst,
+		RestartPolicy: kapi.RestartPolicyNever,
+		Containers:    containers,
```
dgoodwin: This is clever, I don't think I'd have thought to run all the containers in one pod. 

This tested fine for me, but what would happen if one of the jobs failed? Is it going to re-try every container in the pod? Or just the one that failed.

staebler: If the pod is reused, it will retry just the containers that failed. In the current state of jobs, however, the pods are re-created rather than reused. If we take #114, then only the playbooks that failed will be retried.

dgoodwin: Forgot about that PR, I'll go get it in.

staebler:
There is one other thing that I forgot to mention. We need to decide on some mechanism to avoid hardcoding the component playbooks into cluster-operator. We don't want to have to release a new version of cluster-operator in order to add or remove the components that are installed. I will write up an issue about it with some possible solutions.

staebler:
Rebased with unit tests added.

dgoodwin:
Just realized I should not have merged that until the openshift-ansible patch merges, if anyone hits problems 2c012bb6a5aa02c63171e505b3900292a15aaadb should be the last good rev.

CC @abutcher @joelddiaz @csrwng

PR #117, opened 2018-03-12
Use an openshift-ansible image appropriate for a ClusterVersion

staebler:
* Removes the ansible image command-line arguments to controller-manager. The controllers will determine the ansible image to use based on the information in the ClusterVersion resource. The default is "openshift/origin-ansible:{TAG}", where {TAG} is `ClusterVersion.Spec.Version`. The default is overridden by specifying a value for `ClusterVersion.Spec.OpenshiftAnsibleImage`. The default pull policy is IfNotPresent. The default pull policy is overridden by specifying a value for `ClusterVersion.Spec.OpenshiftAnsibleImagePullPolicy`.
* Removes the `use_real_aws` parameter from the ansible deployment playbook. This is replaced with a `USE_REAL_AWS` variable in the create cluster script. Additionally, the ansible image and pull policy can be specified explicitly by setting the `ANSIBLE_IMAGE` and `ANSIBLE_IMAGE_PULL_POLICY` variables, respectively, when calling the create cluster script.

### DIFF
--- a/pkg/ansible/image.go
+++ b/pkg/ansible/image.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ansible
+
+import (
+	"fmt"
+
+	kapi "k8s.io/api/core/v1"
+
+	coapi "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
+)
+
+const (
+	defaultImageName = "openshift/origin-ansible"
+)
+
+// GetAnsibleImageForClusterVersion gets the openshift-ansible image and pull
+// policy to use for clusters that use the specified ClusterVersion.
+func GetAnsibleImageForClusterVersion(cv *coapi.ClusterVersion) (string, kapi.PullPolicy) {
+	image := fmt.Sprintf("%s:%s", defaultImageName, cv.Spec.Version)
+	if cv.Spec.OpenshiftAnsibleImage != nil {
+		image = *cv.Spec.OpenshiftAnsibleImage
+	}
+	pullPolicy := kapi.PullIfNotPresent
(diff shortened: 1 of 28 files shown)


pkg/apis/clusteroperator/validation/clusterversion.go
```diff
@@ -46,6 +47,24 @@ var validDeploymentTypeValues = func() []string {
...
+	kapi.PullAlways:       true,
+	kapi.PullNever:        true,
+	kapi.PullIfNotPresent: true,
+}
+
+// validPullPolicyValues is an array of every valid pull policy value.
+var validPullPolicyValues = func() []string {
+	validValues := make([]string, len(validPullPolicies))
+	i := 0
+	for dt := range validPullPolicies {
```
dgoodwin: Nit: "for i, dt" would work here I think.

pkg/ansible/image.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// GetAnsibleImageForClusterVersion gets the openshift-ansible image and pull
+// policy to use for clusters that use the specified ClusterVersion.
+func GetAnsibleImageForClusterVersion(cv *coapi.ClusterVersion) (string, kapi.PullPolicy) {
+	image := fmt.Sprintf("%s:%s", defaultImageName, cv.Spec.Version)
+	if cv.Spec.OpenshiftAnsibleImage != nil {
+		image = *cv.Spec.OpenshiftAnsibleImage
+	}
+	pullPolicy := kapi.PullIfNotPresent
```
dgoodwin: Any precedent for using this as the default pull policy in Kube or Origin? A little nervous it could lead to nodes not picking up images that had to be fixed or updated. Typical tag usage I wouldn't expect that to be very common, if we're using versions in tags that *should* only ever go out once, but I'm not positive.

pkg/apis/clusteroperator/validation/clusterversion.go
staebler: Oh man. I copied that code over from the `validDeploymentTypeValues` function without giving it more than a glance. I wonder who wrote that code. ;)

pkg/ansible/image.go
staebler: I don't know what the precedent is. I'll change it over to Always to be safe. We can change it later if we find out that IfNotPresent is appropriate.

pkg/apis/clusteroperator/validation/clusterversion.go
dgoodwin: Some golang newb no doubt.

staebler: That's right. `for i, dt` does not work because `validPullPolicies` is a map not an array.

dgoodwin: Ahh I see now.

PR #121, opened 2018-03-15
Switch to t2.medium for test cluster.

dgoodwin:
More commonly used for development, and seems to work fine for our purposes.


staebler:
Looks reasonable to me. @csrwng upped it to its current size, so he may have something to say about the degradation.

csrwng:
I switched it to match what we use in CI since that would be the first test of this. But no objection for our use.

PR #128, opened 2018-03-21
Add node-config controller to deploy daemonset for node config.

dgoodwin:


dgoodwin:
Should be operational in current form, this adds support for in general, re-running jobs at some interval with jitter.

@csrwng suggested yesterday just using a cron job, some notes on this approach:

  * We would want to avoid running this job for every machine set for every cluster all at once, right now this uses the jitter code from kube, but with a cronjob we would need to boil down to a cron string. This is a little more tricky but we could probably work something out, support every X hours, randomize the minute field.
  * We still want to run immediately once a machineset is ready. I could keep this part of the controller and just replace the bits to run on an interval.
  * Today we generate a vars/hosts configmap for every job execution, there would be nothing doing this for the cronjob. The controller would probably want to ensure the cronjob is setup on every machineset change, and maintain one static configmap for that cronjob which is also kept up to date so the vars are always correct. Both would need cleanup on machineset deletion.

Anyone have thoughts on which approach to pursue?

staebler:
I do not like the idea of using a cronjob for periodically running the ansible playbooks. I think it will be difficultand complex to get the coordination right between the controller and the cron jobs. I don't know what difficulties or limitations you are encountering getting resync within the controller to work, but I would be very surprised if the end solution for controller resyncing was more complex than a solution using cron jobs.

dgoodwin:
@csrwng felt working off timestamps for last run might be unreliable. This PR has a working implementation with controller resyncing, it was quite difficult to get the logic correct without creating infinite or multiple duplicate jobs, but it should be working as is now.

For CronJobs the last plan discussed on IRC was to run a job which would both deploy the daemonset initially, and configure a cron job on the CO cluster, which would then re-run this job. This requires a service account and role binding in each cluster namespace with permissions to create cronjobs, which will need to be maintained. We would need to recreate cronjob and configmap on each spec change and cleanup old. There may not be any coordination required with the controllers once initial deploy is done.

csrwng:
@dgoodwin @staebler A more reliable pattern (than tracking a date in status) is having a controller keep track (in a queue or expectations, or some other synced struct) when it needs to process the next resource. It seemed to me that cronjobs already do this, so why not leverage them?

staebler:
After giving it more thought, I am less concerned about the complexity of synchronizing cron jobs with the controller. I do still think that there are things  that  will  be easier to  handle if done within the controller. One  is handling intermittent and persistent failures after the cron job is established. For example, let's say that the periodic task fails on one run because of netwrok issues. Will the cron job be able to retry the task after a backoff delay? Or will the cron job have to wait for the next periodic run? I think the former is preferred. How will the status of the resource be updated to reflect that there is a problem?

pkg/controller/jobcontrol.go
```diff
@@ -214,10 +232,21 @@ func (c *jobControl) ControlJobs(
...
+			"generationJob": generationJob.Name,
+		}).Debugf("generationJob")
+
 		switch {
 		// A successful job exists for the current generation of the owner.
 		// The owner's status needs to be updated accordingly.
+		case buildReprocessingJob && (isSuccessful(generationJob) || isFailed(generationJob)):
+			logger.Debugf("successful generation job found but reprocess job requested, building new job")
+			_, err := c.createJob(ownerKey, owner, jobFactory, logger)
+			return JobControlCreatingJob, nil, err
```
dgoodwin: DoesOwnerNeedReprocessing and the subsequent block here were added  because lumping the reprocess request into the pre-existing DoesOwnerNeedProcessing, there's no way to tell when a job is successful the first time and we *don't* want to immediately re-run for config management. Both states seem necessary to get the proper behavior and not duplicate/infinite jobs.

csrwng:
> let's say that the periodic task fails on one run because of netwrok issues. Will the cron job be able to retry the task after a backoff delay?

All the cron job does is launch individual jobs with all the same execution characteristics as jobs we launch individually. So yes, the job controller will do its thing with retries and backoff.

pkg/controller/jobcontrol.go
```diff
@@ -209,13 +225,25 @@ func (c *jobControl) ControlJobs(
 	// jobs need to be deleted. A job to process the current generation of the
 	// owner cannot be started until all outstanding jobs are deleted.
 	if len(jobsToDelete) > 0 {
-		logger.Debugf("deleting %d old jobs", len(jobsToDelete))
 		err := c.deleteOldJobs(ownerKey, jobsToDelete, logger)
 		return JobControlDeletingJobs, nil, err
 	}
 
+	// The owner needs a job to process its current generaton. A job needs to
+	// be created to do this processing.
```
csrwng: Seems that this comment still applies to the block starting on line 264

pkg/controller/jobsync_test.go
```diff
@@ -207,14 +207,15 @@ func TestJobSyncWithErrorControllingJobs(t *testing.T) {
 
 	owner := &metav1.ObjectMeta{}
 	needsProcessing := true
+	needsReprocessing := false
```
csrwng: Not sure that a separate variable is needed here since you can just pass false below. If for doc purposes, maybe it should be renamed what its value means ... doesNotNeedReprocessing

pkg/controller/nodeconfig/nodeconfig_controller.go
```diff
@@ -0,0 +1,462 @@
...
+	lister "github.com/openshift/cluster-operator/pkg/client/listers_generated/clusteroperator/v1alpha1"
+	"github.com/openshift/cluster-operator/pkg/controller"
+	"github.com/openshift/cluster-operator/pkg/kubernetes/pkg/util/metrics"
+	"github.com/openshift/cluster-operator/pkg/logging"
+)
+
+const (
+	// maxRetries is the number of times a service will be retried before it is dropped out of the queue.
+	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the
+	// sequence of delays between successive queuings of a service.
```
csrwng: s/service/machine set/

pkg/controller/nodeconfig/nodeconfig_controller.go
```diff
@@ -0,0 +1,462 @@
...
+		return false
+	}
+
+	msLog := logging.WithMachineSet(s.controller.logger, machineSet).WithField("func", "DoesOwnerNeedReProcessing")
+
+	// Process if it has been more than the configurable interval of time since we last ran:
+	installingCondition := controller.FindMachineSetCondition(machineSet, clusteroperator.MachineSetNodeConfigInstalling)
+	installedCondition := controller.FindMachineSetCondition(machineSet, clusteroperator.MachineSetNodeConfigInstalled)
+	if installedCondition == nil || installingCondition == nil {
+		msLog.Warn("machine set has node config installed generation but missing relevant conditions")
```
csrwng: why? it could be that it's just not ready for processing yet (ie. master not installed)

pkg/controller/nodeconfig/nodeconfig_controller.go
```diff
@@ -0,0 +1,462 @@
...
+		return false
+	}
+
+	if installedCondition.Status == kapi.ConditionFalse {
+		msLog.Warn("node config installed condition was false")
+		return true
+	}
+
+	// Check how long it's been since we last configured the node config daemonset:
+	lastInstalled := installedCondition.LastProbeTime.Time
```
csrwng: I think this is abusing the LastProbeTime field, since it doesn't (or shouldn't) mean that's when you last ran stuff. I would think either of the following approaches would be better:
1) Add another field to the status to indicate when a ConfigureNode was last run
2) Keep track within the controller when stuff last ran

pkg/controller/jobcontrol.go
```diff
@@ -209,13 +225,25 @@ func (c *jobControl) ControlJobs(
 	// jobs need to be deleted. A job to process the current generation of the
 	// owner cannot be started until all outstanding jobs are deleted.
 	if len(jobsToDelete) > 0 {
-		logger.Debugf("deleting %d old jobs", len(jobsToDelete))
 		err := c.deleteOldJobs(ownerKey, jobsToDelete, logger)
```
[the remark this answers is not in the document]
dgoodwin: The deleteOldJobs method does similar logging.

pkg/controller/nodeconfig/nodeconfig_controller.go
dgoodwin: Good ideas, I think I'm leaning towards field, I like the idea of us being able to see if on the object when we need to debug something.

dgoodwin:
Updated again to get the job repeats working, now stores the last successful job time. We still need to know when we're seeing a successful job for the first time, otherwise it just looks like we need to launch a new one as we're over the interval and the success never gets recorded, however this is easy just by comparing the latest successful job completion time to the status timestamp.

Job cleanup is a little iffy here, I'm cleaning up jobs but pods are leaking like crazy supposedly because of a kube bug. I'm also not 100% clear if my delete jobs is impacting the actual launching of a new one given how it returns early.

Also I still need to address the rest of Cesar's feedback.

dgoodwin:
I think this is ready for full review now, the delete jobs behavior looks to be working quite well, when we see a new job is required it's created, another sync happens soon after and sees the duplicate jobs, does a couple more syncs where expectations are not met (waiting for jobs to delete), after which the new job is running and we wait for it to complete. 

I'd drop WIP but there's an issue with where to add the node config playbook I'm calling. We can't get it into origin-ansible 3.9, 3.10 is broken if you run against 3.9 AMIs, so I think I need to wait until @joelddiaz PR merges with some support for layering our own Ansible in on top of origin-ansible. We'll likely need to build an image for each major version we support. (v3.9, v3.10, etc)

dgoodwin:
Updated now to layer in the node config playbook based on @joelddiaz work. This exposed other problems with the origin-ansible image, our broken out component playbooks were not being included in their build. We were already thinking about moving all our playbooks to our repo anyhow so this seemed like a good time.

With these changes the node config daemonset will deploy correctly, and we're layering in all our ansible from our own repo. 

Ready for full review or merge.

contrib/examples/create-cluster.sh
```diff
@@ -35,8 +35,8 @@ then
 	: ${ANSIBLE_IMAGE:="fake-openshift-ansible:canary"}
 	: ${ANSIBLE_IMAGE_PULL_POLICY:="Never"}
 else
-	: ${ANSIBLE_IMAGE:="openshift/origin-ansible:v3.9"}
-	: ${ANSIBLE_IMAGE_PULL_POLICY:="Always"}
+	: ${ANSIBLE_IMAGE:="cluster-operator-ansible:canary"}
```
dgoodwin: Now using our image for everything.

PR #129, opened 2018-03-22
Rename all the things

dgoodwin:
Some convenience and consistency renames for pods and jobs. Details in commits.


pkg/controller/machineset/machineset_controller.go
```diff
@@ -59,6 +59,10 @@ const (
 	// Ansible execution.
 	jobPrefix = "provision-machineset-"
 
+	// deprovisionJobPrefixPrefix applies a prefix to the standard job prefix when we're
+	// running a deprovision playbook. (sorry)
+	deprovisionJobPrefixPrefix = "de"
```
staebler: The JobControl is created with a specific job prefix which it looks for. If jobs are created with a different prefix, then the JobControl will not be able to manage them.

staebler:
I think it is a mistake to rename all the cluster-operator-blah names to co-blah. As a user, it will be difficult to figure out what co-blah resources are unless you are already very familiar with cluster operator. I can appreciate that it is cumbersome as a developer to have to type cluster-operator-blah often, but that is not a good excuse to obsfucate things for the end user.

staebler:
If there are strong feelings or outside pressure that cluster-operator-blah is too long, could we compromise on clustop-blah. At least clustop is something that has a chance of being found in a search whereas co will be difficult to find.

staebler:
Many similar changes to the job names are done in #120, as well.

pkg/controller/machineset/machineset_controller.go
dgoodwin: Seemed to be working in practice and all tests but must be something not covered in testing.

dgoodwin:
Don't want to invest a bunch of time in debating this, so closed for now.

PR #130, opened 2018-04-03
use DeploymentType from ClusterVersion

joelddiaz:
rather than hardcode 'origin' use the DeploymentType defined in the clusterVersion

this means we need to provide the clusterVersionSpec so that we can appropriately generate clusterWideVars. update GenerateClusterVars() and GenerateClusterWideVars() (and their callers) to this effect


pkg/ansible/generate.go
```diff
@@ -85,7 +85,7 @@ ansible_ssh_user: [[ .SSHUser ]]
 # Ensure these variables are set for bootstrap
 ################################################################################
 # Deployment type must be specified. ('origin' or 'openshift-enterprise')
-openshift_deployment_type: origin
+openshift_deployment_type: [[ .DeploymentType ]]
```
dgoodwin: I'm rather surprised this was still in there, good catch.

pkg/ansible/generate_test.go
```diff
@@ -99,12 +99,14 @@ func TestGenerateClusterVars(t *testing.T) {
...
 		name             string
 		cluster          *coapi.Cluster
+		clusterVersion   *coapi.ClusterVersion
 		shouldInclude    []string
 		shouldNotInclude []string
 	}{
 		{
 			name:    "cluster",
 			cluster: testCluster(),
+			clusterVersion: testClusterVersion(),
```
dgoodwin: Just need to update these tests with additional string to look for the deployment type you're passing in. (shouldInclude array)

PR #131, opened 2018-04-05
Add a controller to periodically deploy the kube cluster API controllers.

dgoodwin:
Still need to address a bug where the playbook keeps recreation the controllers each run but this should be operational for now.


dgoodwin:
Updated with some review touchups and a rename to "deployclusterapi" controller per conversation last week, I had that change in another branch and brought it over.

PR #134, opened 2018-04-05
Ignore customized local cluster yaml

csrwng:
Ignore local cluster.yaml in the form of $(whoami)-cluster.yaml


csrwng:
@dgoodwin ptal

PR #137, opened 2018-04-05
AWS machine controller for cluster API

csrwng:
Adds an AWS machine controller to use with cluster API

### DIFF
--- a/pkg/clusterapi/aws/actuator.go
+++ b/pkg/clusterapi/aws/actuator.go
@@ -1,0 +1,486 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package aws
+
+import (
+	"encoding/base64"
+	"fmt"
+	"io/ioutil"
+
+	log "github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/kubernetes"
+	clusterv1 "k8s.io/kube-deploy/cluster-api/pkg/apis/cluster/v1alpha1"
+	clusterclient "k8s.io/kube-deploy/cluster-api/pkg/client/clientset_generated/clientset"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	coapi "github.com/openshift/cluster-operator/pkg/api"
+	cov1 "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
+)
+
+const (
+	// Path to bootstrap kubeconfig. This needs to be mounted to the controller pod
+	// as a secret when running this controller.
+	bootstrapKubeConfig = "/etc/origin/master/bootstrap.kubeconfig"
+
+	// Hardcode IAM role for infra/compute for now
+	defaultIAMRole = "openshift_node_describe_instances"
+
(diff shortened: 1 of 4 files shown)


csrwng:
Still needs support for delete, check existing

pkg/clusterapi/aws/actuator.go
```diff
@@ -0,0 +1,403 @@
...
+	return nil
+}
+
+// Checks if the machine currently exists.
+func (a *AWSActuator) Exists(*clusterv1.Machine) (bool, error) {
+	return false, nil
+}
+
+// Helper function to create SVC
+func (a *AWSActuator) createSVC(region string) (*ec2.EC2, error) {
```
dgoodwin: SVC here made me think kube service, is there a better term we could use for this?

pkg/clusterapi/aws/actuator.go
```diff
@@ -0,0 +1,403 @@
...
+	if coClusterVersion.Spec.VMImages.AWSImages == nil {
+		return nil, fmt.Errorf("Cluster version does not contain AWS images")
+	}
+
+	// Get AMI to use
+	amiName := amiForRegion(coClusterVersion.Spec.VMImages.AWSImages.RegionAMIs, region)
+	if len(amiName) == 0 {
+		return nil, fmt.Errorf("Cannot determine AMI name from cluster version %#v and region %s", coClusterVersion, region)
+	}
+	imageIds := []*string{aws.String(amiName)}
```
dgoodwin: In general we should get some good logging in here, doesn't have to be right away though as I know this is very preliminary.

pkg/clusterapi/aws/actuator.go
```diff
@@ -0,0 +1,403 @@
...
+
+	// Determine security groups
+	var groupName, groupNameK8s string
+	if coMachineSet.Spec.Infra {
+		groupName = vpcName + "_infra"
+		groupNameK8s = vpcName + "_infra_k8s"
+	} else {
+		groupName = vpcName + "_compute"
+		groupNameK8s = vpcName + "_compute_k8s"
+	}
```
dgoodwin: We should give some thought to this for the future, been wondering if these implied names should be explicit in the API or not.

cmd/aws-machine-controller/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+	kubeClient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		glog.Fatalf("Could not create kubernetes client to talk to the apiserver: %v", err)
+	}
+
+	actuator := aws.NewActuator(kubeClient, client)
+	shutdown := make(chan struct{})
+	si := sharedinformers.NewSharedInformers(config, shutdown)
+	c := machine.NewMachineController(config, si, actuator)
```
dgoodwin: This is cool, just need to drop in our actuator? I assume you were saying this actuator is probably not something we'll be able to contribute upstream because of our requirements, but we'll keep using the controller from their side?

csrwng:
@dgoodwin for now we'll get the image from quay.io/openshift/kubernetes-cluster-api:latest (as referenced by @joelddiaz PR). We should setup a CI process to regularly publish the image, ideally built from code vendored in this repository. For now, I've manually pushed the image.

cmd/aws-machine-controller/main.go
csrwng: right, we're simply reusing the vendored machine controller that syncs machine resources with whatever the actuator does

pkg/clusterapi/aws/actuator.go
csrwng: Will fix

pkg/clusterapi/aws/actuator.go
```diff
@@ -0,0 +1,403 @@
...
+	if err != nil {
+		return err
+	}
+	machineCopy := machine.DeepCopy()
+	machineCopy.Annotations[instanceIDAnnotation] = *(result.Instances[0].InstanceId)
+	_, err = a.clusterClient.ClusterV1alpha1().Machines(machineCopy.Namespace).Update(machineCopy)
+	return err
+}
+
+func (a *AWSActuator) CreateMachine(cluster *clusterv1.Cluster, machine *clusterv1.Machine) (*ec2.Reservation, error) {
```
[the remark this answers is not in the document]
csrwng: Yeah, I made it public so I could test creating a machine directly (outside of the controller). I will include that test code in the contrib directory so we can use it for testing.

csrwng:
Addressed comments,  added test command, and completed actuator implementation

PR #138, opened 2018-04-06
idempotentize cluster-api-template

joelddiaz:
oc_process template has a particular way of determining whether an object in a template already exists on the cluster. just use 'oc process | oc apply' instead of the oc_process module.
    
this should keep the pods from restarting whenever the cluster-api installation is re-run, since 'oc apply' will know not to modify the already existing 'Deployment' objects.

### DIFF
--- a/build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
+++ b/build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
@@ -76,19 +76,20 @@
         cluster_api_key: "{{ clusterapi_key.content }}"
         cluster_api_ca_bundle: "{{ clusterapi_crt.content }}"
 
-  - name: process template
-    oc_process:
-      namespace: "{{ cluster_api_namespace }}"
-      state: present
-      create: true
-      content: "{{ template.content | b64decode }}"
-      params:
-         CLUSTER_API_NAMESPACE: "{{ cluster_api_namespace }}"
-         SERVING_CA: "{{ cluster_api_ca_bundle }}"
-         SERVING_CERT: "{{ cluster_api_cert }}"
-         SERVING_KEY: "{{ cluster_api_key }}"
-    run_once: true
-    register: template_out
+  - name: create temp file to hold the template
+    tempfile:
+      state: file
+    register: temp_file
 
-  - debug:
-      var: template_out
+  - name: copy template over
+    copy:
+      src: files/cluster-api-template.yaml
+      dest: "{{ temp_file.path }}"
+
+  - name: process template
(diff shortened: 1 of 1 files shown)


build/cluster-operator-ansible/playbooks/cluster-api-prep/files/cluster-api-template.yaml
```diff
@@ -33,73 +33,73 @@ objects:
...
+    sessionAffinity: None
+    type: ClusterIP
 
-- apiVersion: apps/v1beta1
+- apiVersion: extensions/v1beta1
   kind: Deployment
   metadata:
-    name: cluster-api-apiserver
-    namespace: ${CLUSTER_API_NAMESPACE}
+    generation: 1
```
dgoodwin: Does setting this have something to do with getting oc_process to not recreate? 

This makes me a little nervous, doesn't seem like something we should be setting and I'm also wondering if changing the yaml would get applied if you didn't bump the generation?

kubectl apply has the logic we want and sadly it's not server side yet, so anything that's doing it's own thing like oc process makes me a little nervous as there's a lot of complexity in it. Could we do what we do for our dev template for now? https://github.com/openshift/cluster-operator/blob/master/contrib/ansible/deploy-devel.yaml#L58

In future this will hopefully be k8s_raw + server side apply logic.

build/cluster-operator-ansible/playbooks/cluster-api-prep/files/cluster-api-template.yaml
```diff
@@ -121,104 +125,132 @@ objects:
...
+              cpu: 100m
+              memory: 30Mi
+            requests:
+              cpu: 100m
+              memory: 20Mi
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+          volumeMounts:
+          - mountPath: /etcd-data-dir
+            name: etcd-data-dir
```
dgoodwin: Was this file updated with the results of an actual object get? Looks like a lot of defaults might have been added in. In this case I'd also be inclined to stick to just defining what we care about and letting the server apply it's defaults.

joelddiaz:
@dgoodwin this can be re-reviewed. i moved things away from the oc_process module and am just doing an 'oc process | oc apply' workflow.

dgoodwin:
Thanks! FWIW I was reading yesterday that the effort to move apply logic server side has actually kicked off, once that's in place hopefully we can get support in k8s_raw and then we'll be in better shape for stuff like this.

build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
```diff
@@ -76,19 +76,20 @@
...
 
+  - name: create temp file to hold the template
+    tempfile:
+      state: file
+    register: temp_file
+
+  - name: copy template over
+    copy:
+      src: files/cluster-api-template.yaml
+      dest: "{{ temp_file.path }}"
```
dgoodwin: For future consideration we could leave this as a persistent file on the host, then we would accurately see if something changed of not as a result of the copy and could do changed_when: false on the oc process command.

PR #139, opened 2018-04-06
Vendor changes required for cluster API machine controller

csrwng:
Adjusts some prereqs and removes embedded vendor directories


csrwng:
@dgoodwin ptal

PR #140, opened 2018-04-10
Enable launch_instances policy for masters

csrwng:
This will result in masters having an IAM policy that
will allow them to create additional machine instances for
the cluster.


csrwng:
@dgoodwin @abutcher ptal
This will have no effect until https://github.com/openshift/openshift-ansible/pull/7680 merges

PR #143, opened 2018-04-12
Export cluster API resources

csrwng:
Adds a command that creates Cluster API resources based on an existing cluster operator Cluster and MachineSets


csrwng:
@dgoodwin ptal
Export is working now

dgoodwin:
This will be super helpful, thanks.

/lgtm

PR #144, opened 2018-04-12
Secret creation serviceaccount

joelddiaz:
update clusterrole permissions to allow creating/querying serviceaccounts and rolebindings

add steps to master_controller to create a serviceaccount that can create secrets (to eventually hold a target cluster's kubeconfig)

update jobgenerator to allow specifying a serviceaccount that should be attached to the pod/job

### DIFF
--- a/contrib/examples/deploy.yaml
+++ b/contrib/examples/deploy.yaml
@@ -91,10 +91,14 @@ objects:
     verbs:     ["create", "get", "list", "watch", "update", "patch", "delete"]
   # allow operations on required resources in any namespace a cluster is created
   - apiGroups:     [""]
-    resources:     ["configmaps", "pods", "secrets"]
+    resources:     ["configmaps", "pods", "secrets", "serviceaccounts"]
     verbs: ["*"]
   - apiGroups:     ["batch"]
     resources:     ["jobs"]
+    verbs: ["*"]
+  # allow reading and creating rolebindings
+  - apiGroups: ["rbac.authorization.k8s.io"]
+    resources: ["rolebindings"]
     verbs: ["*"]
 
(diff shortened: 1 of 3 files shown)


contrib/examples/deploy.yaml
```diff
@@ -384,6 +388,18 @@ objects:
...
+# from a managed cluster
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: ClusterRole
+  metadata:
+    name: "clusteroperator.openshift.io:master-controller"
+  rules:
+  # create secrets with kubeconfig data
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs:     ["create", "update"]
```
csrwng: I would think that you need get/list here as well, since you'd need to update the secret if it already exists (or even delete it, and recreate it)

joelddiaz: definitely. thanks

pkg/controller/master/master_controller.go
```diff
@@ -355,6 +371,64 @@ func (s *jobSyncStrategy) GetJobFactory(owner metav1.Object, deleting bool) (con
...
+				Kind: "ClusterRole",
+			},
+		}
+
+		currentRB, err = s.controller.kubeClient.Rbac().RoleBindings(clusterNamespace).Create(rb)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't create rolebinding to clusterrole")
+		}
+	}
+	s.controller.logger.Debugf("rolbinding to serviceaccount: %+v", currentRB)
```
csrwng: s/rolbinding/rolebinding/

joelddiaz: fixed

PR #145, opened 2018-04-12
Fix AWS actuator, double-encoding of bootstrap kubeconfig

csrwng:
Fix to some issues found during testing of the cluster API with the aws-machine-controller. Also adds additional debugging.


csrwng:
@dgoodwin ptal

PR #146, opened 2018-04-12
Enable auto approval of node CSRs

csrwng:
Adds inventory variable to enable auto-approve of nodes


csrwng:
@dgoodwin @joelddiaz ptal

PR #147, opened 2018-04-16
First draft of syncmachineset controller.

dgoodwin:
Because we don't yet have kubeconfig secrets this controller is disabled
and partially completed, currently only processing master machinesets
and covered by unit tests.

Putting up for review as I might be setting aside temporarily and would like to get it in before it bitrots, as well as while it's a more managable size.


dgoodwin:
Closing, being taken over by @abutcher

PR #149, opened 2018-04-16
Generate certs if secrets do not exist.

dgoodwin:
If either the apiserver or cluster cert secrets do not exist, generate a
new local set of certs, and store as secrets.

etcd in the cluster itself will serve as the shared storage for these
across the team for a deployment.

### DIFF
--- a/contrib/ansible/deploy-devel.yaml
+++ b/contrib/ansible/deploy-devel.yaml
@@ -4,9 +4,14 @@
   connection: local
   gather_facts: no
   vars:
-    aws_section: default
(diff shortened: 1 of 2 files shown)


csrwng:
@dgoodwin looks good to me
What was the motivation for this change? Seems like we almost do the same thing

dgoodwin:
@csrwng This works us towards an ansible playbook that any of us can run to update the deployment in the CI cluster. Previously anyone who ran without a local cert would trigger a full cert redeploy for all. I wasn't sure how we'd keep the certs constant but it feels like we can just use their existence in etcd as the means to share them amongst the team.

csrwng:
@dgoodwin ack

PR #151, opened 2018-04-17
Bump test cluster instance type to t2.xlarge.

dgoodwin:
We're really struggling with performance and stability on t2.medium.


dgoodwin:
This will take us from 2 vcpu 4 GB RAM to 4 vcpu 16 GB RAM.

PR #152, opened 2018-04-17
save kubeconfig after install

joelddiaz:
introduce playbooks/cluster-operator/aws/install_masters.yml. this will call the original playbook/aws/openshift-cluster/install.yml and then save the kubeconfig from one of the masters into a secret for custer operator

update the master controller installation playbook to use the new entrypoint

note: this now means that any and all clusterversion objects will need to reference cluster-operator-ansible as the container image

### DIFF
--- a/build/cluster-operator-ansible/playbooks/cluster-operator/aws/install_masters.yml
+++ b/build/cluster-operator-ansible/playbooks/cluster-operator/aws/install_masters.yml
@@ -1,0 +1,54 @@
+---
+# call openshift-ansible's master provisioning
+- import_playbook: ../../aws/openshift-cluster/install.yml
+
+# The below piece of ansible would be needed if we wanted to only
+# do the save-kubeconfig-from-target-cluster steps that follow.
+# This playbook can one day be extended to handle a workflow of
+# just-get-me-an-updated-kubeconfig by enabling/conditionalizing
+# the commented out piece below.
+#- name: setup master ansible hosts
+#  hosts: localhost
+#  gather_facts: false
+#  tasks:
+#  - name: setup the master node group
+#    import_role:
+#      name: openshift_aws
+#      tasks_from: setup_master_group.yml
+
+- name: save kubeconfig into secret
+  hosts: masters[0]
+  gather_facts: false
+  remote_user: root
+  tasks:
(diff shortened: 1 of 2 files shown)


build/cluster-operator-ansible/playbooks/cluster-operator/aws/install_masters.yml
```diff
@@ -0,0 +1,49 @@
...
+#      tasks_from: setup_master_group.yml
+
+- name: save kubeconfig into secret
+  hosts: masters[0]
+  gather_facts: false
+  remote_user: root
+  tasks:
+  - name: slurp kubeconfig
+    slurp:
+      src: /root/.kube/config
```
csrwng: I'd prefer /etc/origin/master/admin.kubeconfig

build/cluster-operator-ansible/playbooks/cluster-operator/aws/install_masters.yml
```diff
@@ -0,0 +1,49 @@
+---
+# call openshift-ansible's master provisioning
+- import_playbook: ../../aws/openshift-cluster/install.yml
+
+#- name: setup master ansible hosts
```
csrwng: if not needed, I'd delete this commented section

build/cluster-operator-ansible/playbooks/cluster-operator/aws/install_masters.yml
joelddiaz: changed

build/cluster-operator-ansible/playbooks/cluster-operator/aws/install_masters.yml
joelddiaz: the commented out piece is useful for the day when we want a standalone fetch-kubeconfig-from-target-cluster, so i've added comments to make clear why we would want to run this piece one day

build/cluster-operator-ansible/playbooks/cluster-operator/aws/install_masters.yml
joelddiaz: i run ocp with cluster operator (just to keep us honest :) so /etc/origin/master/admin.kubeconfig does work

PR #155, opened 2018-04-17
Fix fake-ansible image missing our playbooks.

dgoodwin:
Right now the fake-ansible image run if you don't specify to USE_REAL_AWS is failing for any job that runs one of our custom playbooks, the only thing this container does is verify the playbook exists, which it wouldn't as it was based off origin-ansible.

Reordered image build to follow the cluster-operator-ansible image, and assumed we can safely build off canary if we're building this image. (as it wouldn't be a part of any official build)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -313,7 +313,7 @@ clean-coverage:
 
 # Building Docker Images for our executables
 ############################################
-images: cluster-operator-image fake-openshift-ansible-image playbook-mock-image cluster-operator-ansible-image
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -313,7 +313,7 @@ clean-coverage:
 
 # Building Docker Images for our executables
 ############################################
-images: cluster-operator-image fake-openshift-ansible-image playbook-mock-image cluster-operator-ansible-image
```
csrwng: Add cluster-operator-ansible-image as a dependency to the fake-openshift-ansible-image target

dgoodwin: Added.

PR #157, opened 2018-04-17
Override openshift_hosted_router_edits to enable ROUTER_USE_PROXY_PROTOCOL.

abutcher:
I'd prefer the env vars to be a parameter to the `oc_adm_router` module but this is how we would add env vars today.


dgoodwin:
Have to see where things land when there's a router operator. This looks ok for now, could you just comment heavily what's going on there and why.

PR #158, opened 2018-04-18
Create etcd pvc in cluster operator namespace

csrwng:
Deploying didn't work for me because the pvc got created in the current namespace instead of the cluster-operator-namespace


csrwng:
@dgoodwin ptal

dgoodwin:
Yup already hit this and had to fix it locally.

PR #159, opened 2018-04-18
building various cluster-operator-ansible images

joelddiaz:
change target cluster-operator-ansible-images to build cluster-operator-ansible images that can be used to install 3.9 and 3.10 OpenShift clusters (still allow building canary/custom).
    
get cluster-creation to use new versioned c-o-a images

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ BUILD_DIR     ?= build
 COVERAGE      ?= $(CURDIR)/coverage.html
 CLUSTERAPI_BIN = $(BINDIR)/cluster-api
 CLUSTER_OPERATOR_PKG = github.com/openshift/cluster-operator
+CLUSTER_OPERATOR_ANSIBLE_IMAGE_NAME = cluster-operator-ansible
 TOP_SRC_DIRS   = cmd pkg
 SRC_DIRS       = $(shell sh -c "find $(TOP_SRC_DIRS) -name \\*.go \
                    -exec dirname {} \\; | sort | uniq")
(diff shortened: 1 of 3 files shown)


dgoodwin:
Do you think there's a way we can get it integrated into the Makefile similar to today? A make target to build images seems important for a lot of our targets, and it would be really slick if we could continue to make images and eventually make push-images to publish to quay.

Makefile
```diff
@@ -345,9 +344,24 @@ endif
 CLUSTER_OP_ANSIBLE_REPO   ?= https://github.com/openshift/openshift-ansible.git
 CLUSTER_OP_ANSIBLE_BRANCH ?= master
 
+define build-cluster-operator-ansible-image #(repo, branch, imagename, tag)
+	echo "repo $1"
+	echo "branch $2"
+	echo "image $3:$4"
+	docker build -t "$3:$4" --build-arg=CO_ANSIBLE_URL=$1 --build-arg=CO_ANSIBLE_BRANCH=$2 build/cluster-operator-ansible
```
dgoodwin: These pre-existing vars could possibly use a rename for clarity, this isn't the location of CO ansible, it's the openshift-ansible ansible, maybe OA_ANSIBLE_URL and BRANCH?

Makefile
```diff
@@ -345,9 +344,24 @@ endif
...
+	$(call build-cluster-operator-ansible-image,$(CLUSTER_OP_ANSIBLE_REPO),"master",$(CLUSTER_OPERATOR_ANSIBLE_IMAGE_NAME),$(VERSION))
+	docker tag $(CLUSTER_OPERATOR_ANSIBLE_IMAGE_NAME):$(VERSION) $(CLUSTER_OPERATOR_ANSIBLE_IMAGE_NAME):$(MUTABLE_TAG)
+
+
+cluster-operator-ansible-versions:
+	# build v3.9 on openshift-ansible:release-3.9
+	$(call build-cluster-operator-ansible-image,$(CLUSTER_OP_ANSIBLE_REPO),"release-3.9",$(CLUSTER_OPERATOR_ANSIBLE_IMAGE_NAME),"v3.9")
+
+	# build v3.10 on openshift-ansible:master
+	$(call build-cluster-operator-ansible-image,$(CLUSTER_OP_ANSIBLE_REPO),"master",$(CLUSTER_OPERATOR_ANSIBLE_IMAGE_NAME),"v3.10")
```
dgoodwin: Getting closer, I would kind of like any normal make images invocation to result in building both of these, either cluster-operator-ansible:v3.9 + v3.10, or v3.0-canary v3.10-canary. I'm hoping to have a default devel deploy have prepopulated ClusterVersions for both with a PR I'm working on now.

So in theory just a Make target cluster-operator-ansible-images that replaces these.

Would this be reasonable to do or any gotchas I'm not thinking of?

dgoodwin:
Looks great, just need to wire it up so we're operational creating a cluster, suspect you just need to change the tag for the cluster versions in create-cluster.sh.

dgoodwin:
I just saw canary is still in the mix, so the repo should be fully working if we merge?

joelddiaz:
@dgoodwin i thought i'd leave canary in there so we can easily custom builds for testing (by setting/overriding OA_ANSIBLE_URL and/or OA_ANSIBLE_BRANCH)

also updated the cluster.yaml template and create-cluster.sh to use v3.10 by default (which just happens to match 'canary' right now)

PR #160, opened 2018-04-18
More changes to support deployment to CI cluster

dgoodwin:
Details in individual commits.


contrib/examples/cluster-operator-roles-template.yaml
```diff
@@ -0,0 +1,120 @@
...
+    name: cluster-operator-controller-manager
+    namespace: ${CLUSTER_OPERATOR_NAMESPACE}
+
+# API Server gets the auth-delegator role to delegate auth decisions to
+# the core API Server
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: ClusterRoleBinding
+  metadata:
+    name: "clusteroperator.openshift.io:apiserver-auth-delegator"
+    namespace: ${CLUSTER_OPERATOR_NAMESPACE}
```
csrwng: no namespace is needed

contrib/examples/cluster-operator-roles-template.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+# API Server gets the ability to read authentication. This allows it to
+# read the specific configmap that has the requestheader-* entries to
+# enable api aggregation
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: RoleBinding
+  metadata:
+    name: "clusteroperator.openshift.io:apiserver-authentication-reader"
+    namespace: kube-system
```
csrwng: needs to be parameterized

contrib/examples/cluster-operator-template.yaml
```diff
@@ -338,7 +245,7 @@ objects:
   kind: APIService
   metadata:
     name: v1alpha1.clusteroperator.openshift.io
-    namespace: ${KUBE_SYSTEM_NAMESPACE}
+    namespace: kube-system
```
csrwng: needs to be parameterized

dgoodwin: The parameter was commented with "don't change this", it seems like we should not treat as a parameter at that point. Does this actually need to change?

csrwng:
@dgoodwin I think the changes are mostly good. However, I was thinking that it may make more sense to create the secrets (certs, ssh keys, etc) when creating a cluster, and always generate them based on the name of the cluster being created. This would mean that we would need to convert the shell script to create a cluster into its own playbook, but it would solve the issue of re-using the same secret when creating multiple clusters. The other awkward thing is that the cluster version is created globally in the cluster-operator namespace, and once created, it cannot be changed. This makes it difficult to have a cluster resource that uses the real ansible playbook vs one that uses the fake one. I'd prefer that we create a clusterversion that corresponds to the cluster being created.

contrib/examples/cluster-operator-template.yaml
csrwng: No, it doesn't need to change, but it needs to be a parameter because of the way `oc process` works. Namespace attributes that are not parameterized are stripped when processing the template.

dgoodwin: Aha I understand, thanks I will update.

dgoodwin:
@csrwng I agree and have been wondering the same as I've been working on this. I can start followup PRs after this to begin moving us in that direction.

PR #161, opened 2018-04-18
Update cluster running status when provisioning has completed

csrwng:


csrwng:
@dgoodwin ptal

pkg/controller/machineset/machineset_controller.go
```diff
@@ -253,6 +262,58 @@ func (c *Controller) handleErr(err error, key interface{}) {
...
+		}
+		machineSet, err := c.getMachineSet(key)
+		if err != nil {
+			return err
+		}
+		cluster, err := controller.ClusterForMachineSet(machineSet, c.clustersLister)
+		if err != nil {
+			return err
+		}
+		if cluster.Status.Running || !cluster.Status.Provisioned {
```
dgoodwin: Should we rename this from Running to Ready?

pkg/controller/machineset/machineset_controller.go
```diff
@@ -253,6 +262,58 @@ func (c *Controller) handleErr(err error, key interface{}) {
...
+		return nil, err
+	}
+	if len(namespace) == 0 || len(name) == 0 {
+		return nil, fmt.Errorf("invalid key %q: either namespace or name is missing", key)
+	}
+	return c.machineSetsLister.MachineSets(namespace).Get(name)
+}
+
+func (c *Controller) checkClusterRunning(syncFunc func(string) error) func(string) error {
+	return func(key string) error {
```
dgoodwin: Might be helpful to have some debug logs sprinkled through here for some easier visibility as to why a cluster isn't yet marked ready.

pkg/controller/machineset/machineset_controller.go
```diff
@@ -253,6 +262,58 @@ func (c *Controller) handleErr(err error, key interface{}) {
...
+		machineSets, err := controller.MachineSetsForCluster(cluster, c.machineSetsLister)
+		if err != nil {
+			return nil
+		}
+		if len(machineSets) != len(cluster.Spec.MachineSets) {
+			return nil
+		}
+		for _, ms := range machineSets {
+			if ms.Spec.NodeType == clusteroperator.NodeTypeMaster {
+				if !ms.Status.Provisioned || !ms.Status.Installed || !ms.Status.ComponentsInstalled || !ms.Status.NodeConfigInstalled {
```
dgoodwin: Should we be checking Accepted here as well for masters?

csrwng: masters are automatically joined to the cluster and don't need to run a separate accept step

csrwng:
@dgoodwin renamed Running to Ready in cluster status and added debug logging to the function that checks whether the cluster is ready.

PR #163, opened 2018-04-19
Set the current context of the cluster kubeconfig to the external one

csrwng:
Before creating the kubeconfig secret, set the current context in the kubeconfig to the one that will allow external access.


csrwng:
@joelddiaz ptal

PR #167, opened 2018-04-19
Override openshift_master_cluster_public_hostname as the external ELB hostname.

abutcher:


csrwng:
@abutcher if this is only temporary then it's fine. I'm concerned about overwriting a base task file and then having to keep it in sync with upstream.
/lgtm

PR #168, opened 2018-04-23
Kubectl apply module

dgoodwin:
These commits switch us from using shell commands / k8s_raw to my in-progress kubectl_apply and oc_process wrapper modules. Per discussion last week we're going to use our deployment Ansible as a testbed for these and see where they go. The path back to k8s_raw once it's more production ready should be quite straightforward.

With this change Ansible 2.5 is no longer required.

The roles are vendored using another tool I wrote (gogitit), but this is only relevant to the person doing the vendoring. None the less the metadata files are available if anyone else needed to do this, and the README is updated with details on how.

### DIFF
--- a/contrib/ansible/create-cluster-playbook.yaml
+++ b/contrib/ansible/create-cluster-playbook.yaml
@@ -15,7 +15,7 @@
     # ID for your cluster, defaults to current username:
     cluster_name: "{{ lookup ('env', 'USER') }}"
 
-    # ClusterVersion to install, default to fake so we don't actually create a cluster unless user
+    # ClusterVersion to use, default to fake so we don't actually create a cluster unless user
     # explicitly opts-in:
     cluster_version: origin-v3-10-fake
 
@@ -43,6 +43,9 @@
     # in the cert dir defined above.
     cluster_ca_name: ca
 
+  roles:
+    - kubectl-ansible
+
   tasks:
 
   # If no cluster_namespace was defined on the CLI, we want to create the cluster in the current:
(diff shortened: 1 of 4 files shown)


dgoodwin:
Going to do a separate PR for vendoring don't like how this looks. Please hold on this one.

dgoodwin:
This one ready for review now. @joelddiaz @abutcher

contrib/ansible/create-cluster-playbook.yaml
```diff
@@ -43,6 +43,9 @@
     # in the cert dir defined above.
     cluster_ca_name: ca
 
+  roles:
```
joelddiaz: can we just use 'import_role' https://docs.ansible.com/ansible/2.4/import_role_module.html in the 'tasks' section below instead of adding a 'roles' section? this should accomplish loading the contents/library of kubectl-ansible into the playbook so that the kubectl_apply and oc_process later will still work.

contrib/ansible/deploy-devel-playbook.yml
```diff
@@ -12,18 +12,28 @@
 - hosts: localhost
   connection: local
   gather_facts: no
+  roles:
```
joelddiaz: same comment about using 'import_role'

contrib/ansible/deploy-playbook.yaml
```diff
@@ -28,17 +28,20 @@
     # in the cert dir defined above.
     apiserver_ca_name: ca
 
+  roles:
```
joelddiaz: import_role

joelddiaz:
@dgoodwin i have a suspicion that 'roles' will be deprecated in upcoming ansible releases (hence my suggestion to use 'import_role')

PR #169, opened 2018-04-23
Vendor in kubectl-ansible role/modules.

dgoodwin:
Disable vetting of headers on our ansible sub-dir.


dgoodwin:
@joelddiaz @abutcher could one of you take at look at this and the follow up: https://github.com/openshift/cluster-operator/pull/168

This one is pretty simple just vendoring to keep the other PR clean, don't feel obligated to review all of these modules in vendoring though if you do spot something let me know and I'll fix it upstream.

Need these to unblock on my next one.

PR #171, opened 2018-04-24
Replace use of openssl_certificate with a custom certgen Ansible role.

dgoodwin:
Replace openssl_certificate with custom certgen role.
    
    The cert module in Ansible is a no-go on RHEL, and we don't want to rule
    that out yet. Replace our older apiserver certgen script with a role we
    can use for both apiserver certs, and individual cluster certs.
    
    Generates a CA if one does not exist and uses it for all certs.
    
    Each cluster gets it's own cert to prevent problems with the same cert
    being uploaded anywhere in AWS.


contrib/ansible/create-cluster-playbook.yaml
```diff
@@ -37,9 +37,11 @@
 
     # Paths where we will generate self-signed certs for the specific cluster. (probably don't need to
     # override these but if you wanted to use pre-existing certs you could)
-    cluster_cert_path: "{{ playbook_dir }}/../../certs/{{ cluster_name }}.pem"
-    cluster_privatekey_path: "{{ playbook_dir }}/../../certs/{{ cluster_name }}-key.pem"
-    cluster_csr_path: "{{ playbook_dir }}/../../certs/{{ cluster_name }}.csr"
+    cluster_cert_dir: "{{ playbook_dir }}/../../certs"
```
csrwng: add this dir to .gitignore

contrib/ansible/create-cluster-playbook.yaml
```diff
@@ -82,25 +84,22 @@
...
-      provider: selfsigned
+  - include_role:
+      name: cluster_operator_certgen
+    vars:
+      cert_dir: "{{ cluster_cert_dir }}"
+      ca_name: "{{ cluster_ca_name }}"
+      cert_name: "{{ cluster_name }}"
+      # TODO: do something meaningful here?
+      cn: "{{ cluster_name }}"
+      alt_names: "{{ cluster_name }}.{{ cluster_namespace }}\",\"{{ cluster_name }}.{{ cluster_namespace }}.svc"
```
csrwng: How does this work? If I expand this, I would get a string like:
`cewong-cluster.myproject","cewong-cluster.myproject.svc`

contrib/ansible/deploy-playbook.yaml
```diff
@@ -44,23 +50,30 @@
...
-      creates: "{{ playbook_dir }}/../../apiserver.pem"
-    when: redeploy_apiserver_cert | bool
+  - include_role:
+      name: cluster_operator_certgen
+    vars:
+      cert_dir: "{{ apiserver_cert_dir }}"
+      ca_name: "{{ apiserver_ca_name }}"
+      cert_name: "apiserver"
+      cn: cluster-operator-apiserver
+      alt_names: "cluster-operator-apiserver.{{ cluster_operator_namespace }}\",\"cluster-operator-apiserver.{{ cluster_operator_namespace }}.svc"
```
csrwng: same comment as above

contrib/ansible/roles/cluster_operator_certgen/templates/cert-config.json
```diff
@@ -0,0 +1 @@
+{"CN":"{{ cn }}","hosts":["{{ alt_names }}"],"key":{"algo":"rsa","size":2048}}
```
csrwng: use .j2 suffix

contrib/ansible/roles/cluster_operator_certgen/templates/ca-config.json
```diff
@@ -0,0 +1 @@
+{"signing":{"default":{"expiry":"43800h","usages":["signing","key encipherment","server"]}}}
```
csrwng: If not a template, it should likely go in role/[name]/files

contrib/ansible/roles/cluster_operator_certgen/templates/ca-setup.json
```diff
@@ -0,0 +1,16 @@
+{
```
csrwng: use .j2 suffix

contrib/ansible/roles/cluster_operator_certgen/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+    dest: "{{ cert_config_file }}"
+  when: not cert_stat.stat.exists
+
+- name: generate cert
+  shell: "cfssl gencert -ca={{ ca_file }} -ca-key={{ ca_key_file }} -config={{ ca_config_file }} {{ cert_config_file }} | cfssljson -bare {{ cert_name }}"
+  args:
+    chdir: "{{ cert_dir }}"
+  when: not cert_stat.stat.exists
+
+#- name: cleanup temporary files
```
csrwng: Is there a reason you commented out this part?

contrib/examples/create-cluster.sh
```diff
@@ -0,0 +1,49 @@
+#! /bin/bash
```
csrwng: delete this file

contrib/ansible/create-cluster-playbook.yaml
dgoodwin: Pretty sure it's all just junk today, this is just modeled on how it was working before, and we were actually using the apiserver's cert in any case. Definitely needs to be looked at so added the TODO for both this and the CN.

contrib/ansible/deploy-playbook.yaml
dgoodwin: Not sure just following what was in the apiserver cert gen script.

contrib/ansible/roles/cluster_operator_certgen/tasks/main.yml
dgoodwin: Just debugging so I could diff these with the old variants, actually exposed the problem. Readding the cleanup next commit.

.gitignore
```diff
@@ -22,12 +22,5 @@ contrib/examples/*-cluster.yaml
...
 .var
 .vscode
-apiserver-key.pem
-apiserver.csr
-apiserver.pem
-cluster-operator-ca.json
-ca-key.pem
-ca.csr
-ca.pem
-server-ca-config.json
```
dgoodwin: Devs can clean these up, they are no longer used.

dgoodwin:
Updated for feedback and cleaned up.

PR #172, opened 2018-04-24
set registry count to the number of infra nodes (max 2)

joelddiaz:
provide the number of infra nodes when generating ansible variables so that we can do things like set the number of replicas for some components based on that size.
    
as a temporary measure, keep registy replicas hardcoded to 1 for now while we wait for a move to S3-backed storage as running more than 1 registry with non-shared storage (separate PVs in the current setup) will result in problems

update tests so that we have clusters with a machineset having infra: true

### DIFF
--- a/pkg/ansible/generate.go
+++ b/pkg/ansible/generate.go
@@ -91,6 +91,8 @@ openshift_master_bootstrap_enabled: True
 
 openshift_hosted_router_wait: False
 openshift_hosted_registry_wait: False
+# cap to size 1 instead of .InfraSize while we use PV-backed registry storage
+#openshift_hosted_registry_replicas: "{{ [ [[ .InfraSize ]], 2 ] | min }}"
 openshift_hosted_registry_replicas: 1
 
 # Override router edits to set the ROUTER_USE_PROXY_PROTOCOL
@@ -329,6 +331,7 @@ type clusterParams struct {
 	SSHUser        string
 	VPCDefaults    string
 	DeploymentType coapi.ClusterDeploymentType
+	InfraSize      int
 }
 
 type machineSetParams struct {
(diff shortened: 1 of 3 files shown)


pkg/ansible/generate.go
```diff
@@ -91,7 +91,7 @@ openshift_master_bootstrap_enabled: True
 
 openshift_hosted_router_wait: False
 openshift_hosted_registry_wait: False
-openshift_hosted_registry_replicas: 1
+openshift_hosted_registry_replicas:  [[ .InfraSize ]]
```
dgoodwin: Can we tie this approach into the router we just hard coded to 1?

joelddiaz: definitely more useful that just determining the number of registry pods to run.

dgoodwin:
Regarding S3, what happens right now with >1 replicas, do they still share storage correctly?

Do we need S3 registry for dedicated? I'm guessing yes, but probably not for CI?

Do we have any way to config S3 registry storage via the API? Probably not and I'm not sure it should be exposed over the API. Seems like it's something that should be handled by a registry operator and it's associated config.

pkg/controller/components/components_controller.go
```diff
@@ -310,6 +311,16 @@ func (s *jobSyncStrategy) GetLastJobSuccess(owner metav1.Object) *time.Time {
...
 }
 
+func (s *jobSyncStrategy) getInfraCount(machineSet *clusteroperator.MachineSet) int {
+	msList, _ := s.controller.machineSetsLister.MachineSets(machineSet.Namespace).List(labels.Everything())
+	for _, ms := range msList {
+		if ms.Spec.Infra && ms.Spec.NodeType == "Compute" && ms.ClusterName == machineSet.ClusterName {
+			return ms.Spec.Size
+		}
+	}
+	// if in doubt, at least make sure count of 1 is returned
```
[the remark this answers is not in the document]
joelddiaz: changed it to error out in the absence of any machineset with infra: true set

dgoodwin:
Feels like we should roll without the S3 buckets for now and make sure there's a trello card up for it in future.

Is this ready as far as you're concerned @joelddiaz ?

joelddiaz:
@dgoodwin got confirmation that running multiple registry replicas with non-shared storage will cause us issues. i've kept the hardcode to '1' for now, but still keeping the plumbing to provide the infra count so we can use in the future as needed

PR #173, opened 2018-04-24
Ensure certs dir exists when deploying cluster-operator

csrwng:


csrwng:
@abutcher @dgoodwin ptal

PR #177, opened 2018-04-26
Convert to builds, image streams and deployment configs.

dgoodwin:
Deploying will now create a BuildConfig, ImageStream, and
DeploymentConfig. However until you push images there's nothing for the
DeploymentConfig to run.

You can push images in two ways, either start-build to build the latest
git master, or a make target to push your locally built images.
(development)

May want to actually test this one.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Dockerfile to perform a full cluster-operator build from source.
+#
+# This Dockerfile is intended for performing in-cluster OpenShift builds. It differs
+# from those in build/ in that it does not expect to compile the binary externally
+# and add to the container, everything is built internally during docker build.
+FROM golang:1.9
+
+ENV PATH=/go/bin:$PATH GOPATH=/go
+
+ADD . /go/src/github.com/openshift/cluster-operator
+
+WORKDIR /go/src/github.com/openshift/cluster-operator
+RUN NO_DOCKER=1 make build
+ENTRYPOINT ["bin/cluster-operator"]
+
--- a/Makefile
+++ b/Makefile
@@ -390,6 +390,17 @@ aws-machine-controller-image: build/aws-machine-controller/Dockerfile $(BINDIR)/
 	docker tag $(AWS_MACHINE_CONTROLLER_IMAGE) $(AWS_MACHINE_CONTROLLER_MUTABLE_IMAGE)
 	docker tag $(AWS_MACHINE_CONTROLLER_IMAGE) $(AWS_MACHINE_CONTROLLER_PUBLIC_IMAGE)
 
+# Push our Docker Images to the integrated registry:
+INTEGRATED_REGISTRY                         ?= 172.30.1.1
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+FROM golang:1.9
+
+ENV PATH=/go/bin:$PATH GOPATH=/go
+
+ADD . /go/src/github.com/openshift/cluster-operator
+
+WORKDIR /go/src/github.com/openshift/cluster-operator
+RUN NO_DOCKER=1 make build
+ENTRYPOINT ["bin/cluster-operator"]
+
```
dgoodwin: Other Dockerfiles are not suitable for OpenShift in-cluster builds, there's a lot of layers of Docker build, multi-arch stuff, and inherently we build the binary outside and add to container. Adding this small new Dockerfile for openshift builds was a simple workaround.

Makefile
```diff
@@ -390,6 +390,16 @@ aws-machine-controller-image: build/aws-machine-controller/Dockerfile $(BINDIR)/
...
 
+# Push our Docker Images to the integrated registry:
+integrated-registry-push:
+	# WARNING: this will fail if logged in as system:admin, see README for creating an "admin" account
+	# you can use separately that will work here:
+	 $(eval OPENSHIFT_TOKEN := $(shell oc whoami -t))
+	docker login -u admin -p $(OPENSHIFT_TOKEN) 172.30.1.1:5000
+	# NOTE: the in-cluster ImageStream tag we use is latest:
+	docker tag cluster-operator:$(MUTABLE_TAG) 172.30.1.1:5000/openshift-cluster-operator/cluster-operator:latest
+	docker push 172.30.1.1:5000/openshift-cluster-operator/cluster-operator:latest
```
dgoodwin: Particularly wondering about registry IP here.

contrib/ansible/create-cluster-playbook.yaml
```diff
@@ -48,6 +48,17 @@
...
+  - name: process cluster versions template
+    oc_process:
+      template_file: "{{ playbook_dir }}/../examples/cluster-versions-template.yaml"
+    parameters:
+      CLUSTER_VERSION_NS: "{{ cluster_version_namespace }}"
+    register: cluster_versions_reg
+
+  - name: create/update cluster versions
+    kubectl_apply:
+      definition: "{{ cluster_versions_reg.result | to_json }}"
```
dgoodwin: Couldn't do this in the deploy playbook anymore as we don't necessarily have a running deployment after it completes, you need to push images or start-build.

README.md
```diff
@@ -30,25 +30,31 @@
     * Fedora: `oc cluster up --image="docker.io/openshift/origin"`
     * Mac OSX: Follow the Minishift [Getting Started Guide](https://docs.openshift.org/latest/minishift/getting-started/index.html)
     * Note: Startup output will contain the URL to the web console for your openshift cluster, save this for later
-  * Login to the OpenShift cluster as admin:
+  * Login to the OpenShift cluster as system:admin:
     * `oc login -u system:admin`
-  * Grant admin rights to login to the web console
+  * Create an "admin" account with cluster-admin role which you can use to login to the [WebUI](https://localhost:8443) or with oc:
```
dgoodwin: You need to oc login as this regular account so the whoami can give you a token, that does not work for system:admin.

Dockerfile
csrwng: We should use our regular Dockerfile to build the images so that they match what would be built by hand or elsewhere. This can be accomplished by using this Dockerfile to build the binaries and push the resulting image to an imagestream. Then use the image as a source with the regular Dockerfile. https://docs.openshift.org/latest/dev_guide/builds/build_inputs.html#image-source

Makefile
csrwng: If using cluster up, the registry IP is fixed.
What is the purpose of this target?

dgoodwin: To bundle all these commands so you've got a simple target to push to the integrate registry, is there a part of this I'm missing?

I could make the IP a variable with this as a default for anyone using minishift or something else.

csrwng: minishift uses cluster up, but yeah, you could make it a variable in case it's done on a different cluster.

PR #180, opened 2018-04-30
Update make build target for kubernetes-cluster-api image

csrwng:
Updates the build target to use the new cluster-api repository.


csrwng:
@dgoodwin ptal

PR #179, opened 2018-04-30
Update cluster-api vendored package

csrwng:
Updates the cluster-api vendored packages to the new official repo.

PR #183, opened 2018-05-01
Sync machine sets

abutcher:


pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,412 @@
...
+		}
+	*/
+
+	return nil
+}
+
+func buildClusterAPICluster(cluster *cov1.Cluster) *clusterapiv1.Cluster {
+	capiCluster := clusterapiv1.Cluster{}
+	capiCluster.Name = cluster.Name
+	capiCluster.Spec.ProviderConfig.Value = &runtime.RawExtension{
```
abutcher: I updated this based on https://github.com/openshift/cluster-operator/pull/179 (and swapped imports to `sig.k8s.io/cluster-api`). Hitting this error on the sync. :sos: @csrwng

```
time="2018-05-01T20:54:43Z" level=info msg="Error syncing master machine set: error creating cluster object in remote cluster: Cluster in version \"v1alpha1\" cannot be handled as a Cluster: v1alpha1.Cluster.Spec: v1alpha1.ClusterSpec.ProviderConfig: ReadString: expects \" or n, but found {, error found in #10 byte of ...|rConfig\":{\"value\":{\"|..., bigger context ...|ocks\":null},\"serviceDomain\":\"\"},\"providerConfig\":{\"value\":{\"kind\":\"Cluster\",\"apiVersion\":\"clusterope|..." controller=syncmachineset machineset="{default/abutcher-master-zp82t abutcher}"
```

csrwng: @abutcher completely forgot about that... that's because we haven't updated the cluster-api image on quay.io to the latest.

csrwng: @abutcher I've now updated the image to one built with the latest code. Please try again.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,555 @@
...
+
+		// Lookup clusterVersion for machineSet
+		coVersionNamespace, coVersionName := ms.Spec.ClusterVersionRef.Namespace, ms.Spec.ClusterVersionRef.Name
+		coClusterVersion, err := c.client.ClusteroperatorV1alpha1().ClusterVersions(coVersionNamespace).Get(coVersionName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("cannot retrieve cluster version %s/%s: %v", coVersionNamespace, coVersionName, err)
+		}
+
+		remoteMachineSet = buildClusterAPIMachineSet(ms, coClusterVersion)
+		_, erree := remoteClusterAPIClient.ClusterV1alpha1().MachineSets(remoteClusterAPINamespace).Create(remoteMachineSet)
```
staebler: What does `erree` mean? Are you just trying to avoid reusing `err`?

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,555 @@
...
+	}
+
+	remoteClusterAPIClient, remoteDeploymentClient, err := c.buildClients(ms)
+	if err != nil {
+		return err
+	}
+
+	// handleErr will requeue machine sets after waiting when the
+	// remote cluster API deployment is not ready.
+	remoteClusterAPIDeployment, err := remoteDeploymentClient.AppsV1().Deployments(remoteClusterAPINamespace).Get(remoteClusterAPIDeploymentName, metav1.GetOptions{})
```
staebler: Same concern here as with master machine set about doing this query on every controller sync of the compute machine set.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,555 @@
...
+	if err != nil {
+		return NewErrClusterAPINotInstalled(fmt.Sprintf("unable to retrieve remote cluster API deployment: %v", err))
+	}
+	if remoteClusterAPIDeployment.Status.ReadyReplicas < 1 {
+		return NewErrClusterAPINotInstalled(errorMsgClusterAPINotInstalled)
+	}
+
+	// Create machineSet object in remote if the machineSet does not already exist.
+	remoteMachineSet, err := remoteClusterAPIClient.ClusterV1alpha1().MachineSets(remoteClusterAPINamespace).Get(ms.Name, metav1.GetOptions{})
```
staebler: Same concern here as with master machine set. We should be deliberate about verifying that the machine set in the remote cluster is up-to-date on a schedule rather than on every controller sync of the compute machine set.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,555 @@
...
+	capiCluster.Spec.ClusterNetwork.Services.CIDRBlocks = []string{"172.30.0.0/16"}
+
+	return &capiCluster
+}
+
+func serializeCOResource(object runtime.Object) string {
+	serializer := jsonserializer.NewSerializer(jsonserializer.DefaultMetaFactory, coapi.Scheme, coapi.Scheme, false)
+	encoder := coapi.Codecs.EncoderForVersion(serializer, cov1.SchemeGroupVersion)
+	buffer := &bytes.Buffer{}
+	encoder.Encode(object, buffer)
```
staebler: Don't swallow this error. The `serializerCOResource` function should return the error.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,555 @@
...
+
+	// These are unused, dummy values.
+	capiCluster.Spec.ClusterNetwork.ServiceDomain = "cluster-api.k8s.io"
+	capiCluster.Spec.ClusterNetwork.Pods.CIDRBlocks = []string{"10.10.0.0/16"}
+	capiCluster.Spec.ClusterNetwork.Services.CIDRBlocks = []string{"172.30.0.0/16"}
+
+	return &capiCluster
+}
+
+func serializeCOResource(object runtime.Object) string {
```
staebler: Return a `[]byte` rather than a `string` as we ultimately want a `[]byte` anyway. This saves us from converting the `[]byte]` to a `string` and then immediately back to a `[]byte`. Call `Bytes()` instead of `String()` on the `bytes.Buffer` to get a `[]byte` to return.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,555 @@
...
+	}
+
+	return nil
+}
+
+func buildClusterAPIMachineSet(machineSet *cov1.MachineSet, clusterVersion *cov1.ClusterVersion) *clusterapiv1.MachineSet {
+	if machineSet.Annotations == nil {
+		machineSet.Annotations = map[string]string{}
+	}
+	machineSet.Annotations["cluster-operator.openshift.io/cluster-version"] = serializeCOResource(clusterVersion)
```
staebler: Move this down to where `machineSet` is serialized into the machine template. Here it looks like a mistake that the annotation is added to `machineSet` instead of `capiMachineSet`.

Also, do a deep copy of `machineSet` and use that instead to be clear that you are not intending to modify the original machine set.

pkg/controller/syncmachineset/syncmachineset_controller.go
abutcher: Just avoiding reusing `err`, exactly.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,555 @@
...
+
+// Creates the cluster in the remote.
+func (c *Controller) syncMasterMachineSet(ms *cov1.MachineSet) error {
+	msLog := logging.WithMachineSet(c.logger, ms)
+	msLog.Debug("loaded master machineset")
+
+	// handleErr will requeue machine sets after waiting when the
+	// remote cluster API is not installed.
+	if !ms.Status.ClusterAPIInstalled {
+		return NewErrClusterAPINotInstalled(errorMsgClusterAPINotInstalled)
```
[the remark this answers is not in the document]
abutcher: machine sets take a long time to sync initially since they get dropped while waiting for the remote cluster API to be ready so this was a workaround to make them sync as soon as the remote API is good. Based on https://github.com/openshift/cluster-operator/pull/183#discussion_r187442333, it sounds like the alternative would be to enqueue compute sets gathered from the cluster object when we're dealing with the master machine set rather than separately?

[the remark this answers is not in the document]
staebler: I don't quite understand this comment. My point with my comment was that the machine set will already be requeued by the informer when the `ClusterAPIInstalled` field is changed. There is no need to requeue it on a timer. In other words, if the controller sees that `ClusterAPIInstalled` is false and the controller has not been notified by the informer that the machine set has changed, then the `ClusterAPIInstalled` field will still be false on the check 20 seconds later.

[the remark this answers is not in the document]
abutcher: Ahhh, I'll remove the requeue from master sets. `ClusterAPIInstalled` isn't currently set on non master machine sets so they don't get requeued. I'll need to do something else to avoid requeueing on wait with compute sets.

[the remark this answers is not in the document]
staebler: For non-master machine sets, they should be requeued when the master machine set for the cluster is updated. In `addMachineSet` and `updateMachineSet` if the machine set being added/updated is a master machine set, then the other non-master machine sets for the cluster should be enqueued.

pkg/controller/syncmachineset/syncmachineset_controller.go
staebler: I would suggest that you just reuse `err`.

staebler:
Let me try to (hopefully) better explain my largest concern. For the sake of scalability, we want to minimize as much as possible the number of queries that the root cluster makes to the remote clusters. So, for example, if the controller has already established that a machine set exists in the remote cluster for a given machine set in the root cluster, then the controller should not be making additional queries looking for that remote machine set unless there are conditions that necessitate a new query. There are two such conditions. The first is that the spec of the machine set in the root cluster has changed. The second is that a certain period of time has elapsed since the last validation of the remote machine set (say 10 minutes).

Here are the changes that I propose we make.
1) The machine set in the root cluster has a generation number stored in its status that reflects the generation of the machine set that has been stamped to the remote cluster.
2) The machine set in the root cluster has a timestamp stored in its status that reflects the last time that the controller has validated the machine set in the remote cluster.
3) When the controller syncs a machine set, it only queries the remote cluster if (1) the generation of the spec of the machine set in the root cluster is different than the generation stored in the status or (2) sufficient time has elapsed since the timestamp stored in the machine set in the root cluster.
4) The controller will need to continuously requeue each machine set so that the controller has an opportunity to sync the machine set after the validation has expired.

abutcher:
@staebler Comments addressed aside from remote queries which I'll leave here for now as this code is temporary and will be around until we switch to syncing machine deployments.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,563 @@
...
+func (c *Controller) enqueue(machineSet *cov1.MachineSet) {
+	key, err := controller.KeyFunc(machineSet)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("couldn't get key for object %#v: %v", machineSet, err))
+		return
+	}
+
+	c.queue.Add(queuedMachineSet{
+		key:         key,
+		clusterName: machineSet.Labels[controller.ClusterNameLabel],
```
staebler: Why are we verifying that the label exists in `enqueueAfter` but not in `enqueue`?

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,563 @@
...
+
+	// Machines that need to be synced
+	queue workqueue.RateLimitingInterface
+
+	logger log.FieldLogger
+}
+
+// queuedMachineSet represents data we need to process a machine set in the queue. This includes
+// data we might need to process deleted objects we can no longer lookup.
+type queuedMachineSet struct {
```
staebler: What is the purpose of this? I don't see `clusterName` used anywhere.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,563 @@
...
+	}
+	coMasterMachineSet, err := c.machineSetsLister.MachineSets(coCluster.Namespace).Get(coCluster.Status.MasterMachineSetName)
+	if err != nil {
+		return fmt.Errorf("error looking up cluster's master machineset: %v", err)
+	}
+
+	// handleErr will requeue machine sets after waiting when the
+	// remote cluster API deployment is not ready.
+	if !coMasterMachineSet.Status.ClusterAPIInstalled {
+		return NewErrClusterAPINotInstalled(errorMsgClusterAPINotInstalled)
```
staebler: I am still opposed to the approach of returning an error here instead of enqueueing the compute machine sets when the master machine set is updated. I don't think that this is a situation where the alternative is too much work to justify given the expected short-term nature of this controller.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,563 @@
...
+	coCluster, err := c.clusterLister.Clusters(ms.Namespace).Get(ms.Labels[controller.ClusterNameLabel])
+
+	if err != nil {
+		return fmt.Errorf("error looking up machineset's cluster: %v", err)
+	}
+	msLog = logging.WithCluster(msLog, coCluster)
+
+	if !ms.Status.ClusterAPIInstalled {
+		msLog.Debugf(errorMsgClusterAPINotInstalled)
+		return fmt.Errorf(errorMsgClusterAPINotInstalled)
```
staebler: There is no reason why this should not return nil instead of an error.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,563 @@
...
+// Creates machinesets in the remote.
+func (c *Controller) syncComputeMachineSet(ms *cov1.MachineSet) error {
+	msLog := logging.WithMachineSet(c.logger, ms)
+	msLog.Debug("loaded compute machineset")
+
+	// Lookup the CO cluster object for this machine set and use
+	// that to get the master machine set for the cluster. Continue
+	// syncing generic machine set if the master machine set is
+	// installed (ms.Status.ClusterAPIInstalled == true).
+	// TODO: Set a flag on cluster object once it has been synced
```
staebler: I am actively working on moving `ClusterAPIInstalled` to the ClusterStatus so that this will be moot.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,563 @@
...
+	capiCluster.Spec.ClusterNetwork.Services.CIDRBlocks = []string{"172.30.0.0/16"}
+
+	return &capiCluster
+}
+
+func serializeCOResource(object runtime.Object) []byte {
+	serializer := jsonserializer.NewSerializer(jsonserializer.DefaultMetaFactory, coapi.Scheme, coapi.Scheme, false)
+	encoder := coapi.Codecs.EncoderForVersion(serializer, cov1.SchemeGroupVersion)
+	buffer := &bytes.Buffer{}
+	encoder.Encode(object, buffer)
```
staebler: Don't swallow errors returning from `Encode`.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,563 @@
...
+	client     coclient.Interface
+	kubeClient kubeclientset.Interface
+
+	// To allow injection of syncMachineSet for testing.
+	syncHandler func(queuedMS queuedMachineSet) error
+
+	// Used for unit testing
+	enqueueMachineSet func(machineSet *cov1.MachineSet)
+
+	buildClients func(cluster *cov1.MachineSet) (clusterapiclient.Interface, kubeclientset.Interface, error)
```
staebler: s/cluster *cov1.MachineSet/*cov1.MachineSet/

pkg/controller/syncmachineset/syncmachineset_controller.go
staebler: Let's remove the `queuedMachineSet` struct entirely. Simply queue the keys. That will keep it consistent with how other controllers work.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,548 @@
...
+	utilruntime.HandleError(err)
+	logger.Infof("dropping machine set out of the queue: %v", err)
+	c.queue.Forget(queuedMS)
+}
+
+func isMasterMachineSet(machineSet *cov1.MachineSet) bool {
+	return machineSet.Spec.NodeType == cov1.NodeTypeMaster
+}
+
+func (c *Controller) buildRemoteClusterClients(machineSet *cov1.MachineSet) (clusterapiclient.Interface, error) {
```
staebler: We should extract this function since it is doing the same thing as this function. We can do that as follow-on work, though.

https://github.com/openshift/cluster-operator/blob/1d8612326880b51296b3fee44537c3d78abde5e6/pkg/controller/cluster/cluster_controller.go#L367

pkg/controller/syncmachineset/syncmachineset_controller_test.go
```diff
@@ -0,0 +1,311 @@
...
+)
+
+const (
+	testRegion           = "us-east-1"
+	testClusterVerUID    = types.UID("test-cluster-version")
+	testClusterVerName   = "v3-9"
+	testClusterVerNS     = "cluster-operator"
+	testClusterUUID      = types.UID("test-cluster-uuid")
+	testClusterName      = "testcluster"
+	testClusterNamespace = "testcluster"
```
staebler: Should use a different name for namespace and name so that we can ensure that we aren't accidentally using one for the other in the code.

pkg/controller/syncmachineset/syncmachineset_controller.go
```diff
@@ -0,0 +1,548 @@
...
+	// itself.
+	if isMasterMachineSet(machineSet) {
+		return c.syncMasterMachineSet(machineSet)
+	}
+
+	return c.syncComputeMachineSet(machineSet)
+}
+
+// Creates the cluster in the remote.
+func (c *Controller) syncMasterMachineSet(ms *cov1.MachineSet) error {
```
staebler: This should really be a controller that is syncing on the cluster instead of the machineset, since it is the cluster that is the operative component. It's fine for now though given that this is temporary.

PR #185, opened 2018-05-02
Fixes to cluster-api deployment and export

csrwng:
In the latest version of cluster API, the controller manager no longer takes a --cloud flag, and some fields in Spec.ClusterNetwork are now required.


csrwng:
@dgoodwin ptal
@abutcher you'll need this fix as well

dgoodwin:
What's going on with https://github.com/openshift/cluster-operator/pull/184 and this? I can't spot the difference.

csrwng:
Absolutely no difference. I looked this morning and didn’t see my pr in the list of open prs so I concluded I must have forgotten to submit it :-D

/close

PR #188, opened 2018-05-03
Add docker-py module to ansible image

csrwng:
Recent PR to openshift-ansible makes this a prereq
https://github.com/openshift/openshift-ansible/pull/8225


csrwng:
@abutcher ptal

PR #189, opened 2018-05-03
Use specific tag for openshift-ansible 1.10

csrwng:
Switches to specific tag for openshift-ansible 1.10 in the Makefile


csrwng:
@dgoodwin @abutcher ptal

PR #191, opened 2018-05-04
Drop orphaned kube-deploy vendored code.

dgoodwin:


csrwng:
Ha, I didn't realize I had left this in there. Thx.
/lgtm

PR #195, opened 2018-05-07
Fix infinitely stalled devel deployment.

dgoodwin:
Due to the order in which things merged, master currently will stall on
a devel deployment waiting for it to complete, when we expect you to
make a decision on how to get images. (make push-integrated registry, or
oc start-build)

With this change devel deployments now assume to run make images, and
make integrated-registry-push, then wait for the deployment to roll out.
You no longer need to choose a deployment method.

This makes a devel deployment playbook run much slower, but this should
not be part of normal workflow, typically you just run make images push
yourself, or run your controller externally when working.

With this change creation of the default cluster versions can move back
to it's more logical location.

### DIFF
--- a/contrib/ansible/create-cluster-playbook.yml
+++ b/contrib/ansible/create-cluster-playbook.yml
@@ -51,17 +51,6 @@
   - import_role:
       name: kubectl-ansible
 
-  - name: process cluster versions template
-    oc_process:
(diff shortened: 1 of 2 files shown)


csrwng:
/lgtm 
(test is past the point where this was failing before)

PR #196, opened 2018-05-07
store cluster.k8s.io resources properly in etcd

staebler:
Store each resource type in its own subkey in etcd.


dgoodwin:
This looks promising, fixes the errors on my branch, I can create/list both clusters and machinesets now.

/lgtm

PR #197, opened 2018-05-07
Create new cluster APIs with a playbook var.

dgoodwin:
@staebler this might be useful.


dgoodwin:
Fixed the bug CI found.

PR #198, opened 2018-05-07
add cluster-api controllers into cluster operator namespace

joelddiaz:
add a template to handle deploying aws-machine-controller and cluster-api-controller-manager into the default cluster operator namespace

fix up some semi-related typos found during this work


dgoodwin:
Deploys nicely but looks like your branch was a bit old, and you don't have Matt's latest API server work, so the logs are full of: ERROR: logging before flag.Parse: E0507 18:46:25.602741       1 reflector.go:205] sigs.k8s.io/cluster-api/pkg/controller/sharedinformers/zz_generated.api.register.go:54: Failed to list *v1alpha1.Cluster: the server could not find the requested resource (get clusters.cluster.k8s.io)

Could you rebase and I'll re-test.

contrib/examples/cluster-api-controllers-template.yaml
```diff
@@ -0,0 +1,160 @@
...
+    name: "clusterapi.openshift.io:controller-manager"
+  roleRef:
+    apiGroup: rbac.authorization.k8s.io
+    kind: ClusterRole
+    name: "clusterapi.openshift.io:controller-manager"
+  subjects:
+  - apiGroup: ""
+    kind: ServiceAccount
+    name: cluster-api-controller-manager
+    namespace: ${CLUSTER_API_NAMESPACE}
```
dgoodwin: This role and role binding should go to the cluster-operator-roles-template.yml, which gets copied to openshift/release so they can run it for us and create what we need. (we don't have cluster admin in CI cluster)

joelddiaz:
@dgoodwin rebased and moved the cluster role definition/binding as requested

PR #199, opened 2018-05-07
Add utility to extract e2e test container logs

csrwng:
Adds command that will extract container logs from a given Jenkins job. It will create a separate file for each container log either in the current directory or a directory specified by `--output-dir`

### DIFF
--- a/contrib/cmd/extract-co-jenkins-logs/main.go
+++ b/contrib/cmd/extract-co-jenkins-logs/main.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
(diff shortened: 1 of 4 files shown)


csrwng:
@dgoodwin ptal

contrib/cmd/extract-logs/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/openshift/cluster-operator/contrib/cmd/extract-logs/util"
+)
+
+// Given a Jenkins URL, extracts logs of relevant containers to current or specified
+// directory
+
+// NewExtractLogsCommand returns a command that will extract logs
+func NewExtractLogsCommand(outputDir string) *cobra.Command {
+	return &cobra.Command{
+		Use: "extract-co-logs JENKINS_BUILD_URL",
```
dgoodwin: Binary name mismatch, I would personally like the more specific form, maybe even extract-co-jenkins-logs.

contrib/cmd/extract-logs/main.go
```diff
@@ -0,0 +1,86 @@
...
+		},
+	}
+}
+
+func main() {
+	log.SetOutput(os.Stdout)
+	var logLevel string
+	var outputDir string
+	pflag.CommandLine.StringVar(&logLevel, "log-level", "info", "Log level: debug, info, warning, error, fatal, panic")
+	pflag.CommandLine.StringVar(&outputDir, "output-dir", "", "Output directory")
```
dgoodwin: Could you default this to a sub-dir created below CWD, pretty easy to dump a whole bunch of logfiles somewhere you might not have intended.

csrwng:
> Ideally this almost seems like something the CI jobs should be doing when they publish the artifacts.

That's a great idea. Once we get this merged, we can have the Jenkins job build it, extract the logs, and publish them in the job page.

dgoodwin:
Would it make sense for all of CI not just us?

csrwng:
Not many other jobs run cluster up to do stuff

dgoodwin:
Ah ok was thinking it might be useful for pulling cluster logs off any test cluster but I may have missed the cluster up specifics.

csrwng:
@dgoodwin comments addressed. It will now allow you to specify local filenames for container.log and docker.info so it can be run inside the test vm.

PR #204, opened 2018-05-08
Provide a means for adding custom validation

staebler:
This is not super clean as supplementing the validation is not something that the apiserver-builder is designed to support. Nevertheless, it works without too much trouble.

* Adds sample validation to Machine that can be removed once some real validation of any of the cluster-api resources has been added.

### DIFF
--- a/pkg/apis/clusterapi/validation/machine.go
+++ b/pkg/apis/clusterapi/validation/machine.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validation
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apiserver/pkg/endpoints/request"
+
+	"github.com/golang/glog"
+
+	"github.com/kubernetes-incubator/apiserver-builder/pkg/builders"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/cluster-api/pkg/apis/cluster"
+)
+
+// MachineStrategy is the strategy that the API server will use for Machine
+// resources.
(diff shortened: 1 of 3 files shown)


pkg/apis/clusterapi/validation/machine.go
```diff
@@ -0,0 +1,54 @@
...
+	builders.StorageBuilder
+}
+
+// Validate checks that an instance of Machine is well formed
+func (m MachineStrategy) Validate(ctx request.Context, obj runtime.Object) field.ErrorList {
+	errors := field.ErrorList{}
+	errors = append(errors, m.StorageBuilder.Validate(ctx, obj)...)
+	machine := obj.(*cluster.Machine)
+	glog.Infof("Custom validation of Machine %s", machine.Name)
+	return errors
```
dgoodwin: Ok for my and @joelddiaz sake let me know if I understand this correctly, in theory here we'd be looking to combine any upstream Machine validation https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha1/machine_types.go#L168 (but they don't actually have any yet), with any of our custom requirements on whether or not a Machine is valid, into one ErrorList here.

pkg/apiserver/etcd_config.go
```diff
@@ -115,7 +117,9 @@ func (c completedEtcdConfig) NewServer(stopCh <-chan struct{}) (*ClusterOperator
 		groupInfos = append(groupInfos, groupInfo)
 	}
 
-	caGroupInfo := ca_apis.GetClusterAPIBuilder().Build(roFactory)
+	caAPIBuilder := ca_apis.GetClusterAPIBuilder()
+	validation.ReplaceStorageBuilder(caAPIBuilder, ca.InternalMachine, validation.ReplaceMachineStorageBuilder)
```
dgoodwin: And we would repeat here for the API types we're more interested in which are Cluster and MachineSet and soon MachineDeployment.

These probably all have valid upstream deployment validation (for example: https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha1/cluster_types.go#L112) which we'd want to combine with our validation on the providerConfig using: https://github.com/openshift/cluster-operator/blob/master/pkg/apis/clusteroperator/validation/cluster.go

Does this sound right?

dgoodwin: And add the check on MachineSet for a label linking it to a cluster.

pkg/apis/clusterapi/validation/machine.go
staebler: Precisely.

pkg/apiserver/etcd_config.go
staebler: Yes.

PR #205, opened 2018-05-08
Provide a way to get the clust-op cluster spec from ProviderConfig

staebler:
`ClusterSpecFromClusterAPI` function added to decode the cluster-operator `ClusterSpec` from `ProviderConfig` in a cluster-api `Cluster`.

*Update*
Do the same for getting `ClusterStatus` from `ProviderStatus`.
This requires the latest version of cluster-api. Did a full glide up to update all the dependencies.


pkg/apis/clusteroperator/types.go
```diff
@@ -144,6 +144,19 @@ type AWSRegionAMIs struct {
...
+
+// ClusterProviderConfigSpec is the cluster specification stored in the
+// ProviderConfig of a cluster.k8s.io Cluster.
+type ClusterProviderConfigSpec struct {
+	// +optional
+	metav1.TypeMeta
+	// +optional
+	metav1.ObjectMeta
+
+	ClusterSpec
```
dgoodwin: Do we get any value from having two layers here, it feels like one might do, i.e. just embed our ClusterSpec (possibly renamed).

staebler: At the moment, we are still using ClusterSpec in the cluster-operator Cluster. So I didn't want to introduce TypeMeta and ObjectMeta into that ClsuterSpec.

PR #208, opened 2018-05-08
Provide function for patch cluster-api Cluster status

staebler:
More work pulled from https://github.com/dgoodwin/cluster-operator/commits/cluster-api.


dgoodwin:
I hope this works, also never got to test it.

/lgtm

PR #209, opened 2018-05-09
Document how to run controllers locally for development productivity.

dgoodwin:
Look right @csrwng ? Seems to be working for me, and super handy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,20 @@
 
 You can then check the provisioning status of your cluster by running `oc describe cluster <cluster_name>`
 
+## Developing Cluster Operator Controllers Locally
+
+If you are actively working on controller code you can save some time by compiling and running locally:
+
+  * Run the deploy playbooks normally.
+  * Disable your controller in the cluster-operator-controller-manager DeploymentConfig using *one* of the below methods:
+    * Scale everything down: `oc scale -n openshift-cluster-operator --replicas=0 dc/cluster-operator-controller-manager`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,6 +68,16 @@
 
 You can then check the provisioning status of your cluster by running `oc describe cluster <cluster_name>`
 
+## Developing Cluster Operator Controllers Locally
+
+If you are actively working on controller code you can save some time by compiling and running locally:
+
+  * Run the deploy playbooks normally.
+  * `oc delete DeploymentConfig cluster-operator-controller-manager` - This will be recreated the next time you run the deploy playbooks.
```
csrwng: You should include the namespace parameter `-n openshift-cluster-operator`

You can also just scale it down if you want to bring it back up later as well
`oc scale --replicas=0 dc/cluster-operator-controller-manager -n openshift-cluster-operator`

csrwng: Or if you want other controllers to run except the one you're developing, `oc edit dc/cluster-operator-controller-manager -n openshift-cluster-operator` and add an argument of of --controllers=[comma-separated list of controllers]`

README.md
```diff
@@ -68,6 +68,16 @@
...
+  * `make build`
```
csrwng: Not sure if it's worth mentioning, but on my Mac, I usually do `go install ./cmd/cluster-operator` to produce a darwin binary.

README.md
staebler: Or just `--controllers=-[name of controller you don't want to run]`. For example, if you are working on the infra controller and don't want to run it as part of the deployment, add the argument `--controllers=-infra`.

dgoodwin: All very handy, the -controller is nice, did not know about that. I will add these all in.

dgoodwin:
Updated.

PR #212, opened 2018-05-09
Check availability of cluster API resources

csrwng:
Includes cluster api resources when getting available resources


csrwng:
But if we eventually run the upstream apiserver separately we would. If ok with you, I'd rather use the separate clients even if they resolve to the same thing right now. If you like, I'll pass the same func to retrieve the client to both calls.

staebler:
@csrwng Good point. You've convinced me. Let's keep it as you have it.

PR #213, opened 2018-05-09
infra controller for cluster-api

staebler:
This modifies the infra controller so that it can be used with the cluster-api resources as well.

The result of syncing infra controller on a `Cluster` from cluster-api is a `ProviderStatus` filled out in the `Cluster` like so.

```
    providerStatus:
      apiVersion: clusteroperator.openshift.io/v1alpha1
      conditions:
      - lastProbeTime: 2018-05-10T18:26:32Z
        lastTransitionTime: 2018-05-10T18:26:32Z
        message: Job completed successful
        reason: JobCompleted
        status: "True"
        type: InfraProvisioned
      kind: ClusterProviderStatus
      machineSetCount: 0
      metadata:
        creationTimestamp: null
      provisioned: true
      provisionedJobGeneration: 1
      ready: false
```


pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -658,3 +660,18 @@ const (
 	// NodeTypeCompute is a node that is a compute node in the cluster
 	NodeTypeCompute NodeType = "Compute"
 )
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// CombinedCluster is a cluster object that combines the information from
+// cluster-operator Cluster and cluster-api Cluster.
+type CombinedCluster struct {
```
dgoodwin: Is this type temporary?

staebler: Not really. We need some way of keeping the decoded ProviderConfig and ProviderStatus around while we work on the Cluster. So I would expect that we will finalize on something similar to this even after we remove Cluster from clusteroperator.openshift.io.

dgoodwin: Ok struck me as a little odd given the ClusterAPI* structs already contain the other ClusterOperator* portions, just not decoded.

pkg/controller/controller_utils.go
```diff
@@ -309,10 +312,10 @@ func MachineSetLabels(cluster *clusteroperator.Cluster, machineSetShortName stri
 
 // JobLabelsForClusterController returns the labels to apply to a job doing a task
 // for the specified cluster.
-func JobLabelsForClusterController(cluster *clusteroperator.Cluster, jobType string) map[string]string {
+func JobLabelsForClusterController(cluster metav1.Object, jobType string) map[string]string {
```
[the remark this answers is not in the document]
staebler: Done.

staebler:
It should be noted that the workaround for `ClusterVersion` will only work if the `ClusterVersion` exists prior to the infra controller attempting to sync the `Cluster`.

pkg/controller/controller_utils_test.go
```diff
@@ -1166,8 +1166,34 @@ kind: Machine
...
 		})
 	}
 }
+
+func TestClusterAPIProviderStatusFromClusterStatus(t *testing.T) {
+	clusterStatus := &clusteroperator.ClusterStatus{
+		MachineSetCount:      1,
+		MasterMachineSetName: "master",
+	}
+	//expectedProviderStatus: `{"apiVersion":"clusteroperator.openshift.io/v1alpha1","kind":"ClusterProviderStatus","machineSetCount":1,"masterMachineSetName":"master"}`,
```
dgoodwin: Can remove now.

dgoodwin:
This looks fine to me, I am going to merge as I would like to build on the status updating code. If @csrwng has any feedback we can do a followup PR.

/lgtm

PR #214, opened 2018-05-10
Delete target machinesets

csrwng:


pkg/apis/clusteroperator/types.go
```diff
@@ -312,6 +312,10 @@ type ClusterStatus struct {
 	// ClusterVersionRef references the resolved clusterversion the cluster should be running.
 	// +optional
 	ClusterVersionRef *corev1.ObjectReference
+
+	// DeprovisionedComputeMachinesets is true of the compute machinesets of this cluster
```
staebler: s/of/if/

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -313,6 +313,10 @@ type ClusterStatus struct {
 	// ClusterVersionRef references the resolved clusterversion the cluster should be running.
 	// +optional
 	ClusterVersionRef *corev1.ObjectReference `json:"clusterVersionRef,omitempty"`
+
+	// DeprovisionedComputeMachinesets is true of the compute machinesets of this cluster
```
staebler: s/of/if/

pkg/controller/cluster/cluster_controller.go
```diff
@@ -356,7 +366,91 @@ func (c *Controller) processNextWorkItem() bool {
...
+	}
+
+	clusterAPIClient, err := clusterapiclient.NewForConfig(restConfig)
+	if err != nil {
+		return nil, err
+	}
+	return clusterAPIClient, nil
+}
+
+func (c *Controller) ensureRemoteMachineSetsAreDeleted(cluster *clusteroperator.Cluster) error {
```
staebler: I have some concern that trying to run operations on a remote cluster to which the connection is slow or inaccessible will adversely effect the ability of the controller to timely deal with other tasks. Is there a way that we can run these remote operations in a separate goroutine and have the controller pick up the results later? I am fine tackling that as later work.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -356,7 +366,91 @@ func (c *Controller) processNextWorkItem() bool {
...
+	if len(machineSets.Items) == 0 {
+		original := cluster
+		cluster = cluster.DeepCopy()
+		cluster.Status.DeprovisionedComputeMachinesets = true
+		return controller.PatchClusterStatus(c.client, original, cluster)
+	}
+
+	for _, ms := range machineSets.Items {
+		if ms.DeletionTimestamp != nil {
+			err := clusterAPIClient.ClusterV1alpha1().MachineSets(remoteClusterNamespace).Delete(ms.Name, &metav1.DeleteOptions{})
```
staebler: Can we use `DeleteCollection` to delete all the relevant machine sets with one call rather than making separate `Delete` calls?

pkg/controller/cluster/cluster_controller.go
```diff
@@ -356,7 +366,91 @@ func (c *Controller) processNextWorkItem() bool {
...
+	}
+	return clusterAPIClient, nil
+}
+
+func (c *Controller) ensureRemoteMachineSetsAreDeleted(cluster *clusteroperator.Cluster) error {
+	clusterAPIClient, err := c.buildRemoteClusterClient(cluster)
+	if err != nil {
+		return err
+	}
+	machineSets, err := clusterAPIClient.ClusterV1alpha1().MachineSets(remoteClusterNamespace).List(metav1.ListOptions{})
```
staebler: Can we be absolutely sure that all of the machine sets in "kube-cluster" are relevant machine sets? Is it possible that a user could create their own machine sets in "kube-cluster" that are associated with a different cluster?

csrwng: For now, we don't have anything linking machinesets to the cluster. I would say that we will need to make kube-cluster (or whatever other namespace we choose) special in that it represents the machinesets that belong to that cluster. I would not expect other machinesets (that do not need to be deprovisioned) to exist in that cluster.

pkg/controller/cluster/cluster_controller.go
csrwng: Sgtm, I'll run these in their own goroutine.

pkg/controller/cluster/cluster_controller.go
csrwng: Sorry in that namespace

pkg/controller/cluster/cluster_controller.go
csrwng: Yup, will fix

csrwng:
@staebler just fyi... this code will become irrelevant when we switch to syncing clusterapi clusters instead of c-o clusters. However, whatever problems we run into with this one will still need to be addressed in the other.

csrwng:
@staebler comments addressed. Created 2 follow-up issues: #215 and #216

pkg/controller/cluster/cluster_controller.go
```diff
@@ -356,7 +366,96 @@ func (c *Controller) processNextWorkItem() bool {
...
+
+func (c *Controller) setDeprovisionedComputeMachinesets(cluster *clusteroperator.Cluster) error {
+	original := cluster
+	cluster = cluster.DeepCopy()
+	cluster.Status.DeprovisionedComputeMachinesets = true
+	return controller.PatchClusterStatus(c.client, original, cluster)
+}
+
+func (c *Controller) ensureRemoteMachineSetsAreDeleted(cluster *clusteroperator.Cluster) error {
+	clusterAPIClient, err := c.buildRemoteClusterClient(cluster)
```
staebler: Can we rename this variable to include the term "remote" so that as I read the code I immediately know that the client is communicating with the remote cluster and not the local cluster?

pkg/controller/cluster/cluster_controller.go
```diff
@@ -356,7 +366,96 @@ func (c *Controller) processNextWorkItem() bool {
...
+	if cluster.Status.DeprovisionedComputeMachinesets {
+		return false, nil
+	}
+	if len(cluster.Status.MasterMachineSetName) == 0 {
+		return false, nil
+	}
+	masterMachineSet, err := c.machineSetsLister.MachineSets(cluster.Namespace).Get(cluster.Status.MasterMachineSetName)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
```
staebler: Are you sure that we want to swallow this error? The Cluster status indicates that there is a master machine set. I would think that it would be an error worth bubbling up if it turns out that said master machine set does not actually exist.

pkg/controller/cluster/cluster_controller.go
```diff
@@ -356,7 +366,96 @@ func (c *Controller) processNextWorkItem() bool {
...
+
 func (c *Controller) ensureMachineSetsAreDeleted(cluster *clusteroperator.Cluster) error {
+	shouldDeleteRemote, err := c.shouldDeleteRemoteMachineSets(cluster)
+	if err != nil {
+		return err
+	}
+	if shouldDeleteRemote {
+		return c.ensureRemoteMachineSetsAreDeleted(cluster)
+	}
+	if !cluster.Status.DeprovisionedComputeMachinesets {
```
staebler: I am concerned about the potential for a race condition here where other controllers may be in the process of installing the cluster-api and provisioning the compute machinesets in the remote cluster.

pkg/controller/jobsyncstrategy.go
```diff
@@ -71,3 +71,10 @@ type JobSyncStrategy interface {
 	// do not want to support reprocessing should return nil.
 	GetLastJobSuccess(owner metav1.Object) *time.Time
 }
+
+// CheckBeforeDeprovision should be implemented by a strategy to have the sync loop check
+// whether a particular owner resource is ready for deprovisioning when a DeletionTimestamp is detected.
+type CheckBeforeDeprovision interface {
```
staebler: We should use the same terminology here that we use elsewhere in `JobSync`. Specifically, the inverse of process is undo. `JobSync` does not have a concept of deprovision. Either `CheckBeforeDeprovision` should be `CheckBeforeUndo` or we should change the undo terminology to something else in `JobSync`.

pkg/controller/cluster/cluster_controller.go
csrwng: So the problem is that let's say that for some reason the master machine set is deleted, then the cluster itself is marked for deletion. The cluster reconcile loop will not recreate the master machineset and if I return an error from here, we will keep retrying to find a master machineset that doesn't exist.

pkg/controller/jobsyncstrategy.go
csrwng: Ack

csrwng:
Additional review comments addressed

pkg/controller/cluster/cluster_controller.go
staebler: OK. Makes sense.

pkg/controller/cluster/cluster_controller.go
staebler: That wouldn't be sufficient. That would only decrease the window of the race condition. There is still the chance that this controller uses a revision of the Cluster that does not have the status updated to reflect that the cluster-api is installing. We could avoid this if we did not patch and retry on conflicts.

csrwng: We talked before about not doing the patch/retry thing, but came back to the conclusion that it could cause more problems having several controllers trying to update the same resource. Something that maybe we could explore doing is having at most one controller per resource. Either change I think would be beyond the scope of this PR :) Would you be ok with creating an issue and looking into fixing it separately?

csrwng: Could a fix be to move this logic to the controller that provisions the machinesets on to the remote cluster? We would be guaranteed that we're not trying to delete the remote machinesets at the same time that we're provisioning them.

staebler: In reconsidering this, I am fine with kicking this can down the road in the interest of pushing forward the migration to using the upstream API fully.

csrwng:
@staebler tests are passing for this now

PR #217, opened 2018-05-14
WIP: Add cluster-api machine/machineset controllers.

dgoodwin:
These are responsible for populating some portions of the machineSpec
providerConfig on both machines and machinesets.

- calculate the VMImage to use (uses cluster version and we expect
additional logic in future, which will only need to exist in the root
cluster)
- copy the cluster hardware spec for use in deletion (when cluster may
no longer exist)
- apply hardware defaults if applicable

These are only used in the root cluster, when we decide to send a
machineset down to the target cluster, we will update the spec
beforehand.

AWS actuator is updated to use the correct data going forward. (followup
PR will solve credentials to the actuator can actually create the master
machineset in the root cluster, presently this is still broken)


pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	clusterSpec, err := controller.ClusterSpecFromClusterAPI(cluster)
+	if err != nil {
+		return err
+	}
+
+	// TODO: once cluster controller is setting resolved cluster version refs on the
+	// ClusterStatus, replace this with the resolved ref not a lookup
+	clusterVersion, err := c.clustopClient.ClusteroperatorV1alpha1().ClusterVersions(clusterSpec.ClusterVersionRef.Namespace).Get(clusterSpec.ClusterVersionRef.Name, metav1.GetOptions{})
```
staebler: Why do we use a lister to fetch the Cluster but the client directly to fetch the ClusterVersion?

pkg/controller/clusterapimachineset/machineset_controller.go
```diff
@@ -0,0 +1,293 @@
...
+		return err
+	}
+
+	clusterVersion, err := c.clustopClient.ClusteroperatorV1alpha1().ClusterVersions(clusterSpec.ClusterVersionRef.Namespace).Get(clusterSpec.ClusterVersionRef.Name, metav1.GetOptions{})
+	if err != nil {
+		msLog.WithField("cluster version does not exist: %v", clusterSpec.ClusterVersionRef)
+		return err
+	}
+
+	machineSpec, err := controller.PopulateMachineSpec(ms.Spec.Template.Spec, clusterSpec, clusterVersion, msLog)
```
staebler: I don't understand why we have this controller at all. Presumably, this is running in the root cluster, where the user is creating the MachineSet objects. We should not have a controller that is modifying user-created objects. The whole point of the capi-machine controller is to update the template for the Machine objects, that cluster-operator (via cluster-api) owns.

If we stop trying to write the Cluster and ClusterVersion information into the MachineSet, then we won't have a problem with trying to store user-specified and cluster-derived defaults in the same field as detailed in my comment in `PopulateMachineSpec`.

contrib/examples/clusterapi-template.yml
```diff
@@ -130,55 +130,64 @@ objects:
...
     name: ${CLUSTER_NAME}-master
     namespace: ${CLUSTER_NAMESPACE}
+    labels:
+      cluster: ${CLUSTER_NAME}
   spec:
     replicas: 1
     selector:
       matchLabels:
-        machineset: master
+        machineset: ${CLUSTER_NAME}-master
```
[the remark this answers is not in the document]
dgoodwin: They do not but IMO they should, no movement on the issues opened though.

pkg/apis/clusteroperator/v1alpha1/types.go
```diff
@@ -416,19 +433,31 @@ type MachineSetConfig struct {
...
 	ClusterVersionRef corev1.ObjectReference `json:"clusterVersionRef"`
+
+	// VMImage contains a single cloud provider specific image to use for this machine set.
+	VMImage VMImage `json:"vmImage"`
+}
+
+// VMImage contains a specified single image to use for a supported cloud provider.
+type VMImage struct {
+	// +optional
+	AWSImage *string `json:"awsImage"`
```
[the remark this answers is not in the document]
dgoodwin: No, just optional per API conventions, as we expand to other clouds this may or may not be set.

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,300 @@
...
+	c.enqueueMachine = c.enqueue
+	c.codecFactory = clustopapi.Codecs
+
+	return c
+}
+
+// Controller manages clusters.
+type Controller struct {
+	clustopClient clustopclientset.Interface
+	capiClient    clusterclientset.Interface
```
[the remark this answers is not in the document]
dgoodwin: Lets go with capi and clustop.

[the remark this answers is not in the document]
dgoodwin: We should probably apply that to the imports as well, capiv1 for ex.

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	if err != nil {
+		return err
+	}
+
+	mLog := colog.WithClusterAPIMachine(c.logger, machine)
+	mLog.Debugf("found machine")
+
+	// Lookup the cluster for this machine:
+	clusterName, ok := machine.Labels[clusterNameLabel]
```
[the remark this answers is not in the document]
dgoodwin: Does upstream require this? I am not aware of such a requirement, though I'd definitely say they expect it to be so, but we explicitly do not. As such I would say that yes the labels are required, and we've got issues open to address upstream, IMO all objects should always be explicitly linked.

pkg/controller/clusterapimachine/machine_controller.go
dgoodwin: Which should be used when?

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,300 @@
...
+
+	machineSpec, err := controller.PopulateMachineSpec(machine.Spec, clusterSpec, clusterVersion, mLog)
+	if err != nil {
+		mLog.Errorf("error populating machine spec: %v", err)
+		return err
+	}
+	machine.Spec = *machineSpec
+
+	// TODO: updating the machine here, should this be a patch operation on the template spec?
+	// NOTE: updating here immediately triggers another sync...
```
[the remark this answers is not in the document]
dgoodwin: Does this imply use of the patching we do on status?

pkg/controller/clusterapimachineset/machineset_controller.go
dgoodwin: The controller exists to make sure the spec template is updated before we send down to the target cluster (soon) and it starts to act on it, as we don't intend to have the capi-machine controller running remotely and there are no cluster versions defined. I expected we would do similar for MachineDeployment. 

We need some place to pass that data, and ideally do some calculation logic that remains under the root clusters control. If all of MachineTemplateSpec providerConfig is off limits I'm not sure how we do that.

pkg/controller/clusterapimachine/machine_controller.go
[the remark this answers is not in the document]
staebler: This is the function that convinces me that upstream, at least as it is currently implemented, requires a one-to-one. https://github.com/kubernetes-sigs/cluster-api/blob/5e323ca383afc6ddb4132b4fe80bd08ff3817374/pkg/controller/machine/controller.go#L153

pkg/controller/clusterapimachine/machine_controller.go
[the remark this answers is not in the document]
staebler: Patching shouldn't used. We don't want to do a partial update of the RawExtension. Patching versus updating is orthogonal to only updating on changes.

pkg/controller/clusterapimachine/machine_controller.go
staebler: I don't know the official stance on this. My ruke of thumb is to prefer Lister for any type that is used regularly and of which we need a large subset of the total population. ClusterVersion fits into that description. Lister saves us from having to go back to the API Server to service the query at the cost of having to store the objects locally and watch the API Server for changes.

pkg/controller/clusterapimachineset/machineset_controller.go
staebler: I'm not sure why we wouldn't fill out the template at the time that we are sending the data to the remote cluster. I don't see why we need to mutate our root cluster MachineSet objects to satisfy the information being in the remote Machine templates.

pkg/controller/clusterapimachine/machine_controller.go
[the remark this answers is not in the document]
dgoodwin: Yup that's not good, we should discuss if we want to change our plan or push for change upstream. We'd need the mandatory labels per https://github.com/kubernetes-sigs/cluster-api/issues/145

pkg/controller/clusterapimachineset/machineset_controller.go
dgoodwin: We could but then we've got different representations of the same thing in the root and the target which strikes me as quite unusual, and if we get into syncing or ever supporting changes made remotely, we'll have some problems to deal with. Possible that can be reconciled but it's something we should watch out for. I can back this controller up and pull it out of this PR though.

pkg/apis/clusteroperator/v1alpha1/types.go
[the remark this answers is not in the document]
dgoodwin: Let me know if this is not correct otherwise I'll assume to leave it in.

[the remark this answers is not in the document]
staebler: I suppose that `*string` is fine for now. I would expect that one of two things happen as we add more clouds. Either (1) the `*string` will be replaced by a pointer to an AWS-specific struct, that may just contain a `string`, or (2) we forego the union idea since the the VM image information is the same for every cloud and there is a one-to-one relationship between machineset and cloud.

pkg/controller/controller_utils.go
```diff
@@ -402,3 +428,129 @@ func ClusterAPIProviderStatusFromClusterStatus(clusterStatus *clusteroperator.Cl
...
+	if err != nil {
+		return nil, err
+	}
+	return &runtime.RawExtension{
+		Raw: buffer.Bytes(),
+	}, nil
+}
+
+// PopulateMachineSpec ensures that a the MachineSetSpec we use for the machine spec provider config is fully populated with defaults and calculated values based for Cluster Operator specific logic. This can be used both on individual machines, as well as on the MachineTemplateSpec used in MachineSets and MachineDeployments.
+func PopulateMachineSpec(machineSpec clusterapi.MachineSpec, clusterSpec *clusteroperator.ClusterSpec, clusterVersion *clusteroperator.ClusterVersion, mLog log.FieldLogger) (*clusterapi.MachineSpec, error) {
```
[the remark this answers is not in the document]
dgoodwin: Per your above comment I think we're ok here for now having deleted the machineset controller, but we'll evolve this if necessary. Clearly separating would be nice depending on what happens with MachineSet + MachineDeployment when we send to target cluster.

pkg/controller/controller_utils.go
```diff
@@ -402,3 +428,129 @@ func ClusterAPIProviderStatusFromClusterStatus(clusterStatus *clusteroperator.Cl
...
+				AWSImage: &regionAMI.AMI,
+			}, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no AWS image defined for region %s", clusterSpec.Hardware.AWS.Region)
+}
+
+// applyDefaultMachineHardwareSpec merges the cluster-wide hardware defaults with the machineset specific hardware specified.
+func applyDefaultMachineSetHardwareSpec(machineSetHardwareSpec, defaultHardwareSpec *clusteroperator.MachineSetHardwareSpec) (*clusteroperator.MachineSetHardwareSpec, error) {
```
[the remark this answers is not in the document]
staebler: Oh, I get it. The zero values will be omitted from the JSON when marshalled.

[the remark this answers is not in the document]
staebler: Can we do better than copying this routine and having to maintain it in two places?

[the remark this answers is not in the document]
dgoodwin: Of course, I did not realize there was still a copy somewhere. cluster_controller.go is going to die off completely, but I will replace it's use regardless.

dgoodwin:
Updated and dropping WIP, ready for merge or more review. I do want to revisit that is generated vs set by user on that template in future though.

staebler:
> Updated and dropping WIP

@dgoodwin You did not drop the WIP. Is that an oversight? Or is it still WIP?

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,301 @@
...
+	c.machinesLister = msInformer.Lister()
+	c.machinesSynced = msInformer.Informer().HasSynced
+
+	c.syncHandler = c.syncMachine
+	c.enqueueMachine = c.enqueue
+
+	return c
+}
+
+// Controller manages clusters.
```
staebler: Comment is not accurate.

cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -338,7 +339,9 @@ func NewControllerInitializers() map[string]InitFunc {
 	controllers["nodeconfig"] = startNodeConfigController
 	controllers["deployclusterapi"] = startDeployClusterAPIController
 
+	// Controllers ported to upstream cluster API
```
staebler: What does it mean for the controllers to be "ported to upstream cluster API"? At best this comment is confusing. At worst it is inaccurate.

cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -536,6 +539,21 @@ func startClusterController(ctx ControllerContext) (bool, error) {
 	return true, nil
 }
 
+func startMachineAPIController(ctx ControllerContext) (bool, error) {
+	if !resourcesAvailable(ctx) {
```
staebler: This always returns true since it is not checking for any resources.

cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -536,6 +539,21 @@ func startClusterController(ctx ControllerContext) (bool, error) {
...
+		return false, nil
+	}
+	go machineapi.NewController(
+		ctx.ClusterAPIInformerFactory.Cluster().V1alpha1().Clusters(),
+		ctx.ClusterAPIInformerFactory.Cluster().V1alpha1().Machines(),
+		ctx.InformerFactory.Clusteroperator().V1alpha1().ClusterVersions(),
+		ctx.ClientBuilder.KubeClientOrDie("clusteroperator-capi-machine-controller"),
+		ctx.ClientBuilder.ClientOrDie("clusteroperator-capi-machine-controller"),
+		ctx.ClientBuilder.ClusterAPIClientOrDie("clusteroperator-capi-machine-controller"),
+	).Run(int(ctx.Options.ConcurrentClusterSyncs), ctx.Stop)
```
staebler: It is not appropriate to use `ConcurrentClusterSyncs` here.

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,301 @@
...
+*/
+
+package clusterapimachine
+
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	//metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
staebler: Remove.

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,301 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/golang/glog"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/openshift/cluster-operator/pkg/kubernetes/pkg/util/metrics"
+
+	//clustopv1 "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
```
staebler: Remove.

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,301 @@
...
+
+	c.syncHandler = c.syncMachine
+	c.enqueueMachine = c.enqueue
+
+	return c
+}
+
+// Controller manages clusters.
+type Controller struct {
+	clustopClient clustopclientset.Interface
```
staebler: Unused.

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// Lookup the cluster for this machine:
+	clusterName, ok := machine.Labels[clusterNameLabel]
+	if !ok {
+		return fmt.Errorf("no cluster label set on machine")
+	}
+	cluster, err := c.clustersLister.Clusters(namespace).Get(clusterName)
+	if errors.IsNotFound(err) {
+		mLog.WithField("cluster", clusterName).Errorf("no cluster found for machine")
+		return nil
```
staebler: A Machine may be created prior to the Cluster with which the Machine is associated. In that case, the controller will not process the Machine until the Machine is resynced from the API Server. We should be watching Clusters and reqeueueing Machines when on updates of their Clusters.

pkg/controller/clusterapimachine/machine_controller.go
```diff
@@ -0,0 +1,301 @@
...
+	newMachine := machine.DeepCopy()
+	err = controller.PopulateMachineSpec(&newMachine.Spec, clusterSpec, clusterVersion, mLog)
+	if err != nil {
+		mLog.Errorf("error populating machine spec: %v", err)
+		return err
+	}
+
+	if !apiequality.Semantic.DeepEqual(machine, newMachine) {
+		// TODO: updating the machine here, should this be a patch operation on the template spec?
+		// NOTE: updating here immediately triggers another sync...
```
staebler: I don't think that this note is necessary. That is overwhelmingly (if not universally) true in controllers.

pkg/clusterapi/aws/actuator.go
```diff
@@ -425,41 +416,20 @@ func (a *Actuator) clusterOperatorClusterSpec(c *clusterv1.Cluster) (*cov1.Clust
...
-		return nil, nil, err
-	}
-	coMachineSet, ok := obj.(*cov1.MachineSet)
-	if !ok {
-		return nil, nil, fmt.Errorf("Unexpected machine set object: %#v", obj)
-	}
-	rawClusterVersion, ok := coMachineSet.Annotations[clusterVersionAnnotation]
-	if !ok {
-		return nil, nil, fmt.Errorf("Missing ClusterVersion resource annotation in MachineSet %#v", coMachineSet)
+func (a *Actuator) clusterOperatorMachineSetSpec(m *clusterv1.Machine) (*cov1.MachineSetSpec, error) {
```
staebler: Can we use `MachineSetSpecFromClusterAPIMachineSpec` from `pkg/controller/controller_utils.go` instead of defining this function?

pkg/controller/controller_utils.go
```diff
@@ -402,3 +405,129 @@ func ClusterAPIProviderStatusFromClusterStatus(clusterStatus *clusteroperator.Cl
...
+	if ms.ProviderConfig.Value == nil {
+		return nil, fmt.Errorf("No Value in ProviderConfig")
+	}
+	obj, _, err := api.Codecs.UniversalDecoder(clusteroperator.SchemeGroupVersion).Decode([]byte(ms.ProviderConfig.Value.Raw), nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	spec, ok := obj.(*clusteroperator.MachineSetProviderConfigSpec)
+	if !ok {
+		return nil, fmt.Errorf("Unexpected object: %#v", obj)
```
staebler: As in `clusterOperatorMachineSetSpec` in `pkg/clusterapi/aws/actuator.go`, let's use the `GroupVersionKind` instead of dumping the object.

pkg/controller/controller_utils.go
```diff
@@ -402,3 +405,129 @@ func ClusterAPIProviderStatusFromClusterStatus(clusterStatus *clusteroperator.Cl
...
+	}
+
+	if clusterVersion.Spec.VMImages.AWSImages == nil {
+		return nil, fmt.Errorf("no AWS images defined for cluster version")
+	}
+
+	for _, regionAMI := range clusterVersion.Spec.VMImages.AWSImages.RegionAMIs {
+		if regionAMI.Region == clusterSpec.Hardware.AWS.Region {
+			return &clusteroperator.VMImage{
+				AWSImage: &regionAMI.AMI,
```
staebler: Copy `regionAMI.AMI` to a new string and use a pointer to that. It would be really hard to track down if somewhere the value stored in this pointer where modified and it mutated a ClusterVersion stored in the lister.

pkg/logging/logging.go
```diff
@@ -33,3 +35,18 @@ func WithMachineSet(logger log.FieldLogger, machineSet *clusteroperator.MachineS
 func WithCluster(logger log.FieldLogger, cluster *clusteroperator.Cluster) log.FieldLogger {
 	return logger.WithField("cluster", fmt.Sprintf("%s/%s", cluster.Namespace, cluster.Name))
 }
+
+// WithClusterAPI expands a logger's context to include info about the given cluster.
+func WithClusterAPI(logger log.FieldLogger, cluster *clusterv1.Cluster) log.FieldLogger {
```
staebler: s/WithClusterAPI/WithClusterAPICluster/

cmd/cluster-operator-controller-manager/app/controllermanager.go
dgoodwin: PR started out as a port of cluster controller to use the new API, and has been through many iterations since. I will update it.

Please keep tone of PR comments in mind.

pkg/controller/clusterapimachine/machine_controller.go
dgoodwin: I assume this is in addition to requeueing on Machine changes.

Would it be reasonable to only requeue on Cluster creation (not update), as I don't think there's any action to take on a cluster update right now, at this point I still think this is just a snapshot in time once it's set.

cmd/cluster-operator-controller-manager/app/controllermanager.go
dgoodwin: Updating to check for both API resources being ready as this uses both.

pkg/clusterapi/aws/actuator.go
dgoodwin: Almost surely, this is quite old and was written in a separate repo IIRC.

pkg/controller/controller_utils.go
dgoodwin: Updating this here and for a couple other similar methods in this package.

dgoodwin:
PR updates.

pkg/controller/clusterapimachine/machine_controller.go
staebler: No, I don't think that is reasonable.
1) The user should be able to update the Cluster.
2) We will very soon be inspecting the ClusterVersionRef from the Status, which will certainly be done on an update rather than at creation.

dgoodwin: (2) has me convinced. Updated with update/delete cluster listeners.

staebler: Why are you not swayed by (1)? What do you expect the behavior of cluster-operator to be when a user changes the spec of a cluster?

dgoodwin:
Updated again.

/test unit

PR #223, opened 2018-05-17
Use global IAM role names

csrwng:
Switches to using global names for IAM roles instead of names suffixed with the cluster ID


staebler:
Looks fine to me. But I don't have enough background around the IAM to give it a lot, so I'll defer to someone else for that.

dgoodwin:
@abutcher could you do the final honors?

PR #226, opened 2018-05-18
Cluster API cluster controller

csrwng:
This controller adds owner references to machinesets that are labeled
with the cluster's name. It uses the same owner reference manager
used in Kube to set owner references on PodSets based on a label
selector.

### DIFF
--- a/cmd/cluster-operator-controller-manager/app/controllermanager.go
+++ b/cmd/cluster-operator-controller-manager/app/controllermanager.go
@@ -330,6 +330,7 @@ func NewControllerInitializers() map[string]InitFunc {
 	controllers := map[string]InitFunc{}
 
 	controllers["cluster"] = startClusterController
+	controllers["clusterapicluster"] = startClusterAPIClusterController
 	controllers["infra"] = startInfraController
 	controllers["machineset"] = startMachineSetController
 	controllers["machine"] = startMachineController
@@ -535,6 +536,19 @@ func startClusterController(ctx ControllerContext) (bool, error) {
 		ctx.ClientBuilder.KubeClientOrDie("clusteroperator-cluster-controller"),
 		ctx.ClientBuilder.ClientOrDie("clusteroperator-cluster-controller"),
 	).Run(int(ctx.Options.ConcurrentClusterSyncs), ctx.Stop)
+	return true, nil
+}
+
+func startClusterAPIClusterController(ctx ControllerContext) (bool, error) {
+	if !clusterOperatorResourcesAvailable(ctx) {
+		return false, nil
+	}
+	go cluster.NewClusterAPIController(
+		ctx.ClusterAPIInformerFactory.Cluster().V1alpha1().Clusters(),
+		ctx.ClusterAPIInformerFactory.Cluster().V1alpha1().MachineSets(),
+		ctx.ClientBuilder.KubeClientOrDie("clusterapi-cluster-controller"),
+		ctx.ClientBuilder.ClusterAPIClientOrDie("clusterapi-cluster-controller"),
+	).Run(int(ctx.Options.ConcurrentClusterAPIClusterSyncs), ctx.Stop)
 	return true, nil
 }
 
--- a/pkg/controller/cluster/clusterapi_cluster_controller.go
+++ b/pkg/controller/cluster/clusterapi_cluster_controller.go
@@ -1,0 +1,335 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cluster
+
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	kubeclientset "k8s.io/client-go/kubernetes"
+	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/golang/glog"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/openshift/cluster-operator/pkg/kubernetes/pkg/util/metrics"
+
+	clusterapi "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+	clusterapiclient "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset"
+	informers "sigs.k8s.io/cluster-api/pkg/client/informers_generated/externalversions/cluster/v1alpha1"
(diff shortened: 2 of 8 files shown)


cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -538,6 +539,19 @@ func startClusterController(ctx ControllerContext) (bool, error) {
...
+func startClusterAPIClusterController(ctx ControllerContext) (bool, error) {
+	if !clusterOperatorResourcesAvailable(ctx) {
+		return false, nil
+	}
+	go cluster.NewClusterAPIController(
+		ctx.ClusterAPIInformerFactory.Cluster().V1alpha1().Clusters(),
+		ctx.ClusterAPIInformerFactory.Cluster().V1alpha1().MachineSets(),
+		ctx.ClientBuilder.KubeClientOrDie("clusterapi-cluster-controller"),
+		ctx.ClientBuilder.ClusterAPIClientOrDie("clusterapi-cluster-controller"),
+	).Run(int(ctx.Options.ConcurrentClusterAPIClusterSyncs), ctx.Stop)
```
dgoodwin: I've just been re-using the non-cluster-api values, as we're going to kill off the distinction shortly. Logical to just re-use cluster syncs here?

pkg/controller/cluster/clusterapi_cluster_controller.go
```diff
@@ -0,0 +1,335 @@
+/*
```
dgoodwin: Filename here is a different convention than I've been doing (adding them in separate packages), but technically I think I like this better.

pkg/controller/cluster/clusterapi_cluster_controller.go
```diff
@@ -0,0 +1,335 @@
...
+
+	"github.com/golang/glog"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/openshift/cluster-operator/pkg/kubernetes/pkg/util/metrics"
+
+	clusterapi "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+	clusterapiclient "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset"
+	informers "sigs.k8s.io/cluster-api/pkg/client/informers_generated/externalversions/cluster/v1alpha1"
+	listers "sigs.k8s.io/cluster-api/pkg/client/listers_generated/cluster/v1alpha1"
```
dgoodwin: Staebler and I tried to settle on some standards for these imports, right now that looks like capiv1, capiclient. informers and listers probably fine if we're not distinguishing between the two.

pkg/controller/cluster/clusterapi_cluster_controller.go
```diff
@@ -0,0 +1,335 @@
...
+
+	"github.com/openshift/cluster-operator/pkg/controller"
+	colog "github.com/openshift/cluster-operator/pkg/logging"
```
dgoodwin: I started using clustoplog, clustop being the prefix equivalent to capi.

cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -330,6 +330,7 @@ func NewControllerInitializers() map[string]InitFunc {
 	controllers := map[string]InitFunc{}
 
 	controllers["cluster"] = startClusterController
+	controllers["clusterapicluster"] = startClusterAPIClusterController
```
dgoodwin: capi-cluster here as well.

pkg/controller/cluster/clusterapi_cluster_controller.go
```diff
@@ -0,0 +1,335 @@
...
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a Cluster %#v", obj))
+			return
+		}
+	}
+	colog.WithClusterAPICluster(c.logger, cluster).Debugf("deleting cluster")
+	c.enqueueCluster(cluster)
+}
+
+// When a machine set is created, enqueue the cluster that
```
dgoodwin: Comment lost it's train of thought. :)

pkg/controller/cluster/clusterapi_cluster_controller.go
```diff
@@ -0,0 +1,335 @@
...
+		if cluster == nil {
+			return
+		}
+	}
+	colog.WithClusterAPIMachineSet(colog.WithClusterAPICluster(c.logger, cluster), machineSet).Debugln("machineset created")
+	c.enqueueCluster(cluster)
+}
+
+// When a machine set is updated, figure out what cluster manages it and wake it
+// up.
```
dgoodwin: Is it ok for these comments to ignore the normal godoc standard of method name first because they're not exported? (for my own knowledge most likely)

pkg/controller/cluster/clusterapi_cluster_controller.go
```diff
@@ -0,0 +1,335 @@
...
+		return err
+	}
+	r, err := labels.NewRequirement(clusterAPIClusterNameLabel, selection.Equals, []string{cluster.Name})
+	if err != nil {
+		return err
+	}
+	machineSetsSelector := labels.NewSelector().Add(*r)
+
+	// If any adoptions are attempted, we should first recheck for deletion with
+	// an uncached quorum read sometime after listing ReplicaSets (see #42639).
```
dgoodwin: Comment looks like it should be updated.

pkg/controller/cluster/machineset_refmgr.go
```diff
@@ -0,0 +1,167 @@
...
+		return m.Selector.Matches(labels.Set(obj.GetLabels()))
+	}
+	adopt := func(obj metav1.Object) error {
+		return m.AdoptMachineSet(obj.(*capiv1.MachineSet))
+	}
+	release := func(obj metav1.Object) error {
+		return m.ReleaseMachineSet(obj.(*capiv1.MachineSet))
+	}
+
+	for _, rs := range sets {
```
dgoodwin: rs could probably use a rename.

pkg/controller/cluster/machineset_refmgr.go
```diff
@@ -0,0 +1,167 @@
...
+		m.Controller.GetName(), m.Controller.GetUID(), rs.UID)
+	_, err := m.machineSetClient.MachineSets(rs.Namespace).Patch(rs.Name, types.StrategicMergePatchType, []byte(addControllerPatch))
+	return err
+}
+
+// ReleaseMachineSet sends a patch to free the MachineSet from the control of the Deployment controller.
+// It returns the error if the patching fails. 404 and 422 errors are ignored.
+func (m *MachineSetControllerRefManager) ReleaseMachineSet(machineSet *capiv1.MachineSet) error {
+	glog.V(2).Infof("patching MachineSet %s_%s to remove its controllerRef to %s/%s:%s",
+		machineSet.Namespace, machineSet.Name, m.controllerKind.GroupVersion(), m.controllerKind.Kind, m.Controller.GetName())
```
dgoodwin: Let's kill off glog here too if we can.

pkg/controller/controller_ref_manager.go
```diff
@@ -0,0 +1,117 @@
...
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		// Either someone else claimed it first, or there was a transient error.
+		// The controller should requeue and try again if it's still orphaned.
+		return false, err
+	}
+	// Successfully adopted.
+	return true, nil
+}
```
dgoodwin: What do we think about the unit test situation here. I assume this is based on Deployment -> ReplicaSet, and unit tested there. Do we want our own variants or is it just too time consuming or awkward to do.

cmd/cluster-operator-controller-manager/app/controllermanager.go
csrwng: sgtm

dgoodwin:
Should DeprovisionedComputeMachinesets be getting set in here? Or how will that work for the new clusters?

pkg/controller/cluster/clusterapi_cluster_controller.go
csrwng: I assume it is ok because none of our linting tools catch it as a problem.

pkg/controller/controller_ref_manager.go
csrwng: I'll add our own

csrwng:
> Should DeprovisionedComputeMachinesets be getting set in here? Or how will that work for the new clusters?

No, it should be set in the controller that provisions them on the remote cluster.

PR #227, opened 2018-05-18
WIP: initial validation for clusterapi cluster object

joelddiaz:
use existing clusteroperator clusterspec validation on clusterapi cluster.spec.providerconfig


pkg/apis/clusterapi/validation/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+	spec := &clusteroperator.ClusterProviderConfigSpec{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ClusterProviderConfigSpec",
+			APIVersion: "clusteroperator.openshift.io/v1alpha1",
+		},
+		ClusterSpec: clusteroperator.ClusterSpec{
+			ClusterVersionRef: clusteroperator.ClusterVersionReference{
+				Name: "someclusterref",
+			},
+			MachineSets: []clusteroperator.ClusterMachineSet{
```
staebler: For a cluster-api Cluster, the MachineSets field in the ProviderConfig must be empty. The MachineSets are defined by the user outside of the Cluster object.

pkg/apis/clusterapi/validation/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return spec
+}
+
+func getValidClusterWithEmptyProviderConfig() *cluster.Cluster {
+	cluster := &cluster.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "test-cluster",
+		},
+		Spec: cluster.ClusterSpec{
```
staebler: This section should not be necessary. A nil value for `ProviderConfig.Value` is the default.

pkg/apis/clusterapi/validation/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+			},
+		},
+	}
+	return cluster
+}
+
+func getRawExtensionFromProviderConfig(providerConfig *clusteroperator.ClusterProviderConfigSpec) *runtime.RawExtension {
+	serializer := jsonserializer.NewSerializer(jsonserializer.DefaultMetaFactory, clusteroperatorapi.Scheme, clusteroperatorapi.Scheme, false)
+	var buffer bytes.Buffer
+	serializer.Encode(providerConfig, &buffer)
```
staebler: Don't swallow the error returned from `Encode`. It will make it more difficult to debug if there is a problem encoding.

pkg/apis/clusterapi/validation/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+func getValidClusterStrategy() *ClusterStrategy {
+	return &ClusterStrategy{}
+}
+
+// TestValidateCluster tests the Validate function.
+func TestValidateCluster(t *testing.T) {
+	cases := []struct {
+		name            string
+		clusterStrategy *ClusterStrategy
```
staebler: Take this out of the cases unless you are planning to have a subtest that uses an invalid ClusterStrategy--which doesn't make too much sense anyway since the ClusterStrategy is the component under test. If you are trying to have different ClusterStrategy behavior based on the contained StorageBuilder, then include the StorageBuilder in the cases, instead of the ClusterStrategy.

pkg/apis/clusterapi/validation/cluster.go
```diff
@@ -0,0 +1,76 @@
...
+type ClusterStrategy struct {
+	builders.StorageBuilder
+}
+
+var ValidateClusterName = apivalidation.ValidateClusterName
+
+// Validate checks that an instance of Cluster is well formed
+func (m ClusterStrategy) Validate(ctx request.Context, obj runtime.Object) field.ErrorList {
+	errors := field.ErrorList{}
+	//errors = append(errors, m.StorageBuilder.Validate(ctx, obj)...)
```
staebler: Is this something that you plan on uncommenting?

pkg/apis/clusterapi/validation/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+			cluster: func() *cluster.Cluster {
+				c := getValidClusterWithEmptyProviderConfig()
+				pc := getValidClusterProviderConfig()
+				pc.MachineSets[0].Size = 0
+				c.Spec.ProviderConfig.Value = getRawExtensionFromProviderConfig(pc)
+				return c
+			}(),
+			valid: false,
+		},
+	}
```
staebler: Include test cases that verify the behavior when the ProviderConfig is malformed and when the ProviderConfig uses a different type (e.g., some other APIVersion).

When the ProviderConfig uses a different type, we should not run any custom validation on the Cluster. We should only run the validation from the StorageBuilder.

pkg/apis/clusterapi/validation/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	ctx := request.NewContext()
+	if ctx == nil {
+		t.Errorf("problem creating context")
+	}
+	for _, tc := range cases {
```
staebler: Use `t.Run` to run each subtest. Then you don't have to include the subtest name in the calls to `t.Errorf`.

pkg/apis/clusterapi/validation/cluster_test.go
```diff
@@ -0,0 +1,153 @@
...
+				c.Spec.ProviderConfig.Value = getRawExtensionFromProviderConfig(pc)
+				return c
+			}(),
+			valid: false,
+		},
+	}
+
+	ctx := request.NewContext()
+	if ctx == nil {
+		t.Errorf("problem creating context")
```
staebler: You want to use `t.Fatalf` here. Otherwise, if something in the tests try to use the context, the tests will fail with a panic.

pkg/controller/controller_utils.go
```diff
@@ -348,12 +350,39 @@ func AddLabels(obj metav1.Object, additionalLabels map[string]string) {
 	}
 }
 
+// take raw providerconfig and return v1alpha1/ClusterProviderConfigSpec
```
staebler: Format this comment so that it is of this form.
```
clusterProviderConfigSpecFromProviderConfigValue takes a raw providerconfig ...
```

pkg/apis/clusterapi/validation/cluster.go
joelddiaz: no, this is some cruft from a previous bug that i uncovered while writing the test cases. i'll remove it in the next go-around

dgoodwin:
Reason was: because we control the creation of the Cluster via our ClusterDeployment, which is where we'll apply validation.

PR #228, opened 2018-05-21
Add cluster API and machine controller image refs to ClusterVersion

csrwng:
The original reason for this is so we can specify which images to use in CI. However, it also makes sense that we track these images per version.


csrwng:
@staebler ptal

PR #230, opened 2018-05-21
Parameterize images to deploy on target cluster

csrwng:
Allows passing variables to the deploy-cluster-api playbook to point to alternate images for the machine controller and cluster-api.

For now, the default location of these images will be the registry on api.ci where we'll update the latest tag every time we merge to master.

For PRs we will push to the same registry with a pr-specific tag and tell the target cluster to use that.

### DIFF
--- a/build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
+++ b/build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
@@ -93,7 +93,17 @@
       dest: "{{ temp_file.path }}"
 
   - name: process template
-    shell: "oc process -f {{ temp_file.path }} -p CLUSTER_API_NAMESPACE={{ cluster_api_namespace }} -p SERVING_CA={{ cluster_api_ca_bundle }} -p SERVING_CERT={{ cluster_api_cert }} -p SERVING_KEY={{ cluster_api_key }} -p BOOTSTRAP_KUBECONFIG='{{ bootstrap_kubeconfig.content }}' | oc apply -f -"
+    shell: |-
+      oc process -f {{ temp_file.path }} \
+         -p CLUSTER_API_NAMESPACE={{ cluster_api_namespace }} \
+         -p SERVING_CA={{ cluster_api_ca_bundle }} \
+         -p SERVING_CERT={{ cluster_api_cert }} \
+         -p SERVING_KEY={{ cluster_api_key }} \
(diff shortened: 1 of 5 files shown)


build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
```diff
@@ -92,8 +92,25 @@
...
   - name: process template
-    shell: "oc process -f {{ temp_file.path }} -p CLUSTER_API_NAMESPACE={{ cluster_api_namespace }} -p SERVING_CA={{ cluster_api_ca_bundle }} -p SERVING_CERT={{ cluster_api_cert }} -p SERVING_KEY={{ cluster_api_key }} -p BOOTSTRAP_KUBECONFIG='{{ bootstrap_kubeconfig.content }}' | oc apply -f -"
+    shell: |-
+      oc process -f {{ temp_file.path }} \
+         -p CLUSTER_API_NAMESPACE={{ cluster_api_namespace }} \
+         -p SERVING_CA={{ cluster_api_ca_bundle }} \
+         -p SERVING_CERT={{ cluster_api_cert }} \
+         -p SERVING_KEY={{ cluster_api_key }} \
+         -p BOOTSTRAP_KUBECONFIG='{{ bootstrap_kubeconfig.content }}' \
+         -p IMAGE='{{ l_cluster_api_image }}' \
```
staebler: Is there a way that these default image can be written in one places. As it is, there are 4 places where the default image is defined. I fear that the defaults will diverge.

build/cluster-operator-ansible/playbooks/cluster-api-prep/files/cluster-api-template.yaml
```diff
@@ -381,10 +383,12 @@ parameters:
 - name: SERVING_KEY
 # location of container image
 - name: IMAGE
```
staebler: Can we be more specific with this name here while we are changing things? I recommend `CLUSTER_API_IMAGE` to match with the label used in other files.

build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
csrwng: We really only need the default in the initial deployment playbook. I'll remove it from elsewhere.

csrwng:
comments addressed

staebler:
Needs rebase, but looks good.
/lgtm

csrwng:
rebased and squashed

PR #231, opened 2018-05-22
Integration tests using CAPI

staebler:
* Added integration tests using the cluster-api resources.
* Fixed flake where integration tests were occasionally failing due to missing deprovision job. See https://trello.com/c/cdGM9KVt.
* Use UPDATE to update cluster status since we cannot do a strategic merge of the ProviderStatus.
* Fixed a number of bugs with capi controllers.
* Moved cluster name label used on MachineSets to a common location.


dgoodwin:
Just needs conflicts resolved but lgtm.

PR #232, opened 2018-05-22
Check for cluster readiness in cluster controller

csrwng:
Now that most status flags go on the cluster and not on the master machineset, in some cases the cluster will not be updated to ready status until after a relist because machinesets are done updating before all flags on the cluster status are set. This moves the readiness check to the cluster controller, which queues the cluster on either cluster or machineset changes.


staebler:
Good change. This was on my list of things to do as part of the status changes. Thanks for taking care of this.
/lgtm

PR #234, opened 2018-05-23
Additional fixes for prod like deployments.

dgoodwin:
These changes were required to get deployed in our CI cluster.

Handle scenario where namespace is pre-existing and we do not have perms
to create it ourselves.

Fixed missing namespace on auth template.

Removed use of image pull policy "never" which we should not need.

### DIFF
--- a/contrib/ansible/deploy-playbook.yml
+++ b/contrib/ansible/deploy-playbook.yml
@@ -38,6 +38,12 @@
       cluster_operator_namespace: "{{ cli_cluster_operator_namespace }}"
     when: cli_cluster_operator_namespace is defined
 
+  - name: check if cluster operator namespace exists
+    command: "oc get namespace {{ cluster_operator_namespace }}"
+    failed_when: false
+    changed_when: false
+    register: namespace_exists_reg
+
   - name: create the cluster operator namespace
     kubectl_apply:
       definition:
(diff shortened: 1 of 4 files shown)


contrib/ansible/deploy-playbook.yml
```diff
@@ -38,13 +38,20 @@
       cluster_operator_namespace: "{{ cli_cluster_operator_namespace }}"
     when: cli_cluster_operator_namespace is defined
 
+  - name: check if cluster operator namespace exists
```
staebler: It's unfortunate that we have to go through these machinations. Isn't the main point of Ansible that it will handle the exists/not-exists for you?

contrib/examples/cluster-operator-roles-template.yaml
```diff
@@ -8,11 +8,10 @@
 # itself.
 #
 ########
-apiVersion: v1
+apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: cluster-operator-deploy-template
-  namespace: ${CLUSTER_OPERATOR_NAMESPACE}
```
staebler: Is removing this namespace needed because the template itself is being created, rather than just processed from a file?

contrib/ansible/deploy-playbook.yml
dgoodwin: The complication here is that we are leaning on kube to be idempotent with the request by using kubectl apply (which we should be doing), but we do not have permissions to create namespaces, so that apply fails regardless.

contrib/examples/cluster-operator-roles-template.yaml
dgoodwin: At one point the plan was to create the template but that did not pan out, it went back to being processed only as a file. We could probably restore namespace here if there's any reason to.

PR #236, opened 2018-05-24
Fix infra controller lookup of infra machineset size.

dgoodwin:
This was set up to look for a infra capi.MachineSet to get the size,
however we do not create this in the root cluster, so the infra job was
not spawning.

Instead it should use the co.Cluster.Spec.MachineSets just as we did
previously.

### DIFF
--- a/pkg/controller/infra/infra_controller.go
+++ b/pkg/controller/infra/infra_controller.go
@@ -85,10 +85,6 @@ func NewClusterOperatorController(
 			return clusterLister.Clusters(namespace).Get(name)
 		},
 		clusterInformer.Informer().HasSynced,
-		func(*clusteroperator.CombinedCluster) bool { return true },
-		func(cluster *clusteroperator.CombinedCluster) (int, error) {
-			return controller.GetClustopInfraSize(cluster)
-		},
 	)
 }
 
@@ -103,7 +99,6 @@ func NewClusterAPIController(
 	clusterapiClient clusterapiclientset.Interface,
 ) *Controller {
 	clusterLister := clusterInformer.Lister()
-	machineSetLister := machineSetInformer.Lister()
 	return newController(
 		clusterapi.SchemeGroupVersion.WithKind("Cluster"),
 		"capi-infra",
@@ -117,13 +112,6 @@ func NewClusterAPIController(
 			return clusterLister.Clusters(namespace).Get(name)
 		},
 		clusterInformer.Informer().HasSynced,
-		func(cluster *clusteroperator.CombinedCluster) bool {
-			_, err := controller.GetCAPIInfraMachineSet(cluster, machineSetLister)
-			return err == nil
-		},
-		func(cluster *clusteroperator.CombinedCluster) (int, error) {
-			return controller.GetCAPIInfraSize(cluster, machineSetLister)
-		},
 	)
 }
 
@@ -138,8 +126,6 @@ func newController(
 	addMachineSetInformerEventHandler func(cache.ResourceEventHandler),
 	getCluster func(namespace, name string) (metav1.Object, error),
 	clustersSynced cache.InformerSynced,
-	canGetInfraSize func(*clusteroperator.CombinedCluster) bool,
-	getInfraSize func(*clusteroperator.CombinedCluster) (int, error),
 ) *Controller {
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartLogging(glog.Infof)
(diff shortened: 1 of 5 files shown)


staebler:
We need to reevaluate where we are getting the infra size from then. I do not expect that the MachineSet config will be in the Cluster spec, as it is not relevant to the spec of the Cluster. I could see us adding an infra size field in the ProviderConfig of the spec.

Can we hold off on making this change until after we get a ClusterDeployment controller? Neither the present code or this fix is the permanent solution.

dgoodwin:
Integration tests updated, they weren't adding CO ClusterMachineSets, used these as the new test input as it was easier to build the capi.MachineSet off them than the inverse, and should position it better for the incoming refactoring to ClusterDeployment.

pkg/controller/infra/infra_controller.go
```diff
@@ -451,15 +432,12 @@ func (s *jobSyncStrategy) DoesOwnerNeedProcessing(owner metav1.Object) bool {
...
-	if !s.controller.canGetInfraSize(cluster) {
-		return false
-	}
-
 	return cluster.ClusterOperatorStatus.ProvisionedJobGeneration != cluster.Generation
 }
 
 func (s *jobSyncStrategy) GetJobFactory(owner metav1.Object, deleting bool) (controller.JobFactory, error) {
 	cluster, err := controller.ConvertToCombinedCluster(owner)
+	fmt.Printf("%v\n", *cluster.ClusterOperatorSpec)
```
staebler: Remove this.

test/integration/controller_test.go
```diff
@@ -435,26 +495,29 @@ func TestCAPIClusterCreate(t *testing.T) {
...
+							t.Fatalf("multiple master machinesets: %s and %s", masterMachineSet.Name, ms.Name)
 						}
-						masterMachinSet = ms
+						masterMachineSet = ms
 					}
 				}
-				if masterMachinSet == nil {
-					t.Fatalf("no master machineset")
-				}
+				coClusterSpec.MachineSets = append(coClusterSpec.MachineSets, clustopClusterMS)
```
staebler: The machinesets have already been added to `coClusterSpec.MachineSets`. This will add them again.

test/integration/controller_test.go
```diff
@@ -423,9 +474,18 @@ func TestCAPIClusterCreate(t *testing.T) {
...
 
 			machineSets := make([]*capiv1alpha1.MachineSet, len(tc.machineSets))
-			var masterMachinSet *capiv1alpha1.MachineSet
+			var masterMachineSet *capiv1alpha1.MachineSet
+
+			for i, clustopClusterMS := range tc.machineSets {
+
+				role := capicommon.NodeRole
+				if clustopClusterMS.NodeType == clustopv1alpha1.NodeTypeMaster {
+					role = capicommon.MasterRole
```
staebler: Set `masterMachineSet` here rather than looking for it by looping over the roles later.

test/integration/controller_test.go
```diff
@@ -423,9 +474,18 @@ func TestCAPIClusterCreate(t *testing.T) {
...
+				}
+
+				machineSet := testCAPIMachineSet(clustopClusterMS.ShortName, int32(clustopClusterMS.Size),
```
staebler: Use a name for the machineset that includes the cluster.
For example, `fmt.Sprintf("%s-%s", cluster.Name, clustopClusterMS.ShortName)`

test/integration/controller_test.go
```diff
@@ -435,26 +495,29 @@ func TestCAPIClusterCreate(t *testing.T) {
...
 			}
 
 			if err := waitForCAPIClusterToExist(capiClient, testNamespace, testClusterName); err != nil {
 				t.Fatalf("error waiting for Cluster to exist: %v", err)
 			}
 
-			for _, machineSet := range tc.machineSets {
-				if err := waitForCAPIMachineSetToExist(capiClient, testNamespace, machineSet.Name); err != nil {
-					t.Fatalf("error waiting for MachinSet %v to exist: %v", machineSet.Name, err)
+			for _, clustopClusterMS := range tc.machineSets {
```
staebler: Iterate over `machineSets` instead so that we can use the full name from the machine set.

test/integration/controller_test.go
dgoodwin: This doesn't quite work as we don't have the capi.MachineSet here, but we need the role to make the capi.MachineSet, I will just do a similar if below instead of iterating the roles.

dgoodwin:
Updated.

dgoodwin:
Updated again I missed the components controller.

PR #237, opened 2018-05-24
Strip out legacy cluster-operator resources

staebler:
* Removes clustop `Machine` and `MachineSet`.
* Removes all controllers that synced clustop resources. Some of this code may later be revived and re-purposed for ClusterDeployment, like the syncmachineset controller. I thought it would be better to remove it now and re-add it later rather than keep around dead code that may never be used again.
* Renames clustop `Cluster` to `ClusterDeployment`.


dgoodwin:
@joelddiaz is working on syncmachineset as we speak, can we keep that one alive with minimal effort to avoid the git conflicts?

dgoodwin:
Scratch that I can see that's a pain not worth doing, notified Joel on IRC that it's incoming.

pkg/apis/clusteroperator/v1alpha1/register.go
```diff
@@ -50,23 +50,17 @@ var (
...
-		&MachineList{},
 		&ClusterProviderConfigSpec{},
 		&ClusterProviderStatus{},
 		&MachineSetProviderConfigSpec{},
 	)
 	metav1.AddToGroupVersion(scheme, SchemeGroupVersion)
 	scheme.AddKnownTypes(schema.GroupVersion{Version: "v1"}, &metav1.Status{})
-	scheme.AddFieldLabelConversionFunc("clusteroperator.openshift.io/v1alpha1", "Cluster", ClusterFieldLabelConversionFunc)
-	scheme.AddFieldLabelConversionFunc("clusteroperator.openshift.io/v1alpha1", "MachineSet", MachineSetFieldLabelConversionFunc)
-	scheme.AddFieldLabelConversionFunc("clusteroperator.openshift.io/v1alpha1", "Node", MachineFieldLabelConversionFunc)
```
dgoodwin: Curious what Node was doing here, looks like it was supposed to be Machine perhaps.

pkg/clusterapi/aws/actuator.go
```diff
@@ -93,7 +93,7 @@ func NewActuator(kubeClient *kubernetes.Clientset, clusterClient *clusterclient.
 
 // Create runs a new EC2 instance
 func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
-	mLog := clustoplog.WithClusterAPIMachine(a.logger, machine)
+	mLog := clustoplog.WithMachine(a.logger, machine)
```
dgoodwin: :+1:

cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -329,22 +324,11 @@ var ControllersDisabledByDefault = sets.NewString()
...
 	controllers["nodeconfig"] = startNodeConfigController
 	controllers["deployclusterapi"] = startDeployClusterAPIController
-	controllers["syncmachineset"] = startSyncMachineSetController
-
-	controllers["capi-infra"] = startClusterAPIInfraController
-	controllers["capi-machine"] = startMachineAPIController
-	controllers["capi-master"] = startClusterAPIMasterController
-	controllers["capi-components"] = startClusterAPIComponentsController
-	controllers["capi-nodeconfig"] = startClusterAPINodeConfigController
-	controllers["capi-deployclusterapi"] = startClusterAPIDeployClusterAPIController
```
dgoodwin: Was hoping we'd revert these as well :+1:

dgoodwin:
FYI issue with actuator is just the loss of the controller populating what we expect there, so the actuator can't actuate. Will be fixed by @csrwng next patch I suspect, I'll figure out at workaround to keep testing.

pkg/apis/clusteroperator/v1alpha1/register.go
staebler: Yes, it was supposed to be Machine. It was a residual from when the resource was called Node at the very beginning. There are no conversions, so it was never noticed.

PR #239, opened 2018-05-25
Fix clusterapi template master role and provider config.

dgoodwin:
In the absence of the now delete machine controller, manually setup the
provider config. This just allows testing until the flip to generate
everything off a ClusterDeployment is ready.

Fixes case on the Master role to match the upstream constant, which was
causing control plane install jobs not to trigger as it couldn't find a
master machine set.

### DIFF
--- a/contrib/examples/clusterapi-template.yml
+++ b/contrib/examples/clusterapi-template.yml
@@ -141,16 +141,33 @@ objects:
     template:
       metadata:
         labels:
-          # TODO: enforce this in validation
           clusteroperator.openshift.io/machineset: ${CLUSTER_NAME}-master
           clusteroperator.openshift.io/cluster: ${CLUSTER_NAME}
       spec:
+        roles:
+          - Master
         providerConfig:
           value:
             apiVersion: clusteroperator.openshift.io/v1alpha1
             kind: MachineSetProviderConfigSpec
-        roles:
-          - master
+            clusterHardware:
+              aws:
+                accountSecret:
+                  name: ${CLUSTER_NAME}-aws-creds
+                keyPairName: ${SSH_KEYPAIR_NAME}
+                region: us-east-1
+                sshSecret:
+                  name: ${CLUSTER_NAME}-ssh-key
+                sshUser: centos
+                sslSecret:
+                  name: ${CLUSTER_NAME}-certs
+            hardware:
+              aws:
+                instanceType: t2.xlarge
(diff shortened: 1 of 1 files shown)


contrib/examples/clusterapi-template.yml
```diff
@@ -141,16 +141,40 @@ objects:
...
+                  name: ${CLUSTER_NAME}-ssh-key
+                sshUser: centos
+                sslSecret:
+                  name: ${CLUSTER_NAME}-certs
+            clusterVersionRef: {}
+            hardware:
+              aws:
+                instanceType: t2.xlarge
+            infra: false
             kind: MachineSetProviderConfigSpec
```
staebler: Bring this up to the top, right under apiVersion.

contrib/examples/clusterapi-template.yml
```diff
@@ -141,16 +141,40 @@ objects:
...
+                accountSecret:
+                  name: ${CLUSTER_NAME}-aws-creds
+                keyPairName: ${SSH_KEYPAIR_NAME}
+                region: us-east-1
+                sshSecret:
+                  name: ${CLUSTER_NAME}-ssh-key
+                sshUser: centos
+                sslSecret:
+                  name: ${CLUSTER_NAME}-certs
+            clusterVersionRef: {}
```
staebler: Remove the fields with empty values. The ones that I see are `clusterVersionRef`, `metadata`, `nodeLabels`, `nodeType`, and `size`.

contrib/examples/clusterapi-template.yml
```diff
@@ -141,16 +141,40 @@ objects:
...
       metadata:
         labels:
-          # TODO: enforce this in validation
           clusteroperator.openshift.io/machineset: ${CLUSTER_NAME}-master
           clusteroperator.openshift.io/cluster: ${CLUSTER_NAME}
       spec:
         providerConfig:
           value:
             apiVersion: clusteroperator.openshift.io/v1alpha1
+            # Manually populating, this will be set automatically based on our ClusterDeployment.
```
staebler: I don't think that this comment is really necessary. These are things that a user would have to supply themselves if they were to create a MachineSet manually. The **entire** `MachineSet` would be populated manually instead of automatically by a `ClusterDeployment`.

dgoodwin:
Updated, moved the roles up too as they should probably be highly visible.

PR #241, opened 2018-05-25
ClusterDeployment controller

csrwng:
Watches cluster deployments and creates a cluster-api cluster and master machineset for each

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -1,0 +1,771 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clusterdeployment
+
+import (
+	"bytes"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	kubeclientset "k8s.io/client-go/kubernetes"
+	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/golang/glog"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/openshift/cluster-operator/pkg/kubernetes/pkg/util/metrics"
+
+	clustercommon "sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
+	capi "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+	capiclientset "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset"
+	capiinformers "sigs.k8s.io/cluster-api/pkg/client/informers_generated/externalversions/cluster/v1alpha1"
+	capilisters "sigs.k8s.io/cluster-api/pkg/client/listers_generated/cluster/v1alpha1"
+
+	clustop "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
+	clustopclientset "github.com/openshift/cluster-operator/pkg/client/clientset_generated/clientset"
+	clustopinformers "github.com/openshift/cluster-operator/pkg/client/informers_generated/externalversions/clusteroperator/v1alpha1"
+	clustoplisters "github.com/openshift/cluster-operator/pkg/client/listers_generated/clusteroperator/v1alpha1"
+
+	"github.com/openshift/cluster-operator/pkg/controller"
+	clustoplog "github.com/openshift/cluster-operator/pkg/logging"
+)
+
+// controllerKind contains the schema.GroupVersionKind for this controller type.
+var controllerKind = clustop.SchemeGroupVersion.WithKind("ClusterDeployment")
+
+const (
+	controllerLogName = "clusterdeployment"
+
+	// versionMissingRegion indicates the cluster's desired version does not have an AMI defined for it's region.
+	versionMissingRegion = "VersionMissingRegion"
+
+	// versionHasRegion indicates that the cluster's desired version now has an AMI defined for it's region.
+	versionHasRegion = "VersionHasRegion"
+
+	// versionMissing indicates that the cluster's desired version does not yet exist
+	versionMissing = "VersionMissing"
+
+	// versionExists indicates that the cluster's desired version does exist
+	versionExists = "VersionExists"
+
+	clusterDeploymentLabel = "clusteroperator.openshift.io/cluster-deployment"
+
+	// Domain for service names in the cluster. This is not configurable for OpenShift clusters.
+	defaultServiceDomain = "svc.cluster.local"
+
+	// CIDR for service IPs. This is configured in Ansible via openshift_portal_net. However, it is not yet
+	// configurable via cluster operator.
+	defaultServiceCIDR = "172.30.0.0/16"
+
+	// CIDR for pod IPs. This is configured in Ansible via osm_cluster_network_cidr. However, it is not yet
+	// configurable via cluster operator.
+	defaultPodCIDR = "10.128.0.0/14"
+
+	machineSetNameLabel = "clusteroperator.openshift.io/machineset"
+)
+
+// NewController returns a new cluster deployment controller.
+func NewController(
+	clusterDeploymentInformer clustopinformers.ClusterDeploymentInformer,
+	clusterInformer capiinformers.ClusterInformer,
+	machineSetInformer capiinformers.MachineSetInformer,
+	clusterVersionInformer clustopinformers.ClusterVersionInformer,
+	kubeClient kubeclientset.Interface,
+	clustopClient clustopclientset.Interface,
+	capiClient capiclientset.Interface) *Controller {
+
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(glog.Infof)
+	// TODO: remove the wrapper when every clients have moved to use the clientset.
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(kubeClient.CoreV1().RESTClient()).Events("")})
+
+	if kubeClient != nil && kubeClient.CoreV1().RESTClient().GetRateLimiter() != nil {
+		metrics.RegisterMetricAndTrackRateLimiterUsage("clusteroperator_clusterdeployment_controller", kubeClient.CoreV1().RESTClient().GetRateLimiter())
+	}
+
+	logger := log.WithField("controller", controllerLogName)
+
+	c := &Controller{
+		kubeClient:    kubeClient,
+		clustopClient: clustopClient,
+		capiClient:    capiClient,
+		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "clusterdeployment"),
+		logger:        logger,
+	}
+
+	clusterDeploymentInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.addClusterDeployment,
+		UpdateFunc: c.updateClusterDeployment,
+		DeleteFunc: c.deleteClusterDeployment,
+	})
+	c.clusterDeploymentsLister = clusterDeploymentInformer.Lister()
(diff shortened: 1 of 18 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+func getMachineSetKey(machineSet *capi.MachineSet) string {
+	return fmt.Sprintf("%s/%s", machineSet.Namespace, machineSet.Name)
+}
+
+func getClusterKey(cluster *capi.Cluster) string {
+	return fmt.Sprintf("%s/%s", cluster.Namespace, cluster.Name)
+}
+
+func masterMachineSetName(clusterDeploymentName string) string {
+	return fmt.Sprintf("%s-master", clusterDeploymentName)
```
staebler: We should use the name from the `MachineSetConfig` instead of using "master".

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+	if versionNS == "" {
+		versionNS = clusterDeployment.Namespace
+	}
+
+	cv, err := c.clusterVersionsLister.ClusterVersions(versionNS).Get(
+		clusterDeployment.Spec.ClusterVersionRef.Name)
+	return cv, err
+}
+
+func getMachineSetKey(machineSet *capi.MachineSet) string {
```
staebler: Not used.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+	return fmt.Sprintf("%s/%s", machineSet.Namespace, machineSet.Name)
+}
+
+func getClusterKey(cluster *capi.Cluster) string {
```
staebler: Not used.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+	return fmt.Sprintf("%s/%s", cluster.Namespace, cluster.Name)
+}
+
+func masterMachineSetName(clusterDeploymentName string) string {
+	return fmt.Sprintf("%s-master", clusterDeploymentName)
+}
+
+func masterMachineSetConfig(clusterDeployment *clustop.ClusterDeployment) *clustop.MachineSetConfig {
```
staebler: Let's return an `ok` bool to indicate whether the master was found rather than returning nil when not found.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+			clustoplog.WithClusterDeployment(c.logger, clusterDeployment).Debugf("conflict creating cluster, skipping")
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("error creating cluster for cluster deployment %s/%s: %v", clusterDeployment.Namespace, clusterDeployment.Name, err)
+		}
+		return nil
+	}
+
+	// cluster exists, make sure it reflects the current cluster deployment spec
```
staebler: In the future, we should push upstream for maintaining a generation in the Cluster and MachineSet so that we can compare generations here to avoid unnecessarily doing expensive full comparisons here when the spec of the Cluster and ClusterDeployment haven't changed.

pkg/controller/controller_utils.go
```diff
@@ -298,6 +345,42 @@ func ClusterStatusFromClusterAPI(cluster *clusterapi.Cluster) (*clusteroperator.
 	return &status.ClusterDeploymentStatus, nil
 }
 
+// ClusterDeploymentSpecFromCluster extracts a ClusterDeploymentSpec from the ProviderConfig field of a cluster-api Cluster
+func ClusterDeploymentSpecFromCluster(cluster *clusterapi.Cluster) (*clusteroperator.ClusterDeploymentSpec, error) {
```
staebler: This is the same as `ClusterSpecFromClusterAPI`, except that this function returns an empty spec when the ProviderConfig is empty. I would argue that a Cluster (associated with a ClusterDeployment) with an empty ProviderConfig is an error. Either way, we should get rid of one of the functions.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+	if cluster.DeletionTimestamp != nil {
+		// on a restart of the controller manager, it's possible a cluster shows up in a state that
+		// is already pending deletion. Prevent the cluster from being a creation observation.
+		c.deleteCluster(cluster)
+		return
+	}
+
+	clusterDeployment, err := controller.ClusterDeploymentForCluster(cluster, c.clusterDeploymentsLister)
+	if err != nil {
+		clustoplog.WithCluster(c.logger, cluster).Errorf("error retrieving cluster deployment for cluster: %v", err)
```
staebler: This should only be an error if the Cluster has a clust-op ProviderConfig. We should ignore Clusters intended for other providers.

pkg/controller/controller_utils.go
```diff
@@ -357,8 +440,33 @@ func ClusterAPIMachineProviderConfigFromMachineSetSpec(machineSetSpec *clusterop
...
+
+	// Copy cluster hardware onto the provider config as well. Needed when deleting a cluster in the actuator.
+	msSpec.ClusterHardware = clusterDeploymentSpec.Hardware
+
+	return ClusterAPIMachineProviderConfigFromMachineSetSpec(msSpec)
+}
+
+// PopulateMachineSpec ensures that the MachineSetSpec we use for the machine spec provider config is fully populated with defaults and calculated values based on Cluster Operator specific logic.
+// This can be used both on individual machines, as well as on the MachineTemplateSpec used in MachineSets and MachineDeployments.
+func PopulateMachineSpec(machineSpec *clusterapi.MachineSpec, clusterDeploymentSpec *clusteroperator.ClusterDeploymentSpec, clusterVersion *clusteroperator.ClusterVersion, mLog log.FieldLogger) error {
```
staebler: We should delete this function. It isn't used. The Machine spec will not need to be populate from MachineSet and Cluster any more with the advent of ClusterDeployment. I meant to delete this with my PR but forgot.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+		}
+		return nil
+	}
+
+	// cluster exists, make sure it reflects the current cluster deployment spec
+	deploymentConfigSpec, err := controller.ClusterDeploymentSpecFromCluster(cluster)
+	if err != nil {
+		return fmt.Errorf("cannot retrieve a cluster deployment spec from existing cluster %s/%s: %v", cluster.Namespace, cluster.Name, err)
+	}
+	if !apiequality.Semantic.DeepEqual(deploymentConfigSpec, clusterDeployment.Spec) {
```
dgoodwin: deploymentConfigSpec might be worth a rename.

I'd consider a change to the spec worth a log message at info level.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+	machineSetName := masterMachineSetName(clusterDeployment.Name)
+	machineSet, err := c.machineSetsLister.MachineSets(clusterDeployment.Namespace).Get(machineSetName)
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("error retrieving master machineset %s/%s: %v", clusterDeployment.Namespace, machineSetName, err)
+	}
+
+	if machineSet == nil {
+		// machine set does not exist, it needs to be created
+		machineSet, err = buildMasterMachineSet(clusterDeployment, clusterVersion)
+		c.logger.Debugf("Just built machineset: %#v", machineSet)
```
dgoodwin: I'd be tempted to create or pass around a contextual logger to each of these methods.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+
+	// Set networking defaults
+	cluster.Spec.ClusterNetwork.ServiceDomain = defaultServiceDomain
+	cluster.Spec.ClusterNetwork.Pods.CIDRBlocks = []string{defaultPodCIDR}
+	cluster.Spec.ClusterNetwork.Services.CIDRBlocks = []string{defaultServiceCIDR}
+	return cluster, nil
+}
+
+// validateAWSRegion will ensure the cluster's version has an AMI defined for its region. If not a condition will
+// be added indicating the problem. Similarly if the problem clears up, the condition will be updated.
```
dgoodwin: Almost surely this is my bad but most of these comments state things that don't happen here, they happen up in the caller. Looks like the method is just validating and returning an error if there's a problem.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+		machineSet.Labels = make(map[string]string)
+	}
+	machineSet.Labels[clusterDeploymentLabel] = clusterDeployment.Name
+	machineSet.OwnerReferences = []metav1.OwnerReference{*metav1.NewControllerRef(clusterDeployment, controllerKind)}
+	machineSetLabels := map[string]string{"machineset": machineSet.Name}
+	machineSet.Spec.Selector.MatchLabels = machineSetLabels
+	replicas := int32(machineSetConfig.Size)
+	machineSet.Spec.Replicas = &replicas
+	machineSet.Spec.Template.Labels = machineSetLabels
+	machineSet.Spec.Template.Spec.Labels = machineSetLabels
```
dgoodwin: Looks like you also need to set the roles, it'll need to be "Master" as defined by the upstream constant and not "master". (hit that one myself)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+		cluster, ok = tombstone.Obj.(*capi.Cluster)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("Tombstone contained object that is not a Cluster %#v", obj))
+			return
+		}
+	}
+
+	clusterDeployment, err := controller.ClusterDeploymentForCluster(cluster, c.clusterDeploymentsLister)
+	if err != nil {
+		clustoplog.WithCluster(c.logger, cluster).Errorf("error retrieving cluster deployment for cluster: %v", err)
```
dgoodwin: This might be what you referred to in your question about deletion but I also saw that this triggers on normal clusterdeployment delete.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,720 @@
...
+	}
+
+	machineSet := &capi.MachineSet{}
+	machineSet.Name = masterMachineSetName(clusterDeployment.Name)
+	machineSet.Namespace = clusterDeployment.Namespace
+	machineSet.Labels = clusterDeployment.Labels
+	if machineSet.Labels == nil {
+		machineSet.Labels = make(map[string]string)
+	}
+	machineSet.Labels[clusterDeploymentLabel] = clusterDeployment.Name
```
dgoodwin: Should we also add a cluster label until we get something into upstream? My new AWS PR is relying on the cluster name label during delete. Perhaps I should be changing that to cluster deployment once this merges though.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
csrwng: @staebler I assume you mean the short name in the ClusterMachineSet. Afaik that name can change. Having a fixed name for the life of the cluster makes it easier to sync the machineset.

staebler: @csrwng We should either (1) require the short name in the ClusterMachineSet to be "master", (2) not allow changes to the short name for the master, or (3) not allow the user to supply a name for the master at all and reserve the "master" name so that it cannot be used for compute machinesets. Of those options, I like (3) the best.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
csrwng: We pass the cluster kind when retrieving the owner, so if an error happens, it's because the resource did have a valid owner reference to a clusterdeployment, and something happened retrieving it.

staebler: Good point.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,772 @@
...
+		machineSet, err = buildMasterMachineSet(clusterDeployment, cluster, clusterVersion)
+		clustoplog.WithMachineSet(logger, machineSet).Debugf("About to create machineset")
+		if err != nil {
+			return fmt.Errorf("error building machineSet from clusterDeployment %s/%s: %v", clusterDeployment.Namespace, clusterDeployment.Name, err)
+		}
+		_, err = c.capiClient.ClusterV1alpha1().MachineSets(clusterDeployment.Namespace).Create(machineSet)
+		if err != nil {
+			return fmt.Errorf("error creating machineSet %s/%s: %v", machineSet.Namespace, machineSet.Name, err)
+		}
+		c.logger.Debugf("Just built machineset: %#v", machineSet)
```
dgoodwin: Might be worth removing, it's a bit verbose in the logs.

csrwng:
Addressed comment and rebased

dgoodwin:
Tested this morning, I was surprised to see how close we're getting to delete working as well, deleting the deployment cleaned up machinesets and machines. CAPI cluster still remained because we don't launch an infra deprovision job yet, because we haven't set the flag that the remote machinesets are disabled, but this will be fixed soon.

PR #242, opened 2018-05-28
Update machine status and allow for master deletion in root cluster.

dgoodwin:
Update machine status with core IP and hostname details as well as
instance ID.

AWS actuator now assumes multiple instances may exist matching it's name
and clusterid, and attempts to reconcile the situation if it occurs.

Fixed issues with deleting masters in the root cluster by specifying the
incluster flag.


dgoodwin:
Feeling like it's probably getting close to time to invest in some mocked unit tests for the AWS code.

pkg/clusterapi/aws/actuator.go
```diff
@@ -367,35 +340,114 @@ func (a *Actuator) DeleteMachine(machine *clusterv1.Machine) error {
...
+		return fmt.Errorf("unable to obtain EC2 client: %v", err)
+	}
+
+	instances, err := a.getRunningInstances(machine, client)
+	if err != nil {
+		return err
+	}
+
+	if len(instances) == 0 {
+		// Parent controller should prevent this from ever happening by calling Exists and then Create.
```
staebler: This could happen if the instance was deleted by a third party in between the call to Exists and the call to Update.

pkg/clusterapi/aws/actuator.go
```diff
@@ -367,35 +340,114 @@ func (a *Actuator) DeleteMachine(machine *clusterv1.Machine) error {
...
+			return err
+		}
+	} else {
+		newestInstance = instances[0]
+	}
+
+	mLog = mLog.WithField("instanceID", *newestInstance.InstanceId)
+	mLog.Debug("instance found")
+
+	instance := instances[0]
```
staebler: Should this be `instance := newestInstance`?

pkg/clusterapi/aws/actuator.go
```diff
@@ -413,8 +465,118 @@ func (a *Actuator) Exists(machine *clusterv1.Machine) (bool, error) {
...
+
+// updateStatus calculates the new machine status, checks if anything has changed, and updates if so.
+func (a *Actuator) updateStatus(machine *clusterv1.Machine, instance *ec2.Instance, mLog log.FieldLogger) error {
+
+	mLog.Debug("updating status")
+
+	awsStatus := &cov1.AWSMachineProviderStatus{}
+	var err error
+	if machine.Status.ProviderStatus != nil {
+		awsStatus, err = controller.AWSMachineProviderStatusFromClusterAPIMachine(machine)
```
staebler: May as well have `AWSMachineProviderStatusFromClusterAPIMachine` return a defalut `AWSMachineProviderStatus` when `machine.Status.ProviderStatus` is nil.

contrib/examples/cluster-api-controllers-template.yaml
```diff
@@ -48,6 +48,7 @@ objects:
           args:
           - --log-level=debug
           - --default-availability-zone=${DEFAULT_AVAILABILITY_ZONE}
+          - --incluster=false
```
staebler: How will this work if we want to use cluster-operator in a remote cluster to build out more clusters from there?

pkg/clusterapi/aws/actuator.go
dgoodwin: Good catch, ty.

contrib/examples/cluster-api-controllers-template.yaml
dgoodwin: Oof thats going to be tough. Feels like it should be a cluster specific setting.

staebler: Nevermind. All of the Machine controllers that we run should use `incluster=false`. We will never be creating a control plane machineset in-cluster, even when a user in the remote cluster is creating more clusters. We should make sure that we are setting this flag in the ClusterAPI controller manager that we start in the remote cluster as well.

pkg/controller/controller_utils.go
```diff
@@ -357,6 +357,48 @@ func ClusterAPIMachineProviderConfigFromMachineSetSpec(machineSetSpec *clusterop
...
+	if err != nil {
+		return nil, err
+	}
+	// Encoding json includes a trailing line feed, which will not be present on objects
+	// read back from etcd. If we see the last byte as a line feed, remove it, allowing us
+	// to accurately compare new/old status.
+	rawBytes := buffer.Bytes()
+	if rawBytes[len(rawBytes)-1] == 10 {
+		rawBytes = rawBytes[:len(rawBytes)-1]
+	}
```
dgoodwin: PR updated, added a hack here, status was always showing a change because the freshly encoded raw extension gets a trailing newline from the json encoder, whereas this is stripped off when written to and read back from etcd.

dgoodwin: Just found bytes.TrimSpace, replacing.

contrib/examples/cluster-api-controllers-template.yaml
dgoodwin: Per IRC we actually might be, but in any case the incluster flag only blocks deletion of master machines, which we don't expect to exist there, so it seems safe and correct to leave the incluster controllers with this set to the default of true.

pkg/clusterapi/aws/actuator.go
```diff
@@ -413,8 +466,116 @@ func (a *Actuator) Exists(machine *clusterv1.Machine) (bool, error) {
...
+	mLog.Debug("instance exists")
+	return true, nil
+}
+
+// updateStatus calculates the new machine status, checks if anything has changed, and updates if so.
+func (a *Actuator) updateStatus(machine *clusterv1.Machine, instance *ec2.Instance, mLog log.FieldLogger) error {
+
+	mLog.Debug("updating status")
+
+	awsStatus, err := controller.AWSMachineProviderStatusFromClusterAPIMachine(machine)
```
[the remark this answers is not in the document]
dgoodwin: Will do a full overwrite. Thought we might have something else writing parts but at this point that's not the case.

pkg/clusterapi/aws/actuator.go
```diff
@@ -413,8 +466,116 @@ func (a *Actuator) Exists(machine *clusterv1.Machine) (bool, error) {
...
+		mLog.Errorf("error encoding AWS provider status: %v", err)
+		return err
+	}
+
+	machineCopy := machine.DeepCopy()
+	machineCopy.Status.ProviderStatus = awsStatusRaw
+
+	if !equality.Semantic.DeepEqual(machine.Status, machineCopy.Status) {
+		mLog.Info("machine status has changed, updating")
+		machineCopy.Status.LastUpdated = metav1.Now()
```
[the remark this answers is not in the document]
dgoodwin: Is this something we should pursue now, seems we need to fix upstream before we could.

Out of curiosity, how would that be handled in a sync loop if we did fix it upstream? Compare old/new in updateMachine handler and don't sync on status changes?

pkg/clusterapi/aws/actuator.go
```diff
@@ -426,33 +587,26 @@ func (a *Actuator) Exists(machine *clusterv1.Machine) (bool, error) {
 				Name:   aws.String("instance-state-name"),
 				Values: []*string{aws.String("running"), aws.String("pending")},
```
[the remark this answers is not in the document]
dgoodwin: Mmmm tough one, will give this some thought.

pkg/apis/clusteroperator/types.go
```diff
@@ -564,6 +564,36 @@ type MachineSetAWSHardwareSpec struct {
...
+// AWSMachineProviderStatus is the AWS specific provider status for a cluster.k8s.io Machine.
+type AWSMachineProviderStatus struct {
+	// +optional
+	metav1.TypeMeta
+	// +optional
+	metav1.ObjectMeta
+
+	// InstanceID is the AWS instance ID for this machine.
+	// +optional
+	InstanceID *string
```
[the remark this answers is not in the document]
staebler: I don't find it unusual to nest everything. But I am fine with leaving it as is.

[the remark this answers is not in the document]
dgoodwin: Filed an issue this morning to see if this would be acceptable right on MachineStatus: https://github.com/kubernetes-sigs/cluster-api/issues/234

pkg/controller/controller_utils.go
```diff
@@ -357,6 +357,45 @@ func ClusterAPIMachineProviderConfigFromMachineSetSpec(machineSetSpec *clusterop
...
+	serializer := jsonserializer.NewSerializer(jsonserializer.DefaultMetaFactory, api.Scheme, api.Scheme, false)
+	var buffer bytes.Buffer
+	err := serializer.Encode(awsStatus, &buffer)
+	if err != nil {
+		return nil, err
+	}
+	// Encoding json includes a trailing line feed, which will not be present on objects
+	// read back from etcd. Trim whitespace to prevent, allowing us to accurately check if status
+	// has changed with DeepEqual.
+	rawBytes := bytes.TrimSpace(buffer.Bytes())
```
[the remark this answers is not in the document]
dgoodwin: Added also for the provider config encodings, seemed like it's probably a good idea.

dgoodwin:
PR updated, I also added instance state to the status.

I deferred the stopped instances to another trello card, need to see how discussion goes on that and then will come back to it.

dgoodwin:
Rebased on master to fix the conflict with actuator method signatures.

pkg/clusterapi/aws/actuator.go
```diff
@@ -367,35 +339,114 @@ func (a *Actuator) DeleteMachine(machine *clusterv1.Machine) error {
...
+	}
+	if len(instances) == 0 {
+		mLog.Warnf("no instances found to delete for machine")
+		return nil
 	}
-	terminateInstancesResult, err := client.TerminateInstances(terminateInstancesRequest)
+
+	instanceIDs := []*string{}
+	for _, instance := range instances {
+		mLog.WithField("instanceID", *instance.InstanceId).Infof("deleting AWS instance")
```
csrwng: Remove. You're logging the same thing inside the terminateInstances function that gets called right after this.

pkg/clusterapi/aws/actuator.go
```diff
@@ -367,35 +339,114 @@ func (a *Actuator) DeleteMachine(machine *clusterv1.Machine) error {
...
+	}
+
+	// In very unusual circumstances, there could be more than one machine running matching this
+	// machine name and cluster ID. In this scenario we will keep the newest, and delete all others.
+	var newestInstance *ec2.Instance
+	if len(instances) > 1 {
+
+		// Determine which instance is the newest:
+		for _, instance := range instances {
+			mLog.WithFields(log.Fields{
```
csrwng: Thinking that this might be simpler with a sort on LaunchTime

pkg/clusterapi/aws/actuator.go
```diff
@@ -413,8 +464,100 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+	awsStatusRaw, err := controller.ClusterAPIMachineProviderStatusFromAWSMachineProviderStatus(awsStatus)
+	if err != nil {
+		mLog.Errorf("error encoding AWS provider status: %v", err)
+		return err
+	}
+
+	machineCopy := machine.DeepCopy()
+	machineCopy.Status.ProviderStatus = awsStatusRaw
+
+	if !equality.Semantic.DeepEqual(machine.Status, machineCopy.Status) {
```
csrwng: it may be more efficient to do `bytes.Equal(machine.Status.ProviderStatus.Raw, machineCopy.Status.ProviderStatus.Raw)`

pkg/clusterapi/aws/actuator.go
```diff
@@ -413,8 +464,100 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+
+func (a *Actuator) terminateInstances(client *ec2.EC2, instanceIDs []*string, mLog log.FieldLogger) error {
+	for _, instanceID := range instanceIDs {
+		mLog.WithField("instanceID", *instanceID).Info("terminating instance")
+	}
+
+	terminateInstancesRequest := &ec2.TerminateInstancesInput{
+		InstanceIds: instanceIDs,
+	}
+	_, err := client.TerminateInstances(terminateInstancesRequest)
```
csrwng: For a future update... I'm wondering if it makes sense to not return until we have verified that the provider instance has been actually updated to the terminated state. That way the machine API object can better reflect what's going on in the provider.

contrib/examples/cluster-api-controllers-template.yaml
```diff
@@ -92,6 +93,7 @@ objects:
           imagePullPolicy: ${IMAGE_PULL_POLICY}
           command:
           - "./controller-manager"
+          - --incluster=false
```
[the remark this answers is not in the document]
dgoodwin: It looks like that's the only place it's used, but it's still technically exposed as an option on their controller-manager binary. Without knowing how or when it'll get picked up in the future, would it be safer to leave it set false? I believe that setting is correct for how we're deploying here.

We should probably campaign to get the flag removed entirely upstream though ASAP.

pkg/clusterapi/aws/actuator.go
dgoodwin: Probably a good idea. I've got a followup which refactors this a bit for re-use in the AWS ELB controller, if it's ok I'll incorporate a sort there in next PR.

pkg/clusterapi/aws/actuator.go
dgoodwin: If I'm successful campaigning for IPs and DNS in the MachineStatus itself upstream, we would have to revert to this form to still catch the changes. Worth keeping as is? If not I can switch to bytes for now.

pkg/clusterapi/aws/actuator.go
dgoodwin: Is there a delay between the request and the state change? That did not occur to me. I can add a todo.

pkg/clusterapi/aws/actuator.go
dgoodwin: Removed.

pkg/clusterapi/aws/actuator.go
csrwng: Yup, that's fine

pkg/clusterapi/aws/actuator.go
csrwng: 👍

contrib/examples/cluster-api-controllers-template.yaml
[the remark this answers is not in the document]
csrwng: sgtm

PR #244, opened 2018-05-30
Vendor latest Cluster-API and testify

dgoodwin:
Cluster API recently merged a fix we need for successful machine deletion. This required a couple interface updates which have been applied as a separate commit.

Testify was a very old copy, the author has recently reappeared and been merging some fixes, one of which should improve test output in editors such as vim.


glide.yaml
```diff
@@ -24,3 +24,6 @@ import:
 - package: github.com/kubernetes-incubator/apiserver-builder
   version: ^1.9.0-alpha.3
 - package: sigs.k8s.io/cluster-api
+  version: master
```
staebler: Do we really need to explicitly specify master for the version? Isn't that the default?

dgoodwin: I don't know what's going on but specifying master appears to actually be dead wrong, it was giving me an Aug 17th commit for testify somehow. Removing it and I get the proper master version. Updating.

dgoodwin:
Updated.

glide.yaml
```diff
@@ -24,3 +24,4 @@ import:
 - package: github.com/kubernetes-incubator/apiserver-builder
   version: ^1.9.0-alpha.3
 - package: sigs.k8s.io/cluster-api
+- package: github.com/stretchr/testify
```
staebler: So the follow-on to the comment about not needed master is that we shouldn't need this line at in the glide.yaml. Glide should pick up this dependency automatically by inspecting the imports.

dgoodwin: Are we sure this doesn't happen once and then lock it? I'm not sure what's going on but we were stuck on a very old testify, sometime last year, and I'm sure we've run glide up since then.

dgoodwin: If I run glide up -v on master without these changes, testify jumps from Sep 2016 to Mar 2018. I have no idea how it's stayed at 2016 so long amidst our past updates. I'm inclined to list it explicitly as a result though.

dgoodwin: I think we should keep the explicit list, it sounds like glide gets really weird with your local ~/.glide/cache if you don't explicitly set what you want. Also sounds like even glide itself recommends switching to dep. https://github.com/Masterminds/glide/issues/592

staebler: It has stayed at Sep 2016 because at one point our glide.yaml was tied to v1.1.4. Running `glide cc` clears that up. It still does not use the HEAD of the master branch without explicitly specifying the repo in the glide.yaml because kube-openapi has a dependency on testify. I can live with explicitly including testify in the glide.yaml if there are truly things that we need from testify that are not there in the f6abca593680b2315d2075e0f5e2a9751e3f431a commit.

dgoodwin: Yes the commits I'm after are from earlier this year, f6abca593680b2315d2075e0f5e2a9751e3f431a appears to be from July 2017. 

Where did that commit ID originate from though? We were pinned to 69483b4bd14f5845b5a1e55bca19e954e827f1d0 previously.

staebler: https://github.com/openshift/cluster-operator/blob/6e2970097eb48421552dd8ffe17bd79bacb227d2/glide.yaml#L27

PR #245, opened 2018-05-30
WIP: machine set syncing

joelddiaz:
resurect syncmachineset controller to watch clusteroperator.ClusterDeployment objects and sync non-master machinesets to remote cluster


csrwng:
@joelddiaz apart from properly naming functions (s/syncMachineSet/syncClusterDeployment/), the main thing I see is that you need a sync similar to what was being done in the previous cluster controller:
https://github.com/openshift/cluster-operator/blob/449eaf3938cf0dd8bfc106ae326ec472a39aa30d/pkg/controller/cluster/cluster_controller.go#L621-L785

Basically you need to retrieve the remote machinesets and reconcile them with whatever the clusterDeployment says they should be. I would also use expectations in the same way that they are used in that controller. I wonder if it may not be easier to start with that controller instead of the previous syncMachineSet controller.

PR #247, opened 2018-06-01
Add controller to add masters to internal/external ELB.

dgoodwin:
Watches machines, and if they're masters adds them to internal/external ELB. This happens on every machine sync, the API call appears idempotent, however if we feel this is too much chatter I could modify to only re-try adding to the ELB after a certain amount of time has elapsed since successfully being added.

Also disables the check for deprovisioning remote machinesets as we do not
currently create them, and this blocks the infra deprovision job and
thus cluster deletion.

### DIFF
--- a/cmd/cluster-operator-controller-manager/app/controllermanager.go
+++ b/cmd/cluster-operator-controller-manager/app/controllermanager.go
@@ -63,6 +63,7 @@ import (
 	cov1alpha1 "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
 	clusteroperatorinformers "github.com/openshift/cluster-operator/pkg/client/informers_generated/externalversions"
 	"github.com/openshift/cluster-operator/pkg/controller"
+	"github.com/openshift/cluster-operator/pkg/controller/awselb"
 	"github.com/openshift/cluster-operator/pkg/controller/clusterdeployment"
 	"github.com/openshift/cluster-operator/pkg/controller/components"
 	"github.com/openshift/cluster-operator/pkg/controller/deployclusterapi"
@@ -331,6 +332,7 @@ func NewControllerInitializers() map[string]InitFunc {
 	controllers["nodeconfig"] = startNodeConfigController
 	controllers["deployclusterapi"] = startDeployClusterAPIController
 	controllers["clusterdeployment"] = startClusterDeploymentController
+	controllers["awselb"] = startAWSELBController
 
 	return controllers
 }
@@ -601,6 +603,18 @@ func startClusterDeploymentController(ctx ControllerContext) (bool, error) {
 		ctx.ClientBuilder.ClientOrDie("clusteroperator-clusterdeployment-controller"),
 		ctx.ClientBuilder.ClusterAPIClientOrDie("clusteroperator-clusterdeployment-controller"),
 	).Run(int(ctx.Options.ConcurrentClusterDeploymentSyncs), ctx.Stop)
+	return true, nil
+}
+
+func startAWSELBController(ctx ControllerContext) (bool, error) {
+	if !clusterAPIResourcesAvailable(ctx) {
+		return false, nil
+	}
+	go awselb.NewController(
+		ctx.ClusterAPIInformerFactory.Cluster().V1alpha1().Machines(),
+		ctx.ClientBuilder.KubeClientOrDie("clusteroperator-awselb-controller"),
+		ctx.ClientBuilder.ClientOrDie("clusteroperator-awselb-controller"),
+	).Run(int(ctx.Options.ConcurrentELBMachineSyncs), ctx.Stop)
 	return true, nil
 }
 
--- a/pkg/clusterapi/aws/actuator.go
+++ b/pkg/clusterapi/aws/actuator.go
@@ -27,13 +27,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/kubernetes"
+
 	capicommon "sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	clusterclient "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 
 	coapi "github.com/openshift/cluster-operator/pkg/api"
@@ -133,7 +132,7 @@ func (a *Actuator) CreateMachine(cluster *clusterv1.Cluster, machine *clusterv1.
 
 	region := clusterSpec.Hardware.AWS.Region
 	mLog.Debugf("Obtaining EC2 client for region %q", region)
-	client, err := a.ec2Client(coMachineSetSpec, machine.Namespace, region)
+	client, _, err := CreateAWSClients(a.kubeClient, coMachineSetSpec, machine.Namespace, region)
 	if err != nil {
 		return nil, fmt.Errorf("unable to obtain EC2 client: %v", err)
 	}
@@ -335,12 +334,12 @@ func (a *Actuator) DeleteMachine(machine *clusterv1.Machine) error {
 		return fmt.Errorf("machine does not contain AWS hardware spec")
 	}
 	region := coMachineSetSpec.ClusterHardware.AWS.Region
-	client, err := a.ec2Client(coMachineSetSpec, machine.Namespace, region)
+	client, _, err := CreateAWSClients(a.kubeClient, coMachineSetSpec, machine.Namespace, region)
 	if err != nil {
 		return fmt.Errorf("error getting EC2 client: %v", err)
 	}
 
-	instances, err := a.getRunningInstances(machine, client)
+	instances, err := GetRunningInstances(machine, client)
 	if err != nil {
 		return err
 	}
@@ -349,12 +348,7 @@ func (a *Actuator) DeleteMachine(machine *clusterv1.Machine) error {
 		return nil
 	}
 
-	instanceIDs := []*string{}
-	for _, instance := range instances {
-		instanceIDs = append(instanceIDs, instance.InstanceId)
-	}
-
-	return a.terminateInstances(client, instanceIDs, mLog)
+	return TerminateInstances(client, instances, mLog)
 }
 
 // Update attempts to sync machine state with an existing instance. Today this just updates status
@@ -380,12 +374,13 @@ func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 
 	region := clusterSpec.Hardware.AWS.Region
 	mLog.WithField("region", region).Debugf("obtaining EC2 client for region")
-	client, err := a.ec2Client(coMachineSetSpec, machine.Namespace, region)
+	client, _, err := CreateAWSClients(a.kubeClient, coMachineSetSpec, machine.Namespace, region)
 	if err != nil {
 		return fmt.Errorf("unable to obtain EC2 client: %v", err)
 	}
 
-	instances, err := a.getRunningInstances(machine, client)
+	instances, err := GetRunningInstances(machine, client)
+	mLog.Debug("found %d instances for machine", len(instances))
 	if err != nil {
 		return err
 	}
@@ -395,49 +390,24 @@ func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 		// but instance could be deleted between the two calls.
 		return fmt.Errorf("attempted to update machine but no instances found")
 	}
+	SortInstances(instances)
 
 	// In very unusual circumstances, there could be more than one machine running matching this
 	// machine name and cluster ID. In this scenario we will keep the newest, and delete all others.
-	var newestInstance *ec2.Instance
-	if len(instances) > 1 {
-
-		// Determine which instance is the newest:
-		for _, instance := range instances {
-			mLog.WithFields(log.Fields{
-				"instanceID": *instance.InstanceId,
-				"state":      *instance.State.Name,
-				"launchTime": *instance.LaunchTime,
-			}).Warn("found multiple instances for machine in pending/running state")
-
-			if newestInstance == nil {
-				newestInstance = instance
-				continue
-			}
-			newestInstance = chooseNewestInstance(newestInstance, instance)
-		}
-
-		// Cleanup all the older instances:
-		terminateInstanceIDs := []*string{}
-		for _, instance := range instances {
-			if *instance.InstanceId != *newestInstance.InstanceId {
-				terminateInstanceIDs = append(terminateInstanceIDs, instance.InstanceId)
-			}
-		}
-		err := a.terminateInstances(client, terminateInstanceIDs, mLog)
-		if err != nil {
-			return err
-		}
-	} else {
-		newestInstance = instances[0]
-	}
-
+	newestInstance := instances[0]
 	mLog = mLog.WithField("instanceID", *newestInstance.InstanceId)
 	mLog.Debug("instance found")
 
-	instance := newestInstance
+	if len(instances) > 1 {
+		err = TerminateInstances(client, instances[1:], mLog)
+		if err != nil {
+			return err
+		}
+
+	}
 
 	// We do not support making changes to pre-existing instances, just update status.
-	statusErr := a.updateStatus(machine, instance, mLog)
+	statusErr := a.updateStatus(machine, newestInstance, mLog)
 	if statusErr != nil {
 		return statusErr
 	}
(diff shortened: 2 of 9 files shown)


staebler:
> Also disables the check for deprovisioning remote machinesets as we do not
currently create them, and this blocks the infra deprovision job and
thus cluster deletion.

@dgoodwin Could you go into more detail about this? I understand that if we are not creating them, then there is no *need* to deprovision them. What is blocking though? I would hope that anything waiting for a deprovision would not be held up if there was no corresponding provision. If we are attempting a provision, but it is failing, then I would hope that the deprovision would succeed. I just want to make sure that we are not covering up a resiliency issue that may present itself later and be more difficult to ascertain.

staebler:
> This happens on every machine sync, the API call appears idempotent, however if we feel this is too much chatter I could modify to only re-try adding to the ELB after a certain amount of time has elapsed since successfully being added.

This is definitely going to be too much chatter. Making unnecessary remote calls to AWS is not going to scale, even if the calls are idempotent. I am fine if we want to leave it as is for now, but it is something that we are going to have to address at some point.

pkg/clusterapi/aws/actuator.go
```diff
@@ -395,49 +396,37 @@ func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
-			}
-			newestInstance = chooseNewestInstance(newestInstance, instance)
-		}
+	newestInstance := instances[0]
+	mLog = mLog.WithField("instanceID", *newestInstance.InstanceId)
+	mLog.Debug("instance found")
 
-		// Cleanup all the older instances:
+	if len(instances) > 1 {
 		terminateInstanceIDs := []*string{}
```
staebler: We know the size we need for `terminateInstanceIDs`, so we should create the slice with size already established.

pkg/clusterapi/aws/actuator.go
```diff
@@ -395,49 +396,37 @@ func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+					"instanceID": *instance.InstanceId,
+					"state":      *instance.State.Name,
+					"launchTime": *instance.LaunchTime,
+				}).Warn("cleaning up extraneous instance for machine")
 				terminateInstanceIDs = append(terminateInstanceIDs, instance.InstanceId)
 			}
 		}
-		err := a.terminateInstances(client, terminateInstanceIDs, mLog)
+
+		err = a.terminateInstances(client, terminateInstanceIDs, mLog)
```
staebler: In both places where `terminateInstances` is called, the caller is creating a slice of IDs from a slice of instances. Can we let the callee create the slice of IDs instead and just pass the slice of instances?

pkg/clusterapi/aws/actuator.go
```diff
@@ -395,49 +396,37 @@ func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+	mLog = mLog.WithField("instanceID", *newestInstance.InstanceId)
+	mLog.Debug("instance found")
 
-		// Cleanup all the older instances:
+	if len(instances) > 1 {
 		terminateInstanceIDs := []*string{}
-		for _, instance := range instances {
+		// Cleanup all older instances:
+		for _, instance := range instances[1:] {
 			if *instance.InstanceId != *newestInstance.InstanceId {
```
staebler: Is this check necessary?

dgoodwin:
As for the chatter I should probably do something similar to the jobs that rerun on a staggered interval, store generation + timestamp in the machine status as optional fields. (only relevant for master machines) Might be best as a followup though as this will help unblock Joel and let him test for real.

pkg/clusterapi/aws/actuator.go
```diff
@@ -523,7 +512,7 @@ func (a *Actuator) updateStatus(machine *clusterv1.Machine, instance *ec2.Instan
 	return nil
 }
 
-func (a *Actuator) terminateInstances(client *ec2.EC2, instanceIDs []*string, mLog log.FieldLogger) error {
+func (a *Actuator) terminateInstances(client ec2iface.EC2API, instanceIDs []*string, mLog log.FieldLogger) error {
```
staebler: Should we move this into the aws utils too?

pkg/clusterapi/aws/utils.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	return instances, nil
+}
+
+// CreateAWSClients creates clients for the AWS services we use, using either the cluster AWS credentials secret
+// if defined (i.e. in the root cluster), otherwise the IAM profile of the master where the
+// actuator will run. (target clusters)
+func CreateAWSClients(kubeClient kubernetes.Interface, mSpec *cov1.MachineSetSpec, namespace, region string) (ec2iface.EC2API, elbiface.ELBAPI, error) {
```
staebler: Let's have this create and return a struct that encapsulates the various AWS clients. Let's also create an interface for the other functions that the struct implements. Then, we can unit test the actuator with a mock.

pkg/clusterapi/aws/utils.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/aws/aws-sdk-go/service/elb/elbiface"
+
+	cov1 "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
+)
+
+// SortInstances sorts the given slice of instances based on their launch time. The first item in the
+// slice after sorting will be the most recently launched, and can be considered the definitive instance
+// for the machine, a caller may wish to terminate all others. This function should only be called with
+// running instances, not those which are stopped or terminated.
+func SortInstances(instances []*ec2.Instance) {
```
staebler: Is there any reason to fully sort the instances? All we want is the newest instance, right? Can we simply return the newest instance or the index in the slice for the newest instance?

pkg/clusterapi/aws/utils.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	SortInstances(instances)
+	return instances[0], true, nil
+}
+
+// GetRunningInstances returns all running instances that have a tag matching our machine name,
+// and cluster ID.
+func GetRunningInstances(machine *clusterv1.Machine, client ec2iface.EC2API) ([]*ec2.Instance, error) {
+	instances := []*ec2.Instance{}
```
staebler: Can we make a good guess on what the size of instances will be? Is there typically one instance per reservation, so that we could guess that the number of instances is equal to the number of reservations?

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -0,0 +1,313 @@
...
+
+// NewController returns a new *Controller.
+func NewController(machineInformer informers.MachineInformer, kubeClient kubeclientset.Interface, clustopClient clustopclient.Interface) *Controller {
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(glog.Infof)
+	// TODO: remove the wrapper when every clients have moved to use the clientset.
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(kubeClient.CoreV1().RESTClient()).Events("")})
+
+	if kubeClient != nil && kubeClient.CoreV1().RESTClient().GetRateLimiter() != nil {
+		metrics.RegisterMetricAndTrackRateLimiterUsage("clusteroperator_aws_elb_controller", kubeClient.CoreV1().RESTClient().GetRateLimiter())
```
staebler: "clusteroperator_awselb_controller" to stay consistent with the name of the controller being awselb.

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	c := &Controller{
+		client:     clustopClient,
+		kubeClient: kubeClient,
+		queue:      workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "aws_elb"),
```
staebler: s/aws_elb/awselb/

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -0,0 +1,313 @@
...
+		go wait.Until(c.worker, time.Second, stopCh)
+	}
+
+	<-stopCh
+}
+
+func (c *Controller) enqueue(machine *capiv1.Machine) {
+
+	// We only need to add masters to ELBs at this time.
+	if !controller.MachineHasRole(machine, capicommon.MasterRole) {
```
staebler: This check should be done before calling `enqueueMachine` so that a controller that has its `enqueue` function replaced retains the logic that the controller only acts on master machines.

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -0,0 +1,313 @@
...
+
+	mLog := clustoplog.WithMachine(c.logger, machine)
+
+	clusterID, ok := machine.Labels[clustopv1.ClusterNameLabel]
+	if !ok {
+		return fmt.Errorf("unable to lookup cluster for machine: %s", machine.Name)
+	}
+
+	// Deep-copy otherwise we are mutating our cache.
+	//m := machine.DeepCopy()
```
staebler: Remove unused code.

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	region := coMachineSetSpec.ClusterHardware.AWS.Region
+	mLog.Debugf("Obtaining AWS clients for region %q", region)
+	ec2Client, elbClient, err := clustopaws.CreateAWSClients(c.kubeClient, coMachineSetSpec, machine.Namespace, region)
+	if err != nil {
+		return err
+	}
+
+	instance, exists, err := clustopaws.GetInstance(machine, ec2Client)
```
staebler: I would rather have not found be an error, albeit a well-defined error. In this case, though, we don't care whether the error was not found or something more insidious. The behavior of the controller is the same in either case.

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -0,0 +1,313 @@
...
+const (
+	// maxRetries is the number of times a service will be retried before it is dropped out of the queue.
+	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the
+	// sequence of delays between successive queuings of a service.
+	//
+	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
+	maxRetries     = 15
+	controllerName = "awselb"
+
+	awsCredsSecretIDKey     = "awsAccessKeyId"
```
staebler: These aws creds consts are unused.

pkg/controller/infra/infra_controller.go
```diff
@@ -408,11 +408,14 @@ func (s *jobSyncStrategy) UpdateOwnerStatus(original, owner metav1.Object) error
 }
 
 func (s *jobSyncStrategy) CanUndo(owner metav1.Object) bool {
-	cluster, ok := owner.(*clusteroperator.CombinedCluster)
+	_, ok := owner.(*clusteroperator.CombinedCluster)
 	if !ok {
 		return false
 	}
-	return cluster.ClusterDeploymentStatus.DeprovisionedComputeMachinesets
+	// TODO: re-enable this check once joeldiaz work on provisioning machinesets to remote
```
staebler: Put this in a separate PR.

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	registerInput := elb.RegisterInstancesWithLoadBalancerInput{
+		Instances:        []*elb.Instance{{InstanceId: instance.InstanceId}},
+		LoadBalancerName: aws.String(elbName),
+	}
+
+	// This API call appears to be idempotent, so for now no need to check if the instance is
+	// registered first, we can just request that it be added.
+	_, err := elbClient.RegisterInstancesWithLoadBalancer(&registerInput)
+	if err != nil {
+		return err
```
staebler: It would be nice to have status in the machine that states that the elb registration failed, with a reason why.

cmd/cluster-operator-controller-manager/app/controllermanager.go
```diff
@@ -331,6 +332,7 @@ func NewControllerInitializers() map[string]InitFunc {
 	controllers["nodeconfig"] = startNodeConfigController
 	controllers["deployclusterapi"] = startDeployClusterAPIController
 	controllers["clusterdeployment"] = startClusterDeploymentController
+	controllers["awselb"] = startAWSELBController
```
staebler: Start the new controller in the integration tests, too.

pkg/clusterapi/aws/actuator.go
dgoodwin: Good idea.

pkg/clusterapi/aws/utils.go
dgoodwin: This was based on a request from Cesar in previous PR, after implementing it, it did seem a little simpler in the code when it comes time to figure out what instance to use, and obtain a slice to delete. This does however assume we're not getting into scenarios where we have thousands of instances existing for a single machine.

staebler: Whic are we trying to make simpler, the caller code or the function code? If it is the caller code, then can we have this function swap the newest instance to the front of the slice?

pkg/clusterapi/aws/utils.go
dgoodwin: A good guess would be one instance per reservation, I don't think anything we create would deviate from that, and I would *hope* that it's not possible to have a reservation return with no instances and we'd undershoot the size.

dgoodwin: Ah even then it's fine because it's a slice, will modify.

pkg/controller/awselb/aws_elb_controller.go
dgoodwin: Condition I assume you meant? (no condition arrays on any of these types yet) Or explicit fields on our provider status?

capi.MachineStatus has an error message/reason field but it's documented as something that should not be used for transient errors.

dgoodwin:
Updated for most of feedback not being deferred to followup.

I removed the actual sorting, instead SortInstances now returns the active instance and a list of all others as explicit return args, I prefer this to assumptions about positions in a single array.

Haven't exposed errors from the ELB add call yet, that may be better suited for a larger conversation and followup PR.

pkg/controller/awselb/aws_elb_controller.go
staebler: Yes, conditions would be the approach that I would be the most comfortable with. But however it is done, I would like for a user to be able to see that the master machine has been added to the elb or why it hasn't.

staebler: By the way, this can certainly be done as later work, too.

pkg/clusterapi/aws/utils.go
```diff
@@ -0,0 +1,206 @@
...
+
+	cov1 "github.com/openshift/cluster-operator/pkg/apis/clusteroperator/v1alpha1"
+)
+
+// SortInstances will examine the given slice of instances and return the current active instance for
+// the machine, as well as a slice of all other instances which the caller may want to terminate. The
+// active instance is calculated as the most recently launched instance.
+// This function should only be called with running instances, not those which are stopped or
+// terminated.
+func SortInstances(instances []*ec2.Instance) (*ec2.Instance, []*ec2.Instance) {
```
staebler: This is the approach that I like the very least. There is no reason that this function needs to return the second argument. It only adds a concern to the function that is orthogonal to what you want the function to do--which is to find the instance that is the most recent. I **highly** recommend that this function simply returns the index of the most recent instance. It is simple enough for the caller to do what it needs to do with that index.

```
newest := FindNewestInstance(instances)
instances[newest] = instances[len(instances)-1] 
instances = instances[:len(instances)-1]
```
or
```
newest := FindNewestInstance(instances)
instances = append(instances[:newest], instances[i+newest:]...)
```

dgoodwin: Your interpretation of what the function is meant to do is not the same as intended,  GetInstance below is for finding the instance that is most recent. I want this function to separate the active from the rest, for callers who need to act on both. I don't want a higher level caller to have to interpret positions in an array or do a bunch of slice logic especially when we're literally talking about destroying machines. The logic here is nicely isolated and very easy to unit test, and the caller gets back exactly what they need to act on.

I think we're way too far into the weeds here and I'm on about the 4th rewrite of this, we need to move this along. I am guessing your issue is with the allocation of the terminate slice unnecessarily for callers of GetInstance who literally just want the active, assuming some bizarre situations where there are thousands of instances matching the machine?

I don't want to push that slice logic up as high as the actuator because it gets harder to test, so I will add a probably unexported findNewestInstanceIndex, and use it in GetInstance. I'd like to keep SortInstances and use your implementation above and continue returning the separate arguments. I will document that you should not use sort instances for performance reasons unless you need to act on the inactive instances. Maybe rename to SortActiveFromInactive. Does this sound ok?

staebler: No, that is not my issue. My issue is that the `SortInstance` function is trying to do too many things--the smell being the multiple returns from the function.

Your argument that the logic is nicely isolated and very easy to unit test is inaccurate. In the end, we need unit tests for `GetInstance`, which will duplicate in many ways the same test cases and results in `TestSortInstances`. If we instead had a `GetInactiveInstances` function which returned only the inactive instances, then the logic for `GetActiveInstance` and `GetInactiveInstances` would be more isolated than the logic is currently for `SortInstances`. And their corresponding unit tests would be smaller and even easier to test.

The calling code is objectively simpler to read, too
```
newestInstance := GetActiveInstances(instances)
terminateInstances := GetInactiveInstances(instances)
if len(terminateInstances) > 0 {
  err = TerminateInstances(client, terminateInstances, mLog)
  if err != nil {
    return err
  }
}
```

With that being said, these are just my opinions and not something that I would hold up the PR for. If you are satisfied with the way the functions are laid out, then that is good enough for me.

dgoodwin: This last proposal doesn't look bad to me, my main goal was that the caller (currently the actuator) not have to play with array indicies and assumptions. However the separating the two in one function does seem ok for now. Thanks.

PR #249, opened 2018-06-02
Create a fake cluster operator environment by default

csrwng:
Adds a fake machine controller and introduces a variable that can be set on deployment to either use real or fake images.

Whether you use a real or fake cluster now depends on the `use_real_images` variable passed to the deploy-devel-playbook and not on which cluster_version is used to create the cluster.

### DIFF
--- a/cmd/fake-machine-controller/machine-controller.go
+++ b/cmd/fake-machine-controller/machine-controller.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"os"
+
+	"github.com/golang/glog"
+	"github.com/spf13/pflag"
+
+	log "github.com/sirupsen/logrus"
+
+	"k8s.io/apiserver/pkg/util/logs"
+
+	"github.com/kubernetes-incubator/apiserver-builder/pkg/controller"
+
+	"sigs.k8s.io/cluster-api/pkg/controller/config"
+	"sigs.k8s.io/cluster-api/pkg/controller/machine"
+	"sigs.k8s.io/cluster-api/pkg/controller/sharedinformers"
+
+	"github.com/openshift/cluster-operator/pkg/clusterapi/fake"
+)
+
+var (
+	defaultAvailabilityZone string
+	logLevel                string
+)
+
+const (
+	controllerLogName = "fakeMachine"
+	defaultLogLevel   = "info"
+)
+
+func init() {
+	config.ControllerConfig.AddFlags(pflag.CommandLine)
+	pflag.CommandLine.StringVar(&defaultAvailabilityZone, "default-availability-zone", "us-east-1c", "Default availability zone for machines created by this controller.")
+	pflag.CommandLine.StringVar(&logLevel, "log-level", defaultLogLevel, "Log level (debug,info,warn,error,fatal)")
+}
+
+func main() {
+	pflag.Parse()
+
+	logs.InitLogs()
+	defer logs.FlushLogs()
+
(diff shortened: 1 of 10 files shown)


cmd/fake-machine-controller/machine-controller.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/openshift/cluster-operator/pkg/clusterapi/fake"
+)
+
+var (
+	defaultAvailabilityZone string
+	logLevel                string
+)
+
+const (
+	controllerLogName = "awsMachine"
```
staebler: Should this be "fakeMachine"?

cmd/fake-machine-controller/machine-controller.go
```diff
@@ -0,0 +1,79 @@
...
+	defaultLogLevel   = "info"
+)
+
+func init() {
+	config.ControllerConfig.AddFlags(pflag.CommandLine)
+	pflag.CommandLine.StringVar(&defaultAvailabilityZone, "default-availability-zone", "us-east-1c", "Default availability zone for machines created by this controller.")
```
staebler: Is this command-line argument accepted so that scripts that are passing it in for the real AWS machine controller do not need to be modified? Or was it an oversight to retain it?

cmd/fake-machine-controller/machine-controller.go
csrwng: yup, will fix

cmd/fake-machine-controller/machine-controller.go
csrwng: Yes we need to have compatible flags so we can just replace the image in the deployment config template. Otherwise we have parameterize the args as well.

contrib/examples/cluster-api-controllers-template.yaml
```diff
@@ -41,10 +41,8 @@ objects:
         serviceAccountName: cluster-api-controller-manager
         containers:
         - name: machine-controller
-          image: cluster-operator:latest
+          image: " "
```
dgoodwin: What's the trick here, and if there is one could you add a comment on it. I'm confused where the fake machine controller imagestream is getting created, is it just implicit because we pushed an image?

contrib/ansible/deploy-playbook.yml
```diff
@@ -34,13 +34,21 @@
     cluster_api_image: "registry.svc.ci.openshift.org/openshift-cluster-operator/kubernetes-cluster-api:latest"
     cluster_api_image_pull_policy: "Always"
 
+    use_real_images: False
+    machine_controller_imagestream: "fake-machine-controller"
```
dgoodwin: I'm inclined to think that if you're calling deploy-playbook, use_real_images should default to true. That would only be done for real world deployments, whereas it would make sense to default to false for deploy-devel. Setting it when calling devel should override the default here.

dgoodwin: Could you also add a glaring entry in the README outlining the default behavior and how to work around it.

contrib/ansible/deploy-devel-playbook.yml
```diff
@@ -66,6 +74,8 @@
         CLUSTER_API_IMAGE_PULL_POLICY: "{{ cluster_api_image_pull_policy }}"
         MACHINE_CONTROLLER_IMAGE: "{{ machine_controller_image }}"
         MACHINE_CONTROLLER_IMAGE_PULL_POLICY: "{{ machine_controller_image_pull_policy }}"
+        ANSIBLE_IMAGE: "{{ ansible_image }}"
+        ANSIBLE_IMAGE_PULL_POLICY: "{{ ansible_image_pull_policy }}"
```
dgoodwin: It might be a good idea to use debug output to print a big WARNING on what type of deployment was used and what it means, to help folks catch if they did something unintentional.

contrib/ansible/deploy-devel-playbook.yml
```diff
@@ -25,9 +25,17 @@
     machine_controller_image: "registry.svc.ci.openshift.org/openshift-cluster-operator/cluster-operator:latest"
     machine_controller_image_pull_policy: "Always"
 
+    use_real_images: False
```
dgoodwin: What do you think about a rename for this var to something like "fake_deployment" and inverting the booleans? use_real_images could possibly slide under someone's visual radar, not realizing what that implies.

contrib/examples/cluster-api-controllers-template.yaml
csrwng: Yes, pushing an image to the integrated registry will create an imagestream

csrwng:
comments addressed

csrwng:
@dgoodwin bump

PR #250, opened 2018-06-04
Fix update of machineset replicas on clusterdeployment sync

csrwng:
A change in size of the master machineset config inside a clusterdeployment was not resulting in a change in replicas of the cluster API master machineset.


csrwng:
@staebler currently the size of the machineset is part of the providerconfig.

staebler:
@csrwng OK. That makes me feel better. I assume that eventually the size will be removed from the provider config, right? We should put in the replica comparison now in expectation of that eventuality--rather than hoping that we remember later.

csrwng:
Updated to compare providerConfig and replica size separately

PR #251, opened 2018-06-04
WIP: Add cluster reference to machine spec

staebler:
Add a local object reference in Machine spec that points to the Cluster to which the Machine belongs. Add validation that the Cluster reference is populated in every object that uses a Machine spec. Replace places where the cluster-name label had been used to now use the Cluster reference.

These changes pull in the latest from cluster-api and my cluster-api changes from https://github.com/staebler/cluster-api/commit/edabcdca84a66b34f8d0e100df8a787897d7e635. This PR is a WIP because changes need to go upstream first. I would appreciate any feedback here prior to submitting the PR upstream.


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -637,19 +637,18 @@ func buildMasterMachineSet(clusterDeployment *clustop.ClusterDeployment, cluster
 		machineSet.Labels = make(map[string]string)
 	}
 	machineSet.Labels[clusterDeploymentLabel] = clusterDeployment.Name
-	machineSet.Labels[clustop.ClusterNameLabel] = cluster.Name
 	machineSet.OwnerReferences = []metav1.OwnerReference{*metav1.NewControllerRef(clusterDeployment, controllerKind)}
 	machineSetLabels := map[string]string{
-		machineSetNameLabel:      machineSet.Name,
-		clusterDeploymentLabel:   clusterDeployment.Name,
-		clustop.ClusterNameLabel: cluster.Name,
```
dgoodwin: Where did we land on keeping this as convention for ease of use on the command line? I know we mentioned a selector on the field but I can't spot that in the cluster-api PR. I do kind of like being able to see at a glance what's going on with labels.

staebler: Yes, good catch. I did intend to add field selectors, but it slipped my mind. I will add them.

staebler:
I have closed the upstream cluster-api PR for this. In OpenShift 4.0, we will need master Machine objects to be in the remote cluster instead of the root cluster. So we no longer have a need to support multiple clusters in a single namespace.

PR #253, opened 2018-06-07
Fix master machine security groups and add AWS actuator create tests.

dgoodwin:
Actuator was still assuming all machines it created were nodes, and not assigning correct security groups for masters.

Split all our AWS client use into a wrapper object for testing purposes. Added mocks and unit tests for both the security groups, and the AWS actuator CreateMachine function. Will follow up with Update and Delete tests.


pkg/clusterapi/aws/actuator.go
```diff
@@ -542,3 +529,28 @@ func getBootstrapKubeconfig() (string, error) {
...
+	if !isMaster && !isInfra {
+		groupNames = append(groupNames, aws.String(vpcName+"_compute"))
+		groupNames = append(groupNames, aws.String(vpcName+"_compute_k8s"))
+	}
+
+	sgNameFilter := "group-name"
+	describeSecurityGroupsInput := ec2.DescribeSecurityGroupsInput{
+		Filters: []*ec2.Filter{
+			{Name: aws.String("vpc-id"), Values: []*string{&vpcID}},
+			{Name: &sgNameFilter, Values: groupNames},
```
staebler: For consistency with the rest of the function, use `aws.String("group-name")` here.

pkg/clusterapi/aws/actuator.go
```diff
@@ -542,3 +529,28 @@ func getBootstrapKubeconfig() (string, error) {
...
+		groupNames = append(groupNames, aws.String(vpcName+"_infra"))
+		groupNames = append(groupNames, aws.String(vpcName+"_infra_k8s"))
+	}
+	if !isMaster && !isInfra {
+		groupNames = append(groupNames, aws.String(vpcName+"_compute"))
+		groupNames = append(groupNames, aws.String(vpcName+"_compute_k8s"))
+	}
+
+	sgNameFilter := "group-name"
+	describeSecurityGroupsInput := ec2.DescribeSecurityGroupsInput{
```
staebler: Return here rather than storing in a temporary variable first.

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -0,0 +1,338 @@
...
+		},
+		{
+			name:               "compute",
+			expectedGroupNames: []string{"cn", "cn_compute", "cn_compute_k8s"},
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			input := buildDescribeSecurityGroupsInput("", "cn", tc.isMaster, tc.isInfra)
+			assert.Equal(t, len(tc.expectedGroupNames), len(input.Filters[1].Values))
```
staebler: Is the order of the filters important? Is this testing an implementation detail that the second filter is applied to the group name?

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -0,0 +1,338 @@
...
+			for i, sgn := range input.Filters[1].Values {
+				assert.Equal(t, tc.expectedGroupNames[i], *sgn)
```
staebler: Same thing here with the test requiring that the group names be in the filter in a specific order.

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -0,0 +1,338 @@
...
+	mockAWSClient.EXPECT().DescribeSubnets(&input).Return(&output, nil)
+}
+
+func addDescribeSecurityGroupsMock(mockAWSClient *MockAWSClient, vpcID, vpcName string, isMaster, isInfra bool) {
+	// Using the input builder from the production code. Behavior of this function is tested separately.
+	input := buildDescribeSecurityGroupsInput(vpcID, vpcName, isMaster, isInfra)
+	output := ec2.DescribeSecurityGroupsOutput{
+		SecurityGroups: []*ec2.SecurityGroup{},
+	}
+	// Using the explicit index where we store our name filters, may want to search for it if this
```
staebler: As noted in previous comments from earlier in the file, I think we should search explicitly for the group-name filters. I don't think that the amount of effort required justifies the fragility,

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -0,0 +1,338 @@
...
+				Namespace: testClusterVerNS,
+			},
+		},
+	}
+	return clusterDeployment
+}
+
+func testCluster() *capiv1.Cluster {
+	clusterDeployment := testClusterDeployment()
+	cluster, _ := clusterdeployment.BuildCluster(clusterDeployment)
```
staebler: Don't swallow this error. Pass in the `*testing.T` and fail the test if the cluster cannot be built.

staebler: I don't like creating a cluster indirectly from a clusterdeployment for these tests. I don't think that it's any more difficult to create the cluster directly, and then it removes a possible layer of complication when analysing why a test is failing. In the end, I expect that the `ProviderConfig` in cluster is going to be pretty lean, where the test will be filling out a lot more in clusterdeployment than it would if creating a cluster directly.

If you are going to use `BuildCluster`, move it to a more centrally accessible location rather than leaving it in the package of the clusterdeployment controller.

pkg/clusterapi/aws/client.go
```diff
@@ -0,0 +1,115 @@
...
+	DescribeSubnets(*ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error)
+	DescribeSecurityGroups(*ec2.DescribeSecurityGroupsInput) (*ec2.DescribeSecurityGroupsOutput, error)
+	RunInstances(*ec2.RunInstancesInput) (*ec2.Reservation, error)
+	DescribeInstances(*ec2.DescribeInstancesInput) (*ec2.DescribeInstancesOutput, error)
+	TerminateInstances(*ec2.TerminateInstancesInput) (*ec2.TerminateInstancesOutput, error)
+
+	RegisterInstancesWithLoadBalancer(*elb.RegisterInstancesWithLoadBalancerInput) (*elb.RegisterInstancesWithLoadBalancerOutput, error)
+}
+
+type RealAWSClient struct {
```
staebler: Does this need to be an exported type?

pkg/clusterapi/aws/client.go
```diff
@@ -0,0 +1,115 @@
...
+
+	RegisterInstancesWithLoadBalancer(*elb.RegisterInstancesWithLoadBalancerInput) (*elb.RegisterInstancesWithLoadBalancerOutput, error)
+}
+
+type awsClient struct {
+	ec2Client ec2iface.EC2API
+	elbClient elbiface.ELBAPI
+}
+
+func (c awsClient) DescribeImages(input *ec2.DescribeImagesInput) (*ec2.DescribeImagesOutput, error) {
```
staebler: It doesn't matter much in this case since the struct is small, but it's usually better to use a pointer receiver (i.e., `c *awsClient` instead of `c awsClient`).

pkg/clusterapi/aws/client.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (c awsClient) RegisterInstancesWithLoadBalancer(input *elb.RegisterInstancesWithLoadBalancerInput) (*elb.RegisterInstancesWithLoadBalancerOutput, error) {
+	return c.elbClient.RegisterInstancesWithLoadBalancer(input)
+}
+
+// NewAWSClient creates our client wrapper object for the actual AWS clients we use.
+// For authentication the underlying clients will use either the cluster AWS credentials
+// secret if defined (i.e. in the root cluster),
+// otherwise the IAM profile of the master where the actuator will run. (target clusters)
+func NewAWSClient(kubeClient kubernetes.Interface, mSpec *cov1.MachineSetSpec, namespace, region string) (Client, error) {
```
staebler: s/NewAWSClient/NewClient/, unless you are going to use awsClient as the return type.

pkg/clusterapi/aws/actuator_test.go
dgoodwin: I'm a bit torn on this one, I considered it at the time and just tried doing up a patch for it but it just doesn't feel right, we need a ClusterDeployment for the ProviderConfig right now either way, it feels wrong to duplicate what BuildCluster does and introduce more room for deviations when the logic is right there and easily accessible. We can evolve what it means to convert a ClusterDeployment to a capi.Cluster over time with less mass test updates. I can see how it's another thing to debug if something goes wrong, but with the requirement on a ClusterDeployment for the provider config anyhow right now, this seems like the lesser to me. If this is reasonable I'd like to keep re-using, and more the function somewhere more suitable.

dgoodwin:
All feedback addressed, left the BuildCluster change as a separate commit just in case.

pkg/clusterapi/aws/actuator_test.go
staebler: That's fine. I would like in the very near future to create separate types for the ProviderConfig in Cluster and Machine. My concern is that if this test is using `BuildCluster` now, then it always will, even after Cluster ProviderConfig is no longer taking a ClusterDeploymentSpec. But that's not terrible anyway.

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -0,0 +1,351 @@
...
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			input := buildDescribeSecurityGroupsInput("", "cn", tc.isMaster, tc.isInfra)
+			groupNamesFilter, err := findFilter(input.Filters, "group-name")
+			if assert.NoError(t, err) {
+				assert.Equal(t, len(tc.expectedGroupNames), len(groupNamesFilter.Values))
+				for i, sgn := range groupNamesFilter.Values {
+					assert.Equal(t, tc.expectedGroupNames[i], *sgn)
```
staebler: This is still requiring that the groups are in a particular order. I recommend creating a `[]string` from `groupNamesFilter.Values` and then simply calling `assert.ElementsMatch`.

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -0,0 +1,351 @@
...
+			expectedGroupNames: []string{"cn", "cn_infra", "cn_infra_k8s"},
+		},
+		{
+			name:               "compute",
+			expectedGroupNames: []string{"cn", "cn_compute", "cn_compute_k8s"},
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			input := buildDescribeSecurityGroupsInput("", "cn", tc.isMaster, tc.isInfra)
```
staebler: Also, why does this test not verify that the "vpc-id" filter exists and is valid and that there are no other filters?

dgoodwin:
OK that one was weird. Mock failures, not expecting calls that it should be suddenly, passing locally.

/retest

dgoodwin:
Will incorporate these in next one, already working on update tests, thanks!

dgoodwin:
This actually exposed a pretty nasty bug, we were leaking root CO bootstrap kubeconfigs to the target cluster masters. All user data looks to be only relevant for nodes, so this step is now skipped for all masters with tests to verify. Moved user data generation to a function passed to actuator. Also fixed the two test improvements from last comment.

dgoodwin:
/hold

Investigating if masters should get any user data at all, it appears to work without but https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_aws/templates/user_data.j2 indicates that Kenny's work was adding a minimal user data for masters.

dgoodwin:
/hold cancel

Should be good to go now.

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -187,10 +187,20 @@ func TestCreateMachine(t *testing.T) {
 			actuator.clientBuilder = func(kubeClient kubernetes.Interface, mSpec *clustopv1.MachineSetSpec, namespace, region string) (Client, error) {
 				return mockAWSClient, nil
 			}
+			var userDataGeneratorCalled bool
```
staebler: `userDataGeneratorCalled := false` instead of implicitly relying on false being the default.

pkg/clusterapi/aws/actuator.go
```diff
@@ -507,30 +500,70 @@ write_files:
   owner: 'root:root'
   permissions: '0640'
   content: |
-    openshift_group_type: %[1]s
+    openshift_group_type: {{ .NodeType }}{{ if .IsNode }}
```
staebler: For the sake of readability, but the control statements on separate lines.

For example, put the `{{ if .IsNode }}` on a separate line and use `{{- if .IsNode }}` to strip out the whitespace to the left.

dgoodwin: Ooo that's a nice tip thanks.

dgoodwin:
Updated.

pkg/clusterapi/aws/actuator.go
```diff
@@ -507,30 +500,73 @@ write_files:
...
 - path: /etc/origin/node/bootstrap.kubeconfig
   owner: 'root:root'
   permissions: '0640'
   encoding: b64
-  content: %[2]s
+  content: {{ .BootstrapKubeconfig }}
+{{- end }}
 runcmd:
 - [ ansible-playbook, /root/openshift_bootstrap/bootstrap.yml]
+{{ if .IsNode -}}
```
staebler: You really want this to strip the whitespace from the left instead of the right. It doesn't impact the result here, but it would make a big difference if the next line starts with spaces or tabs.

pkg/clusterapi/aws/actuator.go
```diff
@@ -507,30 +500,73 @@ write_files:
...
 - [ systemctl, restart, systemd-hostnamed]
 - [ systemctl, restart, NetworkManager]
 - [ systemctl, enable, origin-node]
 - [ systemctl, start, origin-node]
-`
+{{ end }}`
```
staebler: For consistency, you should strip whitespace from the left for this one, too. If you want to keep the newline that existed before, then move the tick to the next line.

dgoodwin:
Updated the whitespace trimming.

PR #255, opened 2018-06-08
Use clusterID instead of name for identifying the cluster in the cloud.

staebler:
* Set a unique cluster ID when the clusterdeployment is created.
* Trim the elb names down to a cap of 32 characters.

If https://github.com/kubernetes-sigs/cluster-api/pull/288 is accepted, then the endings of the elb names can be shrunk down a few characters to make room for a longer portion of the cluster name in the elb.


dgoodwin:
Creating a cluster deployment will generate a cluster ID like "dgoodwin-7m78k", and use this in the names for ELBs. This will allow for multiple namespaces to use the same name for a cluster deployment without conflicting. Presumably we use this for DNS names as well, does this sound ok to you @twiest @joelddiaz ?

pkg/ansible/generate_test.go
```diff
@@ -106,11 +108,28 @@ func TestGenerateClusterWideVars(t *testing.T) {
...
+			clusterSpec:    testClusterSpec(),
+			infraSize:      2,
+			clusterVersion: testClusterVersion(),
+			shouldInclude: []string{
+				"openshift_aws_clusterid: 012345678901234567890123456789-abcde",
+				"openshift_aws_elb_master_external_name: 0123456789-abcde-master-external",
+				"openshift_aws_elb_master_internal_name: 0123456789-abcde-master-internal",
+				"openshift_aws_elb_infra_name: 0123456789-abcde-infra",
+				// TODO: Use these instead when the ansible playbook is updated to support it.
+				// "openshift_aws_elb_master_external_name: 0123456789012345678-abcde-cp-ext",
```
dgoodwin: Whats the issue with the ansible playbook, and is that in progress or something we need to request or add to get done?

twiest:
@dgoodwin I'm fine with that scheme. Yes, the clusterid is used as part of DNS, and that DNS also has a random "shard id", so we will have multiple random parts of the name.

So, for instance, the wildcard DNS entry for a cluster might be:
*.dgoodwin-7m78k.4c9d.openshift.com

We need the random router shard as OpenShift dedicated clusters can (and many do) have multiple router shards.

Also, please note that ELBs have a maximum of [32 characters](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_CreateLoadBalancer.html) that can be used for the name, so we need to ensure that the entire clusterid is less than that.

pkg/clusterapi/aws/actuator.go
```diff
@@ -492,9 +492,12 @@ func (a *Actuator) updateStatus(machine *clusterv1.Machine, instance *ec2.Instan
...
 
-func getClusterID(machine *clusterv1.Machine) (string, bool) {
-	clusterID, ok := machine.Labels[cov1.ClusterNameLabel]
-	return clusterID, ok
+func getClusterID(machine *clusterv1.Machine) (string, error) {
+	coMachineSetSpec, err := controller.MachineSetSpecFromClusterAPIMachineSpec(&machine.Spec)
+	if err != nil {
+		return "", err
+	}
+	return coMachineSetSpec.ClusterID, nil
```
dgoodwin: This appears broken as the cluster ID on the machine's provider config is an empty string. This is causing the get runinng instances to find nothing and infinitely create vms, until we hit the account limit.

pkg/ansible/generate_test.go
staebler: There is a pending PR that is approved but waiting for the 3.10 branch.

staebler:
> Also, please note that ELBs have a maximum of 32 characters that can be used for the name, so we need to ensure that the entire clusterid + shard name is less than that.

@twiest We are not currently including the shard name in the ELB name. I am limiting the ELB name to 32 characters, however. In cases where the ELB name would be too long, the cluster name portion of the ELB name is truncated.

PR #256, opened 2018-06-11
Add UpdateMachine tests for AWS actuator.

dgoodwin:


pkg/clusterapi/aws/actuator.go
```diff
@@ -344,44 +344,50 @@ func (a *Actuator) DeleteMachine(machine *clusterv1.Machine) error {
...
+func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
+	_, err := a.UpdateMachine(cluster, machine)
+	return err
+}
+
+// UpdateMachine attempts to sync machine state with an existing instance. Today this just updates status
 // for details that may have changed. (IPs and hostnames) We do not currently support making any
 // changes to actual machines in AWS. Instead these will be replaced via MachineDeployments.
-func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
+func (a *Actuator) UpdateMachine(cluster *clusterv1.Cluster, machine *clusterv1.Machine) (*clusterv1.Machine, error) {
```
staebler: Are we returning the machine just so that we can inspect the updated machine in the tests? Or does it truly make sense to return the machine, from an API standpoint? If it is the former, then we could get the machine from the fake client action instead in the tests. Arguably, that is the side effect that you want to test anyway.

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -318,6 +432,51 @@ func addDescribeImagesMock(mockAWSClient *MockClient, imageID string) {
 		}, nil)
 }
 
+func testInstance(instanceID, machineName, hostType, instanceState, clusterName string, age time.Duration) *ec2.Instance {
```
staebler: s/clusterName/clusterID/ to not confuse the two.

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -318,6 +432,51 @@ func addDescribeImagesMock(mockAWSClient *MockClient, imageID string) {
...
+		PublicIP:   aws.String(fmt.Sprintf("%s-publicip", instanceID)),
+		PrivateIP:  aws.String(fmt.Sprintf("%s-privateip", instanceID)),
+		PublicDNS:  aws.String(fmt.Sprintf("%s-publicdns", instanceID)),
+		PrivateDNS: aws.String(fmt.Sprintf("%s-privatednf", instanceID)),
+	}
+}
+
+func addDescribeInstancesMock(mockAWSClient *MockClient, instances []*ec2.Instance) {
+	// Wrap each instance in a reservation:
+	reservations := make([]*ec2.Reservation, 0, len(instances))
```
staebler:
```
reservations := make([]*ec2.Reservation, len(instances))
for i, inst := range instances {
	reservations[i] = &ec2.Reservation{Instances: []*ec2.Instance{inst}}
}
```

pkg/clusterapi/aws/actuator_test.go
```diff
@@ -295,6 +301,114 @@ func TestCreateMachine(t *testing.T) {
...
+		instances          []*ec2.Instance
+		currentStatus      *clustopv1.AWSMachineProviderStatus
+		expectedInstanceID string
+		// expectedUpdate should be true if we expect a cluster-api client update to be executed for changing machine status.
+		expectedUpdate bool
+	}{
+		{
+			name: "one instance running no status change",
+			instances: []*ec2.Instance{
+				testInstance("i1", testMachineName, "master", "running", testClusterDeploymentName, 30*time.Minute),
```
staebler: This should use `testClusterID` instead of `testClusterDeploymentName`.

pkg/clusterapi/aws/actuator.go
dgoodwin: Mostly yes, but it also matched the pattern used for Create and Delete. I'm ok with switching it to get it from the fake client action though, will update.

dgoodwin:
Updated.

PR #257, opened 2018-06-11
WIP: Run master install whenever new machines need it

csrwng:
The master controller now generates an inventory based on which machines are available at the time the controller runs. It uses the Status.Versions.ControlPlane field to keep track of which Machines have had the master job run on them.

### DIFF
--- a/pkg/controller/clusterinstall/controller.go
+++ b/pkg/controller/clusterinstall/controller.go
@@ -68,7 +68,6 @@ func NewController(
 	playbooks []string,
 	clusterInformer capiinformers.ClusterInformer,
 	machineSetInformer capiinformers.MachineSetInformer,
-	machineInformer capiinformers.MachineInformer,
 	jobInformer batchinformers.JobInformer,
 	kubeClient kubeclientset.Interface,
 	clustopClient clustopclientset.Interface,
@@ -113,15 +112,6 @@ func NewController(
 		UpdateFunc: c.updateMachineSet,
 		DeleteFunc: c.deleteMachineSet,
 	})
-
-	if machineInformer != nil {
-		machineInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-			AddFunc:    c.addMachine,
-			UpdateFunc: c.updateMachine,
-			DeleteFunc: c.deleteMachine,
-		})
-		c.machinesSynced = machineInformer.Informer().HasSynced
-	}
 
 	jobOwnerControl := &jobOwnerControl{controller: c}
 	c.jobControl = controller.NewJobControl(controllerName, clusterKind, kubeClient, jobInformer.Lister(), jobOwnerControl, logger)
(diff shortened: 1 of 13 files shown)


pkg/controller/clusterinstall/controller.go
```diff
@@ -68,6 +68,7 @@ func NewController(
 	playbooks []string,
 	clusterInformer capiinformers.ClusterInformer,
 	machineSetInformer capiinformers.MachineSetInformer,
+	machineInformer capiinformers.MachineInformer,
```
staebler: Can you add to the comment for this function indicating that `machineInformer` can be nil if machine changes to not necessitate resyncing the cluster? Or maybe take this out of the `NewController` function and have the master controller set this in a subsequent step after creating the `clusterinstall.Controller`.

pkg/controller/clusterinstall/controller.go
```diff
@@ -353,6 +429,15 @@ func isMasterMachineSet(machineSet *capi.MachineSet) bool {
 	return false
 }
 
+func isMasterMachine(machine *capi.Machine) bool {
```
staebler: `controller.MachineHasRole(capicommon.MasterRole)`

pkg/controller/master/master_controller.go
```diff
@@ -157,16 +182,114 @@ func (s *installStrategy) setupServiceAccountForJob(clusterNamespace string) (*k
...
+	for _, machine := range allMachines {
+		if !metav1.IsControlledBy(machine, masterMachineSet) ||
+			machine.DeletionTimestamp != nil {
+			continue
+		}
+		hasPublicName, err := hasPublicDNS(machine)
+		if err != nil {
+			return nil, fmt.Errorf("error checking whether machine %s/%s has public DNS: %v", machine.Namespace, machine.Name, err)
+		}
+		if !hasPublicName {
```
staebler: If one of the masters does not have a public name yet, should we wait for it before installing the control plane? It would seem to me to be akin to waiting for all the replicas to be created.

pkg/controller/master/master_controller.go
```diff
@@ -157,16 +182,114 @@ func (s *installStrategy) setupServiceAccountForJob(clusterNamespace string) (*k
...
+	}
+	// Only start an install if we have the number of replicas specified in the machineset
+	if len(machines) < int(*masterMachineSet.Spec.Replicas) {
+		return false
+	}
+	// Of the available machines, determine if any has not had its control plane installed
+	// If all of them have a control plane, then no need to run this job.
+	// TODO: determine how to handle upgrades, if they require a job.
+	for _, m := range machines {
+		if m.Status.Versions == nil || len(m.Status.Versions.ControlPlane) == 0 {
```
staebler: We also need to clear `ControlPlane` in the actuator when the instance is created, to handle the case where the instance was deleted externally and had to be re-created.

pkg/controller/master/master_controller.go
```diff
@@ -157,16 +182,114 @@ func (s *installStrategy) setupServiceAccountForJob(clusterNamespace string) (*k
...
+		if err != nil {
+			errs = append(errs, fmt.Errorf("error retrieving machine %s to update the control plane version", machineKey))
+		}
+		updatedMachine := machine.DeepCopy()
+		if updatedMachine.Status.Versions == nil {
+			updatedMachine.Status.Versions = &capi.MachineVersionInfo{}
+		}
+		updatedMachine.Status.Versions.ControlPlane = machine.Spec.Versions.ControlPlane
+		updatedMachine.Status.LastUpdated = metav1.Now()
+		_, err = s.capiClient.ClusterV1alpha1().Machines(machine.Namespace).UpdateStatus(updatedMachine)
```
staebler: I am still a bit squeamish about having a controller update two different resources in a single sync. It is fine for now, but I'd like us to keep thinking about whether we can do better somehow.

pkg/controller/master/master_controller.go
csrwng: We would wait, because the machine would not be included in the list of available machines and those would not match the replica count of the machineset.

csrwng:
comments addressed, rebasing

csrwng:
Closing this as the master will need to manage its own nodes in the 4.0 setup.

PR #260, opened 2018-06-12
Remove AWS credentials before syncing cluster/machinesets

csrwng:
Also use clusterID as a prefix to remote machineset names. Using clusterID for the cluster will be done in a separate pull.


csrwng:
@joelddiaz ptal

PR #261, opened 2018-06-12
Only attempt to re-add masters to ELBs after period of time.

dgoodwin:
Rather than spamming AWS to re-add masters to their ELBs every time we
sync a master machine (when it likely would never be required again),
only do so if some period of time has elapsed since we last successfully
did so.


pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -265,11 +279,24 @@ func (c *Controller) syncMachine(key string) error {
 
 	region := coMachineSetSpec.ClusterHardware.AWS.Region
 	mLog.Debugf("Obtaining AWS clients for region %q", region)
-	client, err := clustopaws.NewClient(c.kubeClient, coMachineSetSpec, machine.Namespace, region)
+	client, err := c.clientBuilder(c.kubeClient, coMachineSetSpec, machine.Namespace, region)
```
staebler: Let's push building this client until later in the function when we've determined that it is actually needed.

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -265,11 +279,24 @@ func (c *Controller) syncMachine(key string) error {
...
+	if err != nil {
+		return err
+	}
+
+	status, err := controller.AWSMachineProviderStatusFromClusterAPIMachine(machine)
 	if err != nil {
 		return err
 	}
 
+	if status.LastELBSync != nil {
```
staebler: Is there anything else about the Machine that we should look at to determine whether an ELB sync may be necessary? For example, if the spec of the Machine changes, then I would expect the controller to sync the ELB even if the resync duration has not elapsed.

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -265,11 +279,24 @@ func (c *Controller) syncMachine(key string) error {
...
+		delta := time.Now().Sub(status.LastELBSync.Time)
+		if delta < elbResyncDuration {
+			mLog.WithField("delta", delta).Debugf("no resync needed")
+			return nil
+		}
+	}
+
```
staebler: We're missing the part where `LastELBSync` is set after successfully syncing the ELB. We should probably add a check for that in the unit tests.

dgoodwin: Yup, that actually exposes a problem with what we just decided to do in the aws actuator where we just rebuild status from nothing and assume nothing else is writing to it. I will get that fixed as well.

pkg/controller/awselb/aws_elb_controller.go
dgoodwin: The only things I can think of are (1) an admin editing a label to tickle the CO into re-adding it, and (2) the instance ID changes in the status indicating a new vm was picked up. As far as the spec goes the only thing I can think of is a machine getting the master role added after the fact, but I'm not sure we'd even want to start down that path.

I could base it on generation though as we do elsewhere, but that wouldn't detect status changes like instance ID.

dgoodwin:
Updated with several changes per commit messages.

dgoodwin:
Should be ready for another round of review if anyone has time.

pkg/clusterapi/aws/actuator.go
```diff
@@ -441,10 +441,22 @@ func (a *Actuator) updateStatus(machine *clusterv1.Machine, instance *ec2.Instan
...
+	if err != nil {
+		return err
+	}
+	// Save this, we need to check if it changed later.
+	origInstanceID := awsStatus.InstanceID
 
 	// Instance may have existed but been deleted outside our control, clear it's status if so:
-	if instance != nil {
+	if instance == nil {
+		awsStatus.InstanceID = nil
```
staebler: Can we add a test case to `TestUpdate` that verifies that these fields are being cleared when there is no longer an instance?

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -56,10 +60,16 @@ const (
 	// 5ms, 10ms, 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1.3s, 2.6s, 5.1s, 10.2s, 20.4s, 41s, 82s
 	maxRetries     = 15
 	controllerName = "awselb"
+
+	// elbResyncDuration represents period of time after which we will re-add machines to the ELBs.
+	// Once successfully added we add a timestamp to status, each time the machine is synced we check if
+	// this duration is exceeded and if so, re-add to the ELB to ensure we repair any machines
+	// accidentally taken out of rotation.
+	elbResyncDuration = 2 * time.Hour
```
staebler: Although not necessary now, eventually we should add command-line arguments to allow this to be configurable.

pkg/controller/awselb/aws_elb_controller.go
```diff
@@ -285,6 +325,36 @@ func (c *Controller) syncMachine(key string) error {
...
+// updateStatus sets the LastELBSync time in status to now, and updates the machine.
+func (c *Controller) updateStatus(machine *capiv1.Machine, mLog log.FieldLogger) error {
+
+	mLog.Debug("updating machine status")
+
+	awsStatus, err := controller.AWSMachineProviderStatusFromClusterAPIMachine(machine)
+	if err != nil {
+		return err
+	}
+	awsStatus.LastELBSync = &metav1.Time{Time: time.Now()}
```
staebler: A nit, but I would rather see this written like this.

```
now := metav1.Now()
awsStatus.LastELBSync = &now
```

Then, you can also reuse `now` later when setting `LastUpdated`.

pkg/controller/awselb/aws_elb_controller_test.go
```diff
@@ -0,0 +1,298 @@
...
+	cases := []struct {
+		name                         string
+		generation                   int64
+		lastSuccessfulSyncGeneration int64
+		lastSuccessfulSync           time.Time
+		resyncExpected               bool
+	}{
+		{
+			name:                         "no last sync",
+			generation:                   int64(1),
```
staebler: Remove the `int64` type conversions throughout these test cases. They are not necessary.

pkg/controller/awselb/aws_elb_controller_test.go
```diff
@@ -0,0 +1,298 @@
...
+				updateAction, ok := action.(clientgotesting.UpdateAction)
+				assert.True(t, ok)
+
+				updatedObject := updateAction.GetObject()
+				machine, ok := updatedObject.(*capiv1.Machine)
+				clustopStatus, err := controller.AWSMachineProviderStatusFromClusterAPIMachine(machine)
+				assert.NoError(t, err)
+				assert.NotNil(t, clustopStatus.LastELBSync)
+				assert.True(t, clustopStatus.LastELBSync.Time.After(tc.lastSuccessfulSync))
+				assert.NotEqual(t, int64(0), clustopStatus.LastELBSyncGeneration)
```
staebler: Test that this equals the generation of the Machine instead of just that it is not zero.

pkg/controller/controller_utils.go
```diff
@@ -683,3 +683,13 @@ func BuildClusterAPIMachineSet(ms *clusteroperator.ClusterMachineSet, clusterDep
 
 	return &capiMachineSet, nil
 }
+
+// StringPtrsEqual safely returns true if the value for each string pointer is equal, or both are nil.
+func StringPtrsEqual(s1, s2 *string) bool {
+	if s1 == nil && s2 == nil {
```
staebler: Getting *really* nit-picky here, but this alternative will save a string comparison if the pointers are equal. Subjectively, the indentation of the flow is better, too, since the expected behavior of comparing the pointed-to values is not indented.

```
if s1 == s2 {
  return true
}
if s1 == nil || s2 == nil {
  return false
}
return *s1 == *s2
```

dgoodwin: Sounds good, but I think I need to add a check if both are nil, because I need that to return true, before checking if just one is nil and returning false.

pkg/clusterapi/aws/actuator.go
dgoodwin: This actually exposed that the code here was unreachable, we'd error out if no instances are found and not update status. Now updating status and then returning the error, test to match.

pkg/controller/controller_utils.go
dgoodwin: Nope I don't, passes as is. Thanks!

PR #267, opened 2018-06-15
Refactor for cloud provider specific provider configs.

dgoodwin:
Stop using ClusterDeploymentSpec for both ClusterDeployment.Spec and Cluster.Spec.ProviderConfig. Similar for status.

ClusterDeployment had a full copy of the status that was never set or
used. Conditions remain for cluster version conditions.

Removed unused fields such as machine set count, and master/infra
machine set names.

Limited cluster version info, in the provider config, to just what is necessary for a cluster.

Added concept of distribution config, which we may push for upstream as a separate field.


pkg/apis/clusteroperator/types.go
```diff
@@ -402,6 +391,21 @@ type ClusterDeploymentStatus struct {
...
+	// DeprovisionedComputeMachinesets is true if the compute machinesets of this cluster
+	// have been deprovisioned.
+	DeprovisionedComputeMachinesets bool
+}
+
+// ClusterDeploymentStatus contains the status for a cluster
+type ClusterDeploymentStatus struct {
 	// Ready is true if the master of the cluster is ready to be used and can be accessed using
 	// the KubeconfigSecret
 	Ready bool
```
csrwng: I don't know that we will reflect the status from the cluster here, I would just leave it on the ClusterProviderStatus for now

pkg/apis/clusteroperator/types.go
```diff
@@ -412,10 +416,6 @@ type ClusterDeploymentStatus struct {
 	// ClusterVersionRef references the resolved clusterversion the cluster should be running.
 	// +optional
 	ClusterVersionRef *corev1.ObjectReference
-
```
csrwng: Should Conditions be of type []ClusterDeploymentCondition and only include ClusterDeployment-specific type of conditions there?

pkg/controller/clusterinstall/controller.go
```diff
@@ -405,16 +405,13 @@ func (s *jobSyncStrategy) DoesOwnerNeedProcessing(owner metav1.Object) bool {
...
-	if cluster.ClusterDeploymentStatus.ClusterVersionRef == nil {
-		// staebler: Temporary work-around until we have a controller that sets
-		// the ClusterVersionRef for cluster-api clusters.
-		_, err := s.controller.clustopClient.ClusteroperatorV1alpha1().
-			ClusterVersions(cluster.ClusterDeploymentSpec.ClusterVersionRef.Namespace).
-			Get(cluster.ClusterDeploymentSpec.ClusterVersionRef.Name, metav1.GetOptions{})
-		if err != nil {
-			return false
-		}
+	// TODO: staebler: Temporary work-around until we have a controller that sets
```
csrwng: I wonder if this is needed here given that we are referencing the ClusterVersion from the ClusterDeployment and we check for its existence in the ClusterDeployment controller before attempting to create the Cluster resource.

pkg/controller/infra/infra_controller.go
```diff
@@ -305,19 +305,16 @@ func (s *jobSyncStrategy) DoesOwnerNeedProcessing(owner metav1.Object) bool {
...
-	if cluster.ClusterDeploymentStatus.ClusterVersionRef == nil {
-		// staebler: Temporary work-around until we have a controller that sets
-		// the ClusterVersionRef for cluster-api clusters.
-		_, err := s.controller.coClient.ClusteroperatorV1alpha1().
-			ClusterVersions(cluster.ClusterDeploymentSpec.ClusterVersionRef.Namespace).
-			Get(cluster.ClusterDeploymentSpec.ClusterVersionRef.Name, metav1.GetOptions{})
-		if err != nil {
-			return false
-		}
+	// TODO: staebler: Temporary work-around until we have a controller that sets
```
csrwng: same here

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -321,12 +321,12 @@ func (c *Controller) syncClusterDeployment(key string) error {
 		return fmt.Errorf("error retrieving cluster object: %v", err)
 	}
 
-	clusterDeploymentStatus, err := controller.ClusterStatusFromClusterAPI(cluster)
+	clusterProviderStatus, err := controller.ClusterProviderStatusFromClusterAPI(cluster)
```
csrwng: I would rename the function `ClusterProviderStatusFromCluster`... there's no reason to distinguish cluster-api vs non-cluster-api clusters anymore

pkg/apis/clusteroperator/types.go
```diff
@@ -406,16 +406,8 @@ type ClusterProviderStatus struct {
 
 // ClusterDeploymentStatus contains the status for a cluster
 type ClusterDeploymentStatus struct {
-	// Ready is true if the master of the cluster is ready to be used and can be accessed using
```
csrwng: Ignore the previous comment to remove this :) ... I'm looking at this commit by commit

pkg/apis/clusteroperator/types.go
dgoodwin: In addition to duplicating the type structs, I think we'd need to duplicate a few functions like https://github.com/openshift/cluster-operator/blob/master/pkg/controller/controller_utils.go#L155. What do you think?

pkg/controller/clusterinstall/controller.go
dgoodwin: Agreed, I think the copy to the cluster provider spec is the new "resolved".

pkg/apis/clusteroperator/types.go
```diff
@@ -178,15 +185,51 @@ type AWSRegionAMIs struct {
...
+	// created. It should not be set by the user.
+	ClusterID string
+}
+
+// OpenShiftDistributionConfig contains configuration specific to the OpenShift distribution.
+type OpenShiftDistributionConfig struct {
+	ClusterConfigSpec
+
+	// Version contains information on the OpenShift version that should be installed on this cluster.
+	Version OpenShiftDistributionConfigVersion
```
csrwng: Should this also be a member of ClusterVersionSpec as well ? (instead of duplicating the fields there)

pkg/apis/clusteroperator/types.go
```diff
@@ -178,15 +185,51 @@ type AWSRegionAMIs struct {
...
+}
+
+type OpenShiftDistributionConfigVersion struct {
```
csrwng: I would leave 'Distribution' out of the name

pkg/apis/clusteroperator/types.go
csrwng: agree

pkg/apis/clusteroperator/types.go
```diff
@@ -178,15 +185,51 @@ type AWSRegionAMIs struct {
...
+	// OpenShiftConfig contains configuration specific to the OpenShift distribution.
+	OpenShiftConfig OpenShiftDistributionConfig
+}
+
+// ClusterProviderConfig contains cluster specification relevant to all cloud providers.
+type ClusterProviderConfig struct {
+	// ClusterID is the ID to assign to the cluster in the cloud provider.
+	// This will be generated by the API Server when the ClusterDeployment is
+	// created. It should not be set by the user.
+	ClusterID string
```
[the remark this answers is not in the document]
dgoodwin: Good point, will remove.

pkg/apis/clusteroperator/types.go
dgoodwin: Agree as in lets fork the code or should we leave it as is? Assuming it was a +1 for forking.

pkg/apis/clusteroperator/types.go
dgoodwin: The objects are starting to look weird no matter how I slice this.

We could do:

```
ClusterVersion
- VMImages
- OpenShiftConfigVersion
  - DeploymentType
  - Version
  - Images
OpenShiftDistributionConfig
- OpenShiftConfigVersion
  - DeploymentType
  - Version
  - Images
- other things
```

Or I could move Images up one level on both.

csrwng: true... thinking about it some more, I'd be ok leaving it as is.

pkg/controller/controller_utils.go
```diff
@@ -344,28 +407,28 @@ func AddLabels(obj metav1.Object, additionalLabels map[string]string) {
 	}
 }
 
-// ClusterDeploymentSpecFromCluster gets the cluster-operator ClusterDeploymentSpec from the
+// AWSClusterDeploymentSpecFromCluster gets the cluster-operator ClusterDeploymentSpec from the
```
csrwng: wrong function name

pkg/controller/controller_utils.go
```diff
@@ -344,28 +407,28 @@ func AddLabels(obj metav1.Object, additionalLabels map[string]string) {
...
-	spec, ok := obj.(*clusteroperator.ClusterProviderConfigSpec)
+	spec, ok := obj.(*clusteroperator.AWSClusterProviderConfig)
 	if !ok {
 		return nil, fmt.Errorf("Unexpected object: %#v", gvk)
 	}
-	return &spec.ClusterDeploymentSpec, nil
+	return spec, nil
 }
 
 // ClusterStatusFromClusterAPI gets the cluster-operator ClusterStatus from the
```
csrwng: Fix comment

pkg/controller/controller_utils.go
```diff
@@ -375,17 +438,33 @@ func ClusterStatusFromClusterAPI(cluster *clusterapi.Cluster) (*clusteroperator.
...
-	return &status.ClusterDeploymentStatus, nil
+	return status, nil
 }
 
-// ClusterProviderConfigSpecFromClusterDeploymentSpec returns an encoded RawExtension with a ClusterProviderConfigSpec from a cluster deployment spec
-func ClusterProviderConfigSpecFromClusterDeploymentSpec(clusterDeploymentSpec *clusteroperator.ClusterDeploymentSpec) (*runtime.RawExtension, error) {
-	clusterProviderConfigSpec := &clusteroperator.ClusterProviderConfigSpec{
+// BuildClusterProviderConfig returns an encoded cloud provider specific RawExtension
+// for use in the cluster spec's ProviderConfig.
+func BuildClusterProviderConfig(clusterDeploymentSpec *clusteroperator.ClusterDeploymentSpec, cv clusteroperator.ClusterVersionSpec) (*runtime.RawExtension, error) {
```
csrwng: should the name be BuildAWSClusterProviderConfig ?

dgoodwin:
Ok all feedback addressed. I was able to get "mostly" through an e2e test, my remote machines did not seem to get actuated even though their machinesets/machines in etcd were created in remote cluster. Did not have time to dig into this yet. Deprovision worked smoothly.

csrwng:
Please remove wip from the title so we can merge it.

csrwng:
Also please squash the commits

dgoodwin:
Done.

PR #269, opened 2018-06-18
Skip role creation on infra

csrwng:
For now, skips role creation on infra until we get back to per-cluster roles.


csrwng:
@abutcher ptal
@twiest fyi

twiest:
@csrwng great, thanks!

PR #271, opened 2018-06-18
Update wait-for-cluster-ready command

csrwng:
Modifies wait-for-cluster-ready to only wait until the ClusterAPI is installed on the target cluster. 
It also introduces a timeout flag to quit when a timeout expires.

### DIFF
--- a/contrib/cmd/wait-for-cluster-ready/waitforcluster.go
+++ b/contrib/cmd/wait-for-cluster-ready/waitforcluster.go
@@ -21,6 +21,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


csrwng:
@dgoodwin ptal
This is for the CI e2e test to wait for when the cluster deployment is ready

contrib/cmd/wait-for-cluster-ready/waitforcluster.go
```diff
@@ -89,29 +105,33 @@ func waitForClusterReady(capiClient *capiclient.Clientset, cluster *capiv1.Clust
 				if err != nil {
 					continue
 				}
-				if status.Ready {
-					fmt.Printf("Cluster %s/%s is ready.\n", cluster.Namespace, cluster.Name)
+				if status.ClusterAPIInstalled {
+					fmt.Printf("Cluster %s/%s has cluster API installed.\n", cluster.Namespace, cluster.Name)
```
dgoodwin: What is the thought behind this switch? It seems like we might still be installing components for example, which might leave the cluster in a state that wouldn't pass some e2e tests. (no router / registry for example)

csrwng: There's an issue with the service catalog image, latest image only crash-loops. The reason is that the deployment for it only points to the :latest image instead of a version-specific one. My thinking is that in the latest code some flags have changed. Bottom line is that the components playbook never finishes an install. It should not stop us from claiming success. We could add a verification step to the e2e test to make sure that the registry/router is up and running.

PR #272, opened 2018-06-18
Change node instances to 'terminate' by default.

twiest:
- [x] Change the AWS shutdown behavior for node instances to "terminate"
- [x] Explicitly set the AWS shutdown behavior for master instances to "stop"
- [x] Add unit test verification for both master and node shutdown behaviors
- [x] Manually test / verify that master instances are created with the shutdown behavior set to stop
- [x] Manually test / verify that node instances are created with the shutdown behavior set to terminate

### DIFF
--- a/pkg/clusterapi/aws/actuator_test.go
+++ b/pkg/clusterapi/aws/actuator_test.go
@@ -48,19 +48,21 @@ func init() {
 }
 
 const (
-	testNamespace             = "test-namespace"
-	testClusterDeploymentName = "test-cluster-deployment"
-	testClusterDeploymentUUID = types.UID("test-cluster-deployment-uuid")
-	testClusterID             = "test-cluster-id"
-	testClusterVerName        = "v3-10"
-	testClusterVerNS          = "cluster-operator"
-	testClusterVerUID         = types.UID("test-cluster-version")
-	testRegion                = "us-east-1"
-	testImage                 = "testAMI"
-	testVPCID                 = "testVPCID"
-	testSubnetID              = "testSubnetID"
-	testAZ                    = "us-east-1c"
-	testMachineName           = "testmachine"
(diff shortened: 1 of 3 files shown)


pkg/clusterapi/aws/actuator_test.go
```diff
@@ -48,19 +48,21 @@ func init() {
...
+	testClusterVerNS           = "cluster-operator"
+	testClusterVerUID          = types.UID("test-cluster-version")
+	testRegion                 = "us-east-1"
+	testImage                  = "testAMI"
+	testVPCID                  = "testVPCID"
+	testSubnetID               = "testSubnetID"
+	testAZ                     = "us-east-1c"
+	testMachineName            = "testmachine"
+	testShutdownBehaviorNodes  = "terminate"
+	testShutdownBehaviorMaster = "stop"
```
dgoodwin: Because you're in the same go package you can just reference the constants defined in actuator.go with the same values.

twiest: @dgoodwin Yeah, I thought about that. I didn't do that for 2 reasons:

1) I wanted to ensure that if those constants change to invalid values in the future (e.g. "stopped" instead of "stop"), that the tests fail.

2) I wanted to follow the established patterns already in that test file (e.g. see how "master" is statically defined as the expected string instead of referencing the hostTypeMaster constant [here](https://github.com/twiest/cluster-operator/blob/f7de5f05d383932e980b8664667ce395c267054a/pkg/clusterapi/aws/actuator_test.go#L272)).


That being said, I will absolutely change it to use the other constants if you feel strongly about it.

dgoodwin: Nah that sounds fine, thanks.

PR #273, opened 2018-06-19
Leverage shorter ELB suffixes in Openshift Ansible

csrwng:


mazzystr:
What does cp pertain to?

staebler:
The elb names need to be changed in this test, too. https://github.com/openshift/cluster-operator/blob/eed4455eab3791da3bc596d0c42af6d9180c9941/pkg/ansible/generate_test.go#L74

staebler:
> What does cp pertain to?

@mazzystr "cp" refers to control plane.

csrwng:
thx @staebler fixed now

PR #275, opened 2018-06-20
Wait for cluster to be created when waiting for it to be ready

csrwng:
Add a polling wait for the cluster to be created in the wait-for-cluster-ready utility. This is needed when invoking the utility right after creating the clusterdeployment. It needs to allow the clusterdeployment controller to create the cluster first.


csrwng:
@dgoodwin ptal

PR #276, opened 2018-06-21
 Detect and terminate stopped nodes

twiest:
- [x] Revert commit of initial attempt at removing stopped instances (aws shutdownBehavior doesn't work in all cases).
- [x] Remove stopped nodes when new nodes are created for the same machine.
- [x] Explicitly do NOT remove stopped master instances.
- [x] Fix missing clusterid in MachineSetSpec in aws-actuator-test/main.go
- [x] Fix go linting issues in aws-actuator-test/main.go
- [x] Fix cli help messages in aws-actuator-test/main.go

### DIFF
--- a/pkg/clusterapi/aws/actuator.go
+++ b/pkg/clusterapi/aws/actuator.go
@@ -113,9 +113,27 @@ func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 	return a.updateStatus(machine, instance, mLog)
 }
 
+// removeStoppedMachine removes all instances of a specific machine that are in a stopped state.
+func (a *Actuator) removeStoppedMachine(machine *clusterv1.Machine, client Client, mLog log.FieldLogger) error {
+	instances, err := GetStoppedInstances(machine, client)
+	if err != nil {
+		return fmt.Errorf("Error getting stopped instances: %v", err)
+	}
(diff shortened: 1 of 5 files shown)


twiest:
@csrwng Would you mind taking a look at this please?

pkg/clusterapi/aws/actuator.go
```diff
@@ -115,9 +113,30 @@ func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+
+	return TerminateInstances(client, instances, mLog)
+}
+
 // CreateMachine starts a new AWS instance as described by the cluster and machine resources
 func (a *Actuator) CreateMachine(cluster *clusterv1.Cluster, machine *clusterv1.Machine) (*ec2.Instance, error) {
 	mLog := clustoplog.WithMachine(a.logger, machine)
+
+	isMaster := false
+	if controller.MachineHasRole(machine, capicommon.MasterRole) {
```
csrwng: Just a nit, you can collapse a few lines by just doing
```go
isMaster := controller.MachineHasRole(machine, capicommon.MasterRole)
```

twiest:
@csrwng thanks for the review and for pointing out that optimization. I believe that I've addressed your comment. PTAL.

PR #277, opened 2018-06-22
Set openshift_node_config_name in userdata.

abutcher:
This will need to be updated on our side once https://github.com/openshift/openshift-ansible/pull/8923 is backported to 3.10 and we've created a new AMI using the updated release-3.10 branch.


csrwng:
@abutcher you'll need to fix the actuator test 
```
s/openshift_group_type/openshift_node_config_name/g
```
did the trick for me

PR #278, opened 2018-06-29
Introduce separate flag to build images on deployment

csrwng:
We likely don't need to rebuild images every time we deploy. This introduces a separate flag so we can control whether images should be built but still allow pushing them to the internal registry.


csrwng:
@joelddiaz @twiest ptal

PR #280, opened 2018-06-29
Use latest ansible 3.10 branch

csrwng:
Uses the latest 3.10 latest commit from openshift-ansible. Until we get a tag in that branch, we will just pull the latest.

Also modifies the actuator to add a `openshift-node-group-config` tag to the instance


csrwng:
@abutcher ptal

PR #287, opened 2018-07-05
Drop unnecessary build_images param to devel deploy playbook.

dgoodwin:
No need for two variables here, we either want to push images, in which
case we should build them, or do neither.


csrwng:
Actually the reason I introduced another variable was because I often find myself starting a new cluster with `oc cluster up` and I want to deploy cluster operator. I've already built the images so don't want to build them again. I just want them to be re-pushed. Building new images takes a while and clutters my docker images.

dgoodwin:
My mistake I thought I did this and for no reason. Will close.

PR #288, opened 2018-07-05
Improve ready check

csrwng:
This is basically a rewrite of the wait-for-cluster-ready utility. It adds timeouts for the different phases of cluster provisioning/install and ensures that the resulting cluster ends up having the nodes it's supposed to have and at least has a valid router and registry deployment.


csrwng:
@dgoodwin ptal

contrib/cmd/wait-for-cluster-ready/waitforcluster.go
```diff
@@ -23,130 +23,500 @@ import (
...
+	startedInstall := false
+	return o.waitForClusterStatus(client, name, o.ClusterAPIInstallTimeout, func(status *coapiv1.ClusterProviderStatus) bool {
+		if status.ClusterAPIInstalled {
+			log.Infof("cluster API installed on cluster %s/%s", o.Namespace, name)
+			return true
+		}
+		installing := controller.FindClusterCondition(status.Conditions, coapiv1.ClusterAPIInstalling)
+		if installing != nil && !startedInstall && installing.Status == corev1.ConditionTrue {
+			startedInstall = true
+			log.Infof("cluster API started installing on cluster %s/%s at %v", o.Namespace, name, installing.LastTransitionTime)
```
dgoodwin: Would suggest just a quick scan of the output for consistency, specifically there's a mix of starting case for each line and trailing punctuation. No pref for either just would look a little cleaner if they were all the same.

csrwng:
Comments addressed

csrwng:
Realized I still had period on some sentences. *Now* it's ready :)

PR #289, opened 2018-07-05
Skip remote machineset cleanup when sync never happend

csrwng:
Fixes issue when clusterdeployment never gets to sync remote machinesets and gets deleted, the finalizer for the remote machinesets remains and doesn't allow the clusterdeployment to be deleted.

### DIFF
--- a/pkg/controller/remotemachineset/remotemachineset_controller.go
+++ b/pkg/controller/remotemachineset/remotemachineset_controller.go
@@ -338,10 +338,6 @@ func (c *Controller) syncClusterDeployment(key string) error {
 		return c.syncDeletedClusterDeployment(clusterDeployment)
 	}
 
-	if !hasFinalizer(clusterDeployment) {
-		return c.addFinalizer(clusterDeployment)
-	}
-
 	cluster, err := c.clusterInformer.Clusters(clusterDeployment.Namespace).Get(clusterDeployment.Spec.ClusterID)
 	if err != nil {
 		return fmt.Errorf("error retrieving cluster object: %v", err)
@@ -544,7 +540,7 @@ func (c *Controller) syncMachineSets(clusterDeployment *cov1.ClusterDeployment, 
 		}
 	}
 
-	return nil
+	return c.setMachineSetsSyncedStatus(clusterDeployment)
 }
 
 // return list of non-master machinesets from a given ClusterDeployment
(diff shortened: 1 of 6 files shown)


csrwng:
@dgoodwin ptal

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -335,6 +335,9 @@ func (c *Controller) syncClusterDeployment(key string) error {
 		if !hasFinalizer(clusterDeployment) {
 			return nil
 		}
+		if !hasSynced(clusterDeployment) {
+			return c.deleteFinalizer(clusterDeployment)
+		}
```
dgoodwin: Would it be better if we only added the finalizer after successfully syncing machine sets?

dgoodwin:
This is great, would be great if we could immediately delete a cluster and see it kill off all jobs and clean up everything running, this gets us one step closer.

pkg/controller/remotemachineset/remotemachineset_controller.go
csrwng: I like that better... updated

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -575,13 +571,19 @@ func (c *Controller) deleteFinalizer(clusterDeployment *cov1.ClusterDeployment)
 	return err
 }
 
-func (c *Controller) addFinalizer(clusterDeployment *cov1.ClusterDeployment) error {
+func (c *Controller) setMachineSetsSyncedStatus(clusterDeployment *cov1.ClusterDeployment) error {
 	clusterDeployment = clusterDeployment.DeepCopy()
 	controller.AddFinalizer(clusterDeployment, cov1.FinalizerRemoteMachineSets)
+	clusterDeployment.Status.RemoteMachineSetsSynced = true
+	clusterDeployment.Status.RemoteMachineSetsSyncedGeneration = clusterDeployment.Generation
```
dgoodwin: I think that may leave these new fields as unused, do you think they still provide some value in having set? Otherwise we can just assume this is the state if the finalizer has been added.

pkg/controller/remotemachineset/remotemachineset_controller.go
```diff
@@ -575,13 +571,19 @@ func (c *Controller) deleteFinalizer(clusterDeployment *cov1.ClusterDeployment)
...
 	return err
 }
 
 func hasFinalizer(clusterDeployment *cov1.ClusterDeployment) bool {
 	return controller.HasFinalizer(clusterDeployment, cov1.FinalizerRemoteMachineSets)
 }
+
+func hasSynced(clusterDeployment *cov1.ClusterDeployment) bool {
+	return clusterDeployment.Status.RemoteMachineSetsSynced
+}
```
dgoodwin: Function now unused as well.

csrwng:
Updated

dgoodwin:
/lgtm

Sounds good to me

PR #290, opened 2018-07-06
Fix check for clusterversion existence

csrwng:
The reason the existence check is failing is because the ServiceUnavailable error is not the only error returned when the apiserver is not ready yet. So now just testing for either of the possible valid outcomes ... either there is no error or the error contains "NotFound". Otherwise, keep retrying.


csrwng:
@dgoodwin ptal

PR #292, opened 2018-07-10
Hookup pod and service network CIDRs in the API.

dgoodwin:
Re-use the cluster-api NetworkConfig struct.

Supports multiple ranges but currently validates that only one is
provided, as this appears to be all openshift-ansible supports.

Will default if none is provided.

Once set these values are immutable.

### DIFF
--- a/pkg/apis/clusteroperator/validation/clusterdeployment.go
+++ b/pkg/apis/clusteroperator/validation/clusterdeployment.go
@@ -86,6 +86,33 @@ func validateClusterDeploymentSpec(spec *clusteroperator.ClusterDeploymentSpec, 
 		allErrs = append(allErrs, field.Required(versionPath.Child("name"), "must specify a cluster version to install"))
 	}
 
+	if spec.NetworkConfig.ServiceDomain == "" {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec", "networkConfig", "serviceDomain"),
+			spec.NetworkConfig.ServiceDomain,
+			"invalid cluster configuration: missing service domain"))
+	}
+
+	if len(spec.NetworkConfig.Pods.CIDRBlocks) == 0 {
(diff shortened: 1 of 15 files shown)


pkg/apis/clusteroperator/validation/clusterdeployment.go
```diff
@@ -86,6 +86,13 @@ func validateClusterDeploymentSpec(spec *clusteroperator.ClusterDeploymentSpec,
 		allErrs = append(allErrs, field.Required(versionPath.Child("name"), "must specify a cluster version to install"))
 	}
 
+	if len(spec.NetworkConfig.Services.CIDRBlocks) > 1 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("networkConfig").Child("services"), &spec.NetworkConfig.Services, "only one service network subnet is currently supported"))
+	}
+	if len(spec.NetworkConfig.Pods.CIDRBlocks) > 1 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("networkConfig").Child("pods"), &spec.NetworkConfig.Pods, "only one pod network subnet is currently supported"))
+	}
+
```
csrwng: We should also do the same validation that's done upstream:
https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha1/cluster_types.go#L118-L136

dgoodwin: I've got defaults being added for all three of those fields, do we need validation in that case or can I assume those defaults will *always* be applied?

dgoodwin:
Updated.

csrwng:
@dgoodwin the integration test keeps failing, you may want to look into it

dgoodwin:
My mistake, forgot to run tests after I added validation to require some of those fields.

PR #295, opened 2018-07-11
Hook up SDN plugin name.

dgoodwin:
Required, immutable, and defaulted to match openshift-ansible.


dgoodwin:
@csrwng could you take a look at this one

PR #298, opened 2018-07-12
Remove ClusterID field from MachineSetSpec

twiest:
- [x] Remove ClusterID field from MachineSetSpec
- [x] Fix bug where remote machine set is created with a label of "cluster:" instead of "clusteroperator.openshift.io/cluster"
- [x] Test cluster creation works as expected

### DIFF
--- a/contrib/cmd/aws-actuator-test/main.go
+++ b/contrib/cmd/aws-actuator-test/main.go
@@ -172,7 +172,6 @@ func testClusterAPIResources(name string) (*clusterv1.Cluster, *clusterv1.Machin
 			Kind:       "MachineSetProviderConfigSpec",
 		},
 		MachineSetSpec: cov1.MachineSetSpec{
-			ClusterID: name,
 			VMImage: cov1.VMImage{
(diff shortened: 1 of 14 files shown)


twiest:
@staebler Hey, this is my first shot at removing ClusterID from MachineSetSpec. It is fully functional, but before I ask for a review, I want to make sure this is what you were thinking.

Is this what you were looking for?

staebler:
@twiest It looks good to me. The only question that I would raise is whether `ClusterID` in `ClusterDeploymentSpec` should be renamed to `ClusterName`. However, even if that is something worthwhile to do, it can be done as a separate PR.

twiest:
@staebler I'm fine making that change (either as part of this PR or a separate PR).

@dgoodwin What are your thoughts on that change and whether it should be part of this PR or another PR?

dgoodwin:
Totally up to you

twiest:
@staebler @dgoodwin Ok, I think I'd prefer to merge this PR and I'll start a new PR with the new changes.

Can I get a lgtm please?

PR #300, opened 2018-07-17
Use aws-machine-controller binary on target cluster

csrwng:
When testing/developing on AWS actuator, we have to go through the trouble of pushing an image to an external registry and manually changing the deployment on the target cluster. Also when CI tests it, we're not really testing the built AWS actuator on the target cluster. This pull does the following:
- Add the aws-machine-controller binary to the cluster-operator-ansible image
- Modifies the playbook that deploys the cluster api to copy the aws-machine-controller binary to the target master
- Modifies the template for the cluster-api to do a host mount for the aws-machine-controller binary so that it uses that instead of whatever was on the image.

This results in the following limitations:
- Only 1 master is supported
- Only the aws-machine-controller binary is copied over. Other changes to the image are not reflected on the target system.

### DIFF
--- a/build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
+++ b/build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
@@ -29,6 +29,13 @@
     import_role:
       name: lib_openshift
 
+  - name: copy cluster-operator binary to master
+    copy:
+      src: files/aws-machine-controller
+      dest: /usr/bin/aws-machine-controller
+      setype: container_file_t
+      mode: 0755
+
   - name: slurp local template content
     slurp:
       src: files/cluster-api-template.yaml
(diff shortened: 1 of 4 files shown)


csrwng:
@dgoodwin @abutcher ptal, this is an initial stab at solving the issue of running the local actuator on the target cluster.

build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
```diff
@@ -29,6 +29,13 @@
     import_role:
       name: lib_openshift
 
+  - name: copy cluster-operator binary to master
+    copy:
+      src: files/aws-machine-controller
+      dest: /usr/bin/aws-machine-controller
+      setype: container_file_t
+      mode: 0755
```
dgoodwin: If you broke this out into a separate play run on hosts: masters, would we have multi-master support?

PR #301, opened 2018-07-17
Change CD.Spec.ClusterID to CD.Spec.ClusterName

twiest:
- [x] Change ClusterDeployment.Spec.ClusterID to ClusterDeployment.Spec.ClusterName
- [x] Test cluster creation works as expected


twiest:
@staebler Hey, this is my first attempt at the ClusterID -> ClusterName change.

Is this what you were looking for?

pkg/apis/clusteroperator/types.go
```diff
@@ -227,10 +227,10 @@ type OpenShiftConfigVersion struct {
 
 // ClusterDeploymentSpec is the specification of a cluster's hardware and configuration
 type ClusterDeploymentSpec struct {
-	// ClusterID is the ID to assign to the cluster in the cloud provider.
+	// ClusterName is the name to assign to the cluster in the cloud provider.
 	// This will be generated by the API Server when the ClusterDeployment is
 	// created. It should not be set by the user.
-	ClusterID string
+	ClusterName string `json:"clusterName"`
```
staebler: In the unversioned API, do not add the json tag to field. In other words, remove the `` `json:"clusterName"`` `

twiest: @staebler ah, ok, thanks for clarifying that. I fixed it for NetworkingConfig as well in the same file.

PR #303, opened 2018-07-18
Remove unused label consts.

staebler:
While reviewing another PR, I noticed that there were a handful of unused labels consts in controller_utils.go. This PR removes those. It also removes the unneeded "cluster-uid" label and replaces the "cluster" label placed on jobs with the more specific and ubiquitous "clusteroperator.openshift.io/cluster" label.


dgoodwin:
/lgtm

thanks @staebler

PR #305, opened 2018-07-20
Honor VPC name and Subnet in AWS hardware spec

csrwng:
Currently we ignore VPCName from the hardware spec. This PR makes it so we honor it and assume it exists when set. We also need a fix in the ansible openshift_aws role to use the vpc name instead of assuming it's the cluster name. That is getting fixed here: https://github.com/openshift/openshift-ansible/pull/9249
When that fix goes through and it gets backported to 3.10, we can remove the override in this repo.


csrwng:
@dgoodwin @abutcher ptal

PR #307, opened 2018-07-23
Consolidate separate utilities under single command

csrwng:
- Creates a new top-level command to hold utilities under contrib/bin -> `coutil`
- Removes obsolete export-cluster-api utility
- Temporarily leaves a separate wait-for-cluster-ready utility while CI is transitioned to new command.


dgoodwin:
Looks good, could you just add a quick note to the README on how to build and roughly what types of things will be in there.

csrwng:
@dgoodwin README updated.

PR #309, opened 2018-07-26
Add controller to link machines to nodes and apply labels + taints.

dgoodwin:
- applies the machine annotation to the node
- applies the desired labels/taints to the node
- now runs the new controller in a "cluster-operator-remote-controllers" deployment in the target cluster.
- NOTE: changing labels on a machineset is not reflected on pre-existing machines, let alone their nodes. Issue filed upstream to see if this is desired: https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha1/machine_types.go#L52
- NOTE: labels will not be applied to masters as they are not created in-cluster, where the controller is running today.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -369,7 +369,7 @@ OA_ANSIBLE_BRANCH ?= release-3.10
 define build-cluster-operator-ansible-image #(repo, branch, imagename, tag, clusterapi_playbook)
         cp build/cluster-operator-ansible/playbooks/cluster-api-prep/$5 build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
         cp bin/aws-machine-controller build/cluster-operator-ansible/playbooks/cluster-api-prep/files
-	cp bin/cluster-operator build/cluster-operator-ansible/playbooks/cluster-api-prep/files
+        cp bin/cluster-operator build/cluster-operator-ansible/playbooks/cluster-api-prep/files
 	docker build -t "$3:$4" --build-arg=CO_ANSIBLE_URL=$1 --build-arg=CO_ANSIBLE_BRANCH=$2 build/cluster-operator-ansible
 endef
 
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -668,6 +668,19 @@ func (c *Controller) syncControlPlane(clusterDeployment *clustop.ClusterDeployme
 		updatedMachineSet.Spec.Replicas = &machineSetSize
 	}
 
+	// Build a new set of expected labels so we clear out any previous:
+	newLabels := map[string]string{
+		clustop.ClusterDeploymentLabel: clusterDeployment.Name,
+		clustop.ClusterNameLabel:       cluster.Name,
+		clustop.MachineSetNameLabel:    machineSet.Name,
+	}
+	for k, v := range machineSetConfig.NodeLabels {
+		newLabels[k] = v
+	}
+	updatedMachineSet.Spec.Template.Spec.Labels = newLabels
+
+	updatedMachineSet.Spec.Template.Spec.Taints = machineSetConfig.NodeTaints
+
 	specProviderConfig, err := controller.MachineProviderConfigFromMachineSetConfig(machineSetConfig, &clusterDeployment.Spec, clusterVersion)
 	if err != nil {
 		return fmt.Errorf("cannot create a machine providerconfig from machineset config for cluster deployment %s/%s: %v", clusterDeployment.Namespace, clusterDeployment.Name, err)
@@ -701,6 +714,9 @@ func buildMasterMachineSet(clusterDeployment *clustop.ClusterDeployment, cluster
 	}
 	machineSet.Labels[clustop.ClusterDeploymentLabel] = clusterDeployment.Name
 	machineSet.Labels[clustop.ClusterNameLabel] = cluster.Name
+	for k, v := range machineSetConfig.NodeLabels {
+		machineSet.Labels[k] = v
+	}
 	blockOwnerDeletion := false
 	ownerRef := metav1.NewControllerRef(clusterDeployment, controller.ClusterDeploymentKind)
 	ownerRef.BlockOwnerDeletion = &blockOwnerDeletion
@@ -714,7 +730,16 @@ func buildMasterMachineSet(clusterDeployment *clustop.ClusterDeployment, cluster
 	replicas := int32(machineSetConfig.Size)
 	machineSet.Spec.Replicas = &replicas
 	machineSet.Spec.Template.Labels = machineSetLabels
-	machineSet.Spec.Template.Spec.Labels = machineSetLabels
+
+	// Transfer node labels onto the machine template:
+	machineSet.Spec.Template.Spec.Labels = map[string]string{}
+	for k, v := range machineSetLabels {
+		machineSet.Spec.Template.Spec.Labels[k] = v
+	}
+	for k, v := range machineSetConfig.NodeLabels {
+		machineSet.Spec.Template.Spec.Labels[k] = v
+	}
+	machineSet.Spec.Template.Spec.Taints = machineSetConfig.NodeTaints
 
 	providerConfig, err := controller.MachineProviderConfigFromMachineSetConfig(machineSetConfig, &clusterDeployment.Spec, clusterVersion)
 	if err != nil {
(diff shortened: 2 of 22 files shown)


Makefile
```diff
@@ -385,6 +385,7 @@ OA_ANSIBLE_BRANCH ?= release-3.10
 define build-cluster-operator-ansible-image #(repo, branch, imagename, tag, clusterapi_playbook)
         cp build/cluster-operator-ansible/playbooks/cluster-api-prep/$5 build/cluster-operator-ansible/playbooks/cluster-api-prep/deploy-cluster-api.yaml
         cp bin/aws-machine-controller build/cluster-operator-ansible/playbooks/cluster-api-prep/files
+        cp bin/cluster-operator build/cluster-operator-ansible/playbooks/cluster-api-prep/files
```
dgoodwin: This moves the cluster-operator binary into the image as well as the aws-machine-controller binary, so we can transfer it to the masters in devel scenarios. Ultimately the aws machine controller should move into this binary.

dgoodwin:
ready for review.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -669,6 +669,12 @@ func (c *Controller) syncControlPlane(clusterDeployment *clustop.ClusterDeployme
 		updatedMachineSet.Spec.Replicas = &machineSetSize
 	}
 
+	// Ensure labels and taints are up to date:
+	for k, v := range machineSetConfig.NodeLabels {
```
csrwng: Shouldn't we just set the Labels to whatever is in the machineSetConfig? I don't think we want to support additional labels in the derived resource.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -702,6 +708,9 @@ func buildMasterMachineSet(clusterDeployment *clustop.ClusterDeployment, cluster
 	}
 	machineSet.Labels[clustop.ClusterDeploymentLabel] = clusterDeployment.Name
 	machineSet.Labels[clustop.ClusterNameLabel] = cluster.Name
+	for k, v := range machineSetConfig.NodeLabels {
```
csrwng: Same as above

pkg/controller/nodelink/node_link_controller.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func (c *Controller) handleErr(err error, key interface{}) {
+	if err == nil {
+		c.queue.Forget(key)
+		return
+	}
+
+	if c.queue.NumRequeues(key) < maxRetries {
+		c.logger.Infof("Error syncing machine %v: %v", key, err)
```
csrwng: s/machine/node/

pkg/controller/nodelink/node_link_controller.go
```diff
@@ -0,0 +1,342 @@
...
+		c.queue.AddRateLimited(key)
+		return
+	}
+
+	utilruntime.HandleError(err)
+	c.logger.Infof("Dropping machine %q out of the queue: %v", key, err)
```
csrwng: s/machine/node/

pkg/controller/nodelink/node_link_controller.go
```diff
@@ -0,0 +1,342 @@
...
+		if err != nil {
+			nodeLog.Infof("machine annotation format is incorrect %v: %v\n", machineKey, err)
+			return err
+		}
+		matchingMachine, err = c.machinesLister.Machines(namespace).Get(machineName)
+		if err != nil {
+			return err
+		}
+	}
+	if matchingMachine == nil {
```
csrwng: Is it possible for the machine annotation to point to a machine that has been deleted and there is a new machine that should correspond to the node based on address? If so, I think it would make sense to always attempt to find the machine via annotation but if no annotation is set OR no machine was found, attempt an address match.

pkg/controller/nodelink/node_link_controller.go
```diff
@@ -0,0 +1,342 @@
...
+		nodeLog.Warnf("no matching machine found for node")
+		return fmt.Errorf("no matching machine found for node: %s", node.Name)
+	}
+
+	modNode := node.DeepCopy()
+
+	modNode.Annotations["machine"] = fmt.Sprintf("%s/%s", matchingMachine.Namespace, matchingMachine.Name)
+
+	if node.ObjectMeta.Labels == nil {
```
csrwng: really minor nit ... ObjectMeta is not needed because it's embedded anonymously... node.Labels works just as well.

pkg/controller/nodelink/node_link_controller.go
```diff
@@ -0,0 +1,342 @@
...
+		node.ObjectMeta.Labels = map[string]string{}
+	}
+	for k, v := range matchingMachine.Spec.ObjectMeta.Labels {
+		nodeLog.Debugf("copying %s = %s", k, v)
```
csrwng: s/copying/copying label/

pkg/controller/nodelink/node_link_controller.go
```diff
@@ -0,0 +1,342 @@
...
+		modNode.Labels[k] = v
+	}
+
+	// Taints are to be an authoritative list on the machine spec per cluster-api comments:
+	modNode.Spec.Taints = matchingMachine.Spec.Taints
+
+	if !reflect.DeepEqual(node, modNode) {
```
csrwng: This test may not be accurate. It will always evaluate to true if you've added your own labels to the node even though no update is needed. A more accurate test would keep track of any label modifications and a deepequal of Taints.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Can we make that assumption about a kube object? I expected they would be a mechanism for arbitrary apps to slap data into, and it wouldn't be safe for us to assume ownership.

csrwng: But this is not the kube object yet... this is just the machineset that was created from the clusterdeployment. In this case, it should be exactly the same.

dgoodwin: My mistake I follow you now. Thanks.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Am I correct that because this one is actually the labels on a master machine set being created, we would leave the door open for external actors setting labels, but a few lines down we would not. (comment incoming there)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -715,8 +724,17 @@ func buildMasterMachineSet(clusterDeployment *clustop.ClusterDeployment, cluster
...
-	machineSet.Spec.Template.Spec.Labels = machineSetLabels
+
+	// Transfer node labels onto the machine template:
+	machineSet.Spec.Template.Spec.Labels = map[string]string{}
+	for k, v := range machineSetLabels {
+		machineSet.Spec.Template.Spec.Labels[k] = v
+	}
+	for k, v := range machineSetConfig.NodeLabels {
+		machineSet.Spec.Template.Spec.Labels[k] = v
+	}
```
dgoodwin: This looks like the spot to just transfer the labels entirely.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Ignore previous comment, this one and below look correct, we need to add the assumed labels for cluster name and deployment, and everything the user requested in NodeLabels, sound right?

pkg/controller/clusterdeployment/clusterdeployment_controller.go
dgoodwin: Ok after staring at this for longer than I'd like to admit (there's an awful lot of labels flying around), I think this is out of sync with the code below, and if we change it to just take the NodeLabels wholesale, we introduce a bug. Below, during creation of the machine set, we transfer all NodeLabels, as well as add three for the cluster deployment, cluster, and machine set names. We should be doing the same here during updates to ensure these stay set correctly I'm thinking.

pkg/controller/nodelink/node_link_controller.go
dgoodwin: Slightly confused on why this would be, if I do a deep copy on the original node (including any labels set by a third party), then set some labels to (probably) what they were set to already, reflect.DeepEqual will consider it a change?

csrwng: Sorry you're right, please ignore. For whatever reason I was thinking we were starting from a new node object.

dgoodwin:
Updated per your feedback @csrwng

dgoodwin:
Updated again.

PR #310, opened 2018-07-30
aws client setup refactor

joelddiaz:
redo the parameters needed to set up an aws client (needed for other stuff i'm working on) so that it doesn't need to know about the internals of machinesetspec objects.

update controllers and tests using the aws client accordingly

### DIFF
--- a/pkg/controller/controller_utils.go
+++ b/pkg/controller/controller_utils.go
@@ -765,3 +765,18 @@ func DeleteFinalizer(object metav1.Object, finalizer string) {
 	finalizers.Delete(finalizer)
 	object.SetFinalizers(finalizers.List())
 }
+
+// GetSecretNameFromMachineSetSpec will retrieve the secret name holding AWS credentials
+// from a machcinesetspec if it exists. Otherwise it will return an empty string.
+func GetSecretNameFromMachineSetSpec(msSpec *clusteroperator.MachineSetSpec) (string, error) {
+	secretName := ""
+	if msSpec.ClusterHardware.AWS == nil {
+		return "", fmt.Errorf("no AWS cluster hardware set on machine spec")
+	}
+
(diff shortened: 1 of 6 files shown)


joelddiaz:
@csrwng i still need to test this once the machine object spec.roles[] changes are settled down, but could use a review on this at this point

pkg/controller/controller_utils.go
```diff
@@ -758,3 +758,18 @@ func DeleteFinalizer(object metav1.Object, finalizer string) {
...
+	if msSpec.ClusterHardware.AWS == nil {
+		return "", fmt.Errorf("no AWS cluster hardware set on machine spec")
+	}
+
+	if msSpec.ClusterHardware.AWS.AccountSecret.Name != "" {
+		secretName = msSpec.ClusterHardware.AWS.AccountSecret.Name
+	}
+
+	return secretName, nil
+}
```
dgoodwin: Should this return an error if there's no AWS account secret? It looks like that was previous behavior, but I can't think of a reason why it would be anything but an error if that wasn't set, and we went to build a client.

PR looks good to me otherwise.

joelddiaz: ok. modified to error in the event that there is no secret name provided

joelddiaz: @dgoodwin so converting to an error has some implications. today the code allows for not defining the AWS account secret name, and it falls back to assuming that there are instance profiles attached to the instance that will allow the AWS API calls. I know we've discussed moving away from instance profiles, but not sure if we're really ready to force creating the AWS creds/secrets. thoughts?

dgoodwin: Good enough reason to leave it, thanks.

dgoodwin:
Still has a WIP tag, let me know if it's ok to merge.

joelddiaz:
@dgoodwin i rebased and retested. things look good on my end. removed the WIP

PR #311, opened 2018-07-31
Merge cluster API commands into cluster-operator binary

csrwng:
Adds the Cluster API apiserver and controller manager commands to the cluster-operator binary. After this, our Cluster API code will come from what is vendored in to our repo and not from latest master upstream. 

This includes an update of upstream code and changes required to work without a MachineRole in the cluster API machine spec.


csrwng:
@dgoodwin @joelddiaz - this should be ready for review. I am now running a separate container for the cluster-api apiserver. This has solved the issue with the double registration of the cluster api group.

dgoodwin:
LGTM just a simple integration test failure for an unused import.

csrwng:
Finally, took me a while to get integration tests running again, but it's now fixed

PR #312, opened 2018-08-02
Consolidate aws-machine-controller into cluster-operator binary

csrwng:
Makes aws-machine-controller a sub-command of cluster-operator, removing the need for a separate binary.


csrwng:
@dgoodwin this should be ready for review

PR #314, opened 2018-08-03
Add support for cluster delete-after annotation.

dgoodwin:
Specifies a duration to be added to creation time, after which the
cluster deployment is considered expired and should be cleaned up.
Cluster deployments will be requeued for one minute after this time to
ensure timely cleanup.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -65,6 +65,9 @@ const (
 
 	// versionExists indicates that the cluster's desired version does exist
 	versionExists = "VersionExists"
+
+	// deleteAfterAnnotation is the annotation that contains a duration after which the cluster should be cleaned up.
+	deleteAfterAnnotation = "clusteroperator.openshift.io/delete-after"
 )
 
 // NewController returns a new cluster deployment controller.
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -467,6 +470,31 @@ func (c *Controller) syncClusterDeployment(key string) error {
...
+			expiry := clusterDeployment.CreationTimestamp.Add(dur)
+			clusterDeploymentLog.Debugf("cluster expires at: %s", expiry)
+			if time.Now().After(expiry) {
+				clusterDeploymentLog.WithField("expiry", expiry).Info("cluster has expired, issuing delete")
+				c.clustopClient.ClusteroperatorV1alpha1().ClusterDeployments(clusterDeployment.Namespace).Delete(clusterDeployment.Name, &metav1.DeleteOptions{})
+				return nil
+			}
+
+			// We have an expiry time, we're not expired yet, lets make sure to enqueue the cluster for
+			// just after it's expiry time:
```
csrwng: s/it's/its/

contrib/examples/cluster-deployment-template.yaml
```diff
@@ -84,6 +84,8 @@ objects:
   metadata:
     name: ${CLUSTER_NAME}
     namespace: ${CLUSTER_NAMESPACE}
+    annotations:
+      clusteroperator.openshift.io/delete-after: 5m
```
csrwng: Why so low? Right now it takes at least 20m for the cluster to finish provisioning

dgoodwin: My mistake. I'm going to probably remove this in the default template, it'll bite someone no matter what we put in there.

dgoodwin:
Fixed.

PR #315, opened 2018-08-03
Use pre-built images instead of BuildConfig

jhernand:
Currently the deployment playbook uses a `BuildConfig` to build the operator from source during deployment. This is no longer needed, because there are already pre-built images published, so this patch changes the playbook to use those pre-built images.

### DIFF
--- a/contrib/ansible/deploy-playbook.yml
+++ b/contrib/ansible/deploy-playbook.yml
@@ -27,9 +27,6 @@
     # This CA will be generated once and then re-used for all certs. Will be created
     # in the cert dir defined above.
     apiserver_ca_name: ca
-    # Cluster operator git repository to build:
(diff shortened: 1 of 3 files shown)


jhernand:
@dgoodwin @csrwng @nimrodshn @zgalor after these changes I was able to deploy the operator running `ansible-playbook contrib/ansible/deploy-playbook.yml` and then adding manually the pre-built images to the `cluster-operator` image stream, with the `latest` tag:

```
$ oc project openshift-cluster-operator
$ oc tag registry.svc.ci.openshift.org/openshift-cluster-operator/cluster-operator:latest
```

csrwng:
/lgtm
The build part should not be part of the regular deployment. We will have a separate config for builds in CI.

jhernand:
Thanks @csrwng.

dgoodwin:
Thanks @jhernand !

PR #321, opened 2018-08-13
Split cluster-operator-ansible build by version

csrwng:
Allows us to remove workarounds that only applied to previous versions
Introduces image based on the master branch, preparing us to provision origin 3.11


csrwng:
@joelddiaz ptal

PR #323, opened 2018-08-14
Fix missing git in Ansible image

csrwng:
Latest ansible image no longer has git in it. We need git to clone a specific version of the repo.


csrwng:
@joelddiaz ptal

PR #326, opened 2018-08-21
Overwrite roles only for ansible v3.9

csrwng:
Removes code to overwrite security_group.yml in openshift-ansible versions >= 3.10


csrwng:
@abutcher ptal

csrwng:
thank you @abutcher !
I realized I needed to update the Makefile to use the new path for roles

PR #327, opened 2018-08-21
Provision util

csrwng:


contrib/pkg/provision/command.go
```diff
@@ -0,0 +1,206 @@
...
+func NewProvisionClusterCommand() *cobra.Command {
+	userName := "user"
+	u, err := user.Current()
+	if err == nil {
+		userName = u.Username
+	}
+
+	opt := &ProvisionClusterOptions{}
+	cmd := &cobra.Command{
+		Use:   "provision NAME",
```
dgoodwin: This is generically named provision but when we make it to multi-cloud, would it be good to just rename the top level command to provision-aws and we can add more as supported?

dgoodwin: Might make sense to apply to the package name as well if we go ahead with that.

contrib/pkg/provision/buildami.go
```diff
@@ -0,0 +1,91 @@
...
+		o.Logger.WithField("RepositoryURL", o.RPMRepositoryURL).Debug("Obtained RPM repository URL")
+	}
+
+	buildAMIOptions := amibuild.BuildAMIOptions{
+		AccountSecretName:      o.AccountSecretName,
+		SSHSecretName:          o.SSHSecretName,
+		AnsibleImage:           o.AnsibleImage,
+		AnsibleImagePullPolicy: corev1.PullIfNotPresent,
+		RPMRepositoryURL:       o.RPMRepositoryURL,
+		RPMRepositorySource:    o.RPMRepositorySource,
```
dgoodwin: I'm almost positive I'm missing something here but are these two RPM fields used anywhere? I can't see other than where they're validated.

contrib/pkg/provision/command.go
```diff
@@ -0,0 +1,206 @@
...
+	if len(o.Namespace) == 0 {
+		o.Namespace = clients.Namespace
+	}
+	o.Logger = o.Logger.WithField("namespace", o.Namespace)
+	err = o.PreflightChecks(clients)
+	if err != nil {
+		return err
+	}
+	if len(o.ImageID) == 0 {
+		err = o.BuildAMI(clients)
```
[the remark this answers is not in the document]
csrwng: It will, in a separate command that deprovisions :)

contrib/pkg/provision/command.go
```diff
@@ -0,0 +1,206 @@
...
+	flags.StringVar(&opt.LogLevel, "loglevel", "info", "log level, one of: debug, info, warn, error, fatal, panic")
+	// TODO: It is unclear whether this is needed if we're already providing an RPM repo and ImageFormat
+	flags.StringVar(&opt.Version, "origin-version", "3.11", "Version of origin to provision")
+	flags.StringVar(&opt.ImageFormat, "images", "registry.svc.ci.openshift.org/openshift/origin-v3.11:{component}", "Image format for images to use in the installation")
+	flags.StringVar(&opt.KeyPairName, "keypair", "libra", "Name of AWS keypair to use for created instances")
+	flags.StringVar(&opt.Region, "region", "us-east-1", "AWS region to use to provision cluster")
+	flags.StringVar(&opt.KubeconfigFile, "kubeconfig", "", "Name of a kubeconfig file to create when server is provisioned")
+	flags.StringVar(&opt.ImageID, "ami", "", "[Optional] Image ID to use to provision the cluster. If not specified, a new AMI is created")
+	return cmd
+}
```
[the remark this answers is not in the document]
csrwng: Yup, we could use this to create a cluster locally... it would require some additional options like not waiting for the cluster to be ready.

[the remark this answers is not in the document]
csrwng: The high level use case for this is to run this command instead of this call in the e2e test:
https://github.com/openshift/release/blob/master/ci-operator/templates/cluster-launch-e2e.yaml#L191-L193
That call stands up a cluster on GCE (along with a corresponding vm image), ensures that it is running properly and returns success or failure.

contrib/pkg/provision/command.go
csrwng: Makes sense, will update

contrib/pkg/provision/buildami.go
csrwng: The RepositoryURL is used by the playbook that builds the AMI here: https://github.com/openshift/cluster-operator/blob/master/contrib/pkg/amibuild/generate.go#L63

The RepositorySource is used to determine a RepositoryURL if not specified directly. This command the code from the other utility commands like buildami and waitforcluster.

csrwng:
@dgoodwin this should be ready

contrib/pkg/provision_aws/command.go
```diff
@@ -0,0 +1,233 @@
...
+	flags.StringVar(&opt.SSHSecretName, "ssh-secret", "libra-ssh", "secret with SSH key")
+	flags.StringVar(&opt.AnsibleImage, "ansible-image", "registry.svc.ci.openshift.org/openshift-cluster-operator/cluster-operator-ansible:latest", "ansible image to use for job")
+	flags.StringVar(&opt.RPMRepositoryURL, "rpm-repository", "", "RPM repository for Origin")
+	flags.StringVar(&opt.RPMRepositorySource, "rpm-repository-source", "https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/releases/openshift/origin/master/origin.repo", "pointer to RPM repository URL")
+	flags.StringVar(&opt.VPCName, "vpc", "", "VPC to use for the cluster. If left blank, a new VPC will be created")
+	flags.StringVar(&opt.AMIVPCName, "ami-vpc", "default", "VPC to use to build the AMI.")
+	flags.StringVar(&opt.LogLevel, "loglevel", "info", "log level, one of: debug, info, warn, error, fatal, panic")
+	// TODO: It is unclear whether this is needed if we're already providing an RPM repo and ImageFormat
+	flags.StringVar(&opt.Version, "origin-version", "3.11", "Version of origin to provision")
```
dgoodwin: Wondering if this would be better off not origin specific, maybe just openshift-version? (in case this ever got used for ocp, or somebody wanted it to be okd :)

contrib/pkg/provision_aws/certs.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+type ServingCert struct {
+	Cert []byte
+	Key  []byte
+}
+
+func (o *ProvisionClusterOptions) GenerateServingCert() (*ServingCert, error) {
```
dgoodwin: This looks like some handy code.

contrib/pkg/provision_aws/command.go
```diff
@@ -0,0 +1,233 @@
...
+				return
+			}
+			if err := opt.Run(); err != nil {
+				log.WithError(err).Error("Runtime error")
+			}
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVarP(&opt.Namespace, "namespace", "", "", "namespace to use. Defaults to current namespace.")
+	flags.StringVar(&opt.AccountSecretName, "account-secret", fmt.Sprintf("%s-aws-creds", userName), "secret to use for AWS account")
```
dgoodwin: Not sure if this is a good idea but maybe aws-creds-secret for this param name? Will leave up to you, not sure which is best.

contrib/pkg/provision_aws/command.go
```diff
@@ -0,0 +1,233 @@
...
+	flags.StringVar(&opt.SSHSecretName, "ssh-secret", "libra-ssh", "secret with SSH key")
```
dgoodwin: Capitalization and punctuation jumps around a bit in these param descriptions.

contrib/pkg/deprovision_aws/command.go
```diff
@@ -0,0 +1,171 @@
...
+			if err := opt.Run(); err != nil {
+				log.WithError(err).Error("Runtime error")
+			}
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVar(&opt.AccountSecretName, "account-secret", fmt.Sprintf("%s-aws-creds", userName), "secret to use for AWS account")
+	flags.StringVarP(&opt.Namespace, "namespace", "", "", "namespace to use. Defaults to current namespace.")
+	flags.StringVar(&opt.LogLevel, "loglevel", "info", "log level, one of: debug, info, warn, error, fatal, panic")
+	flags.StringVar(&opt.Region, "region", "us-east-1", "AWS region to use to provision cluster")
```
dgoodwin: "where the cluster was provisioned" might be better in this context.

contrib/pkg/provision_aws/command.go
```diff
@@ -0,0 +1,233 @@
...
+	flags.StringVar(&opt.AnsibleImage, "ansible-image", "registry.svc.ci.openshift.org/openshift-cluster-operator/cluster-operator-ansible:latest", "ansible image to use for job")
+	flags.StringVar(&opt.RPMRepositoryURL, "rpm-repository", "", "RPM repository for Origin")
+	flags.StringVar(&opt.RPMRepositorySource, "rpm-repository-source", "https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/releases/openshift/origin/master/origin.repo", "pointer to RPM repository URL")
+	flags.StringVar(&opt.VPCName, "vpc", "", "VPC to use for the cluster. If left blank, a new VPC will be created")
+	flags.StringVar(&opt.AMIVPCName, "ami-vpc", "default", "VPC to use to build the AMI.")
```
dgoodwin: Does this one have to exist?

contrib/pkg/provision_aws/command.go
```diff
@@ -0,0 +1,233 @@
...
+			}
+		},
+	}
+	flags := cmd.Flags()
+	flags.StringVarP(&opt.Namespace, "namespace", "", "", "namespace to use. Defaults to current namespace.")
+	flags.StringVar(&opt.AccountSecretName, "account-secret", fmt.Sprintf("%s-aws-creds", userName), "secret to use for AWS account")
+	flags.StringVar(&opt.SSHSecretName, "ssh-secret", "libra-ssh", "secret with SSH key")
+	flags.StringVar(&opt.AnsibleImage, "ansible-image", "registry.svc.ci.openshift.org/openshift-cluster-operator/cluster-operator-ansible:latest", "ansible image to use for job")
+	flags.StringVar(&opt.RPMRepositoryURL, "rpm-repository", "", "RPM repository for Origin")
+	flags.StringVar(&opt.RPMRepositorySource, "rpm-repository-source", "https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/releases/openshift/origin/master/origin.repo", "pointer to RPM repository URL")
```
dgoodwin: I probably should know this but how do these two params differ?

contrib/pkg/provision_aws/command.go
```diff
@@ -0,0 +1,233 @@
...
+	flags.StringVar(&opt.VPCName, "vpc", "", "VPC to use for the cluster. If left blank, a new VPC will be created")
+	flags.StringVar(&opt.AMIVPCName, "ami-vpc", "default", "VPC to use to build the AMI.")
+	flags.StringVar(&opt.LogLevel, "loglevel", "info", "log level, one of: debug, info, warn, error, fatal, panic")
+	// TODO: It is unclear whether this is needed if we're already providing an RPM repo and ImageFormat
+	flags.StringVar(&opt.Version, "origin-version", "3.11", "Version of origin to provision")
+	flags.StringVar(&opt.ImageFormat, "images", "registry.svc.ci.openshift.org/openshift/origin-v3.11:${component}", "Image format for images to use in the installation")
+	flags.StringVar(&opt.KeyPairName, "keypair", "libra", "Name of AWS keypair to use for created instances")
+	flags.StringVar(&opt.Region, "region", "us-east-1", "AWS region to use to provision cluster")
+	flags.StringVar(&opt.KubeconfigFile, "kubeconfig", "", "Name of a kubeconfig file to create when server is provisioned")
```
dgoodwin: Suggest: s/when server/once the control plane

contrib/pkg/deprovision_aws/cluster.go
```diff
@@ -0,0 +1,65 @@
...
+		_, err = clients.ClusterOpClient.ClusteroperatorV1alpha1().ClusterDeployments(o.Namespace).Get(o.Name, metav1.GetOptions{})
+		if errors.IsNotFound(err) {
+			o.Logger.Info("ClusterDeployment has been deleted")
+			return true, nil
+		}
+		if err != nil {
+			o.Logger.WithError(err).Error("unexpected error retrieving ClusterDeployment")
+			return false, err
+		}
+		o.Logger.Debug("ClusterDeployment has not been deleted. Continuing to wait")
```
dgoodwin: Someday it would be neat if this would print where in the process it was, maybe finalizers remaining or something.

contrib/pkg/provision_aws/command.go
dgoodwin: I think I see what's up from going through earlier code again, these could use some more detail in the description to clarify how they're used.

contrib/pkg/deprovision_aws/cluster.go
csrwng: ack

contrib/pkg/provision_aws/command.go
csrwng: It does, because creating an AMI prereqs a VPC. Which one you use depends on which VPCs exist in your account.

csrwng:
@dgoodwin thx for the review, addressed your comments

PR #328, opened 2018-08-23
Set openshift_master_cluster_hostname as internal elb hostname.

abutcher:
Workaround for the removal in https://github.com/openshift/openshift-ansible/pull/9680


csrwng:
@abutcher this looks good to me. Something seems to have changed on the centos ami we use for our tests. I am testing a quick fix, but longer term I think we need to containerize any ansible invocations in the test.

PR #331, opened 2018-08-27
Fix IAM role creation with openshift-ansible v3.11

csrwng:
Fixes issue introduced by a change in openshift-ansible that removes the `openshift_aws_master_group_config` variable from the openshift_aws role.

The override will only be done for the cluster-operator-ansible image that's based on master and will not affect v3.9 or v3.10 builds.


csrwng:
@abutcher ptal

PR #332, opened 2018-08-28
Install git in origin ansible 3.9 image

twiest:


twiest:
@csrwng Would you mind reviewing this?

It's the same fix you did for the origin-ansible 3.10 image, but for 3.9.

PR #338, opened 2018-09-12
Add coutil verify-import to make verify.

twiest:
- [x] Add coutil verify-import to "make verify".
- [x] Verify that a local "make verify" verifies imports


twiest:
@csrwng would you mind reviewing this PR please?

Please note [this part](https://github.com/openshift/cluster-operator/pull/338/files#diff-b67911656ef5d18c4ae36cb6741b7965R146) where I add coutil as part of the .generate_exes make target. Is that acceptable?

PR #343, opened 2018-10-19
Fix image problems

twiest:
Fix image problems
    - [x] Fix golint installation in docker image build
    - [x] Test that "make images" now works
    - [x] Change cluster-operator images to point to quay.io
    - [x] Test that cluster deploy now works again

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,6 @@ GO_BUILD       = env GOOS=$(PLATFORM) GOARCH=$(ARCH) go build -i $(GOFLAGS) \
 BASE_PATH      = $(ROOT:/src/github.com/openshift/cluster-operator/=)
 export GOPATH  = $(BASE_PATH):$(ROOT)/vendor
 
-PUBLIC_REGISTRY=quay.io/openshift/
 
 MUTABLE_TAG                         ?= canary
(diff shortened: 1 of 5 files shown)


dgoodwin:
We should probably shut down E2E testing on cluster-operator, don't know what this is but it's not worth fixing.

I'd force merge this but I don't have perms. :)

twiest:
@csrwng Hey, would you mind taking a look at this please?

Note that I set the image URLs to `quay.io/twiest/cluster-operator:latest` like we discussed. I don't feel great about that, but it's probably the best thing to do for now.

csrwng:
@twiest my only comment is that it may not be a bad idea to update the registry prefix in the Makefile -> https://github.com/openshift/cluster-operator/blob/master/Makefile#L68 (It's never going to be quay.io/openshift)

twiest:
@csrwng I don't see that variable actually used anywhere. Should I just remove it?

csrwng:
@twiest ha,  you're right. Yup, go ahead and remove i

csrwng:
@twiest we should remove the jenkins/e2e test from the list of tests ... that relied on the images coming from the api.ci registry which no longer exist. I'll submit a PR for that.

/lgtm

